PR #401, opened 2024-09-09
feat: handle service change and route-wide detour alerts

boringcactus:
### Summary

_Ticket:_ [Handle alerts with unspecified stops + service changes](https://app.asana.com/0/1201654106676769/1208057797841296/f)

Adds a slot in the `PredictionRowView` for secondary alerts that don't replace the upcoming trips.

### Testing

Wrote some unit and UI tests and manually checked that this correctly handles a currently-active untargeted bus detour.


JackVCurtis (approved):
Looks good! Really appreciate the comments here.

PR #402, opened 2024-09-10
fix(iOS): upgrade deps for Xcode 16 support

boringcactus:
### Summary

_Ticket:_ none

Since Xcode Cloud has started running our builds in Xcode 16 RC, we may as well go ahead and upgrade our dependencies to be compatible with Xcode 16.

### Testing

Checked locally in Xcode 16 that the app builds and runs successfully.


boringcactus:
There's a type checking timeout in `ViewInspector` and no upgrade available. I've set the Xcode Cloud build back to Xcode 15, and we can revisit this later.

PR #404, opened 2024-09-10
feat: Persist cached data to disk

EmmaSimon:
### Summary

_Ticket:_ [Persist cached static data across app launches](https://app.asana.com/0/1205425564113216/1208010699681068/f)

This adds disk caching to the existing `ResponseCache` class. A new dependency on [okio](https://github.com/square/okio) was added to handle cross platform reads and writes. This change on its own isn't as useful as it could be without returning a [data stream](https://app.asana.com/0/1205732265579288/1208254208398831), because every time the cache expires (which will almost always happen between app launches), a request still needs to be made to check if the data is up to date.

### Testing

Added unit tests to ensure that `ResponseCache` handles reading and writing to disk properly


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
```diff
@@ -2,67 +2,87 @@ package com.mbta.tid.mbta_app.cache
...
+import org.koin.core.parameter.parametersOf
 
-sealed class ConditionalResponse {
-    data object NotModified : ConditionalResponse()
+@Serializable
+data class Response(
+    val etag: String?,
+    @Serializable(with = TimeMarkSerializer::class)
+    var fetchTime: TimeSource.Monotonic.ValueTimeMark,
+    val data: String
```
EmmaSimon, line 25: This double encodes the JSON response body, which feels gross, but I don't see another good way to make this support arbitrary response types. I think ideally we'd make `ResponseCache` generic and handle all serde internally, but [Koin doesn't play well](https://insert-koin.io/docs/reference/koin-core/definitions/#dealing-with-generics) with generics. 

Since the `ResponseCache` koin module is just a factory instead of a singleton now, we might be able to get away with defining them normally, as needed in the singleton repositories, but there's still some awkwardness with Json decoding also not loving generics, so we'd still need to pass in a serializer separately.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
```diff
@@ -38,10 +39,15 @@ fun appModule(appVariant: AppVariant) = module {
...
+        if (params.size() == 2) {
+            ResponseCache(cacheKey = params.get(), maxAge = params.get())
+        } else if (params.size() == 1) {
+            ResponseCache(cacheKey = params.get())
         } else {
-            ResponseCache()
+            throw InstanceCreationException(
+                "Wrong number of arguments provided to cacheModule",
+                parent = Exception()
+            )
```
EmmaSimon, line 50: I wish there was a cleaner way to deal with named params in koin, like I said above, maybe it's not worth having this injected at all.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
```diff
@@ -2,67 +2,87 @@ package com.mbta.tid.mbta_app.cache
...
-        val data = this.data ?: return null
-        val dataTimestamp = this.dataTimestamp ?: return null
-        return if (dataTimestamp.elapsedNow() < maxAge) {
-            data.response
+    private fun getData(): Response? {
+        val data = this.data ?: readData() ?: return null
+        return if (data.fetchTime.elapsedNow() < maxAge) {
+            data
         } else {
             null
```
EmmaSimon, line 38: Right now we don't actually ever delete cached data once it's written to disk. The obvious place for deletes would be right here when the data is stale, but it's then possible to get a NotModified response right after, and it seems like a waste to be constantly deleting and recreating the file every time we get NotModified.

JackVCurtis (approved):
👍

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
boringcactus: It's probably not worth injecting it - there's no `IResponseCache` that we could use to implement a fake cache, and even if there was, the only place we'd use it is in `GlobalRepositoryTest` so we could easily just make it a parameter to `GlobalRepository` that defaults to a new real cache. If we wanted to, we could define and inject an explicit `ResponseCacheFactory`, but I'm not sure that'd solve a problem we actually have.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
EmmaSimon: Yeah I considered it, but it seemed cleanest from an implementation standpoint to have them colocated, and it's harder for them to get out of sync this way. If anything, it probably makes the most sense to put it in a DataStore.

boringcactus: Double-encoding 3 MB of JSON might just be worse.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
EmmaSimon: Well, in the future we'll use it in at least one more repo, possibly more. But yeah I don't hate the idea of having it be a default param. We could also potentially make a separate koin module for each repository and key (though if we have a repo that programmatically generates keys with a param, it'll need to pass that in as a koin param still), or if we make it generic, each possible data type.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
EmmaSimon: Yeah fair, that'll also mean we don't need to rewrite the entire 3MB file every time notmodified updates the timestamp. Cool, I'll start on moving them into DataStore.

boringcactus: Making the `ResponseCache` generic seems like it'd be worth doing regardless. Unfortunately, you can't have an `inline constructor` to automatically determine the serializer, but you can declare an `inline fun` on the `companion object` to get a similar effect: https://pl.kotl.in/fH4yXhJaq

EmmaSimon: Oh interesting, yeah that's clean, thanks for the suggestion!

EmmaSimon: Okay, made both of those changes. I'm much more satisfied with this! I ended up just tossing the metadata into another json file after all, DataStore seemed like more trouble than it was worth.

boringcactus (approved):

PR #408, opened 2024-09-13
fix: show now instead of arr/brd for bus in trip details

BrandonTR:
### Summary

_Ticket:_ [For bus, use Now instead of ARR/BRD](https://app.asana.com/0/1205732265579288/1207678190799457/f)

Missed changing this for trip details. Turns out we're not passing in `RouteType` to the format function there.

### Testing

Added a unit test on the affected view for good measure.

### DIFF
--- a/iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
+++ b/iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
@@ -205,7 +205,7 @@ extension HomeMapView {
         })
         let stopSequence = trip?.stopSequence?.intValue
 
-        guard let routeId = vehicle.routeId ?? trip?.trip.id else {
+        guard let routeId = trip?.trip.routeId ?? vehicle.routeId else {
             // TODO: figure out something to do if this is nil
             return
         }
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -229,12 +229,13 @@ struct ContentView: View {
                 tripId: tripId,
                 vehicleId: vehicleId,
                 target: target,
-                routeId: _,
+                routeId: routeId,
                 directionId: _
             ):
                 TripDetailsPage(
                     tripId: tripId,
                     vehicleId: vehicleId,
+                    routeId: routeId,
                     target: target,
                     nearbyVM: nearbyVM,
                     mapVM: mapVM
--- a/iosApp/iosApp/Localizable.xcstrings
+++ b/iosApp/iosApp/Localizable.xcstrings
@@ -290,6 +290,9 @@
     "No Service" : {
 
     },
+    "Now" : {
+
+    },
     "Now at" : {
 
     },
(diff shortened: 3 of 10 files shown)


iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -220,7 +223,8 @@ extension HomeMapView {
                 stopSequence: stopSequence
             ) : nil,
             routeId: routeId,
-            directionId: vehicle.directionId
+            directionId: vehicle.directionId,
+            routeType: patterns?.routes.first?.type
```
BrandonTR, line 227: Is this ok? Not sure if this is a reliable way to determine the type.

EmmaSimon: My initial thought would be to use the `routeId` that's already in the navigation item to fetch the route from global data where you need it in the trip page, and check the type there. Is there a reason I'm missing that that wouldn't work?

EmmaSimon: While trying to think through that approach, I realized that [this](https://github.com/mbta/mobile_app/pull/408/files#diff-0596f8683751c0aab7db8c7ccfd1736f247c979f0e30eea1e92a8803747f751aR211) is using `trip?.trip.id` as a fallback, which seems wrong, shouldn't it be `trip?.trip.routeId`?

FEz4HC5XsAMztm3

Yeah uhh, try changing it and see if it breaks anything, but I think this was my mistake, the previous code before that change was using `trip.routeId`. Then that should make it possible to get the route by ID. And now that I think of it, the trip's routeId should probably be higher priority than the vehicle's, since you've selected a specific trip on a specific route, which may have a vehicle serving a different route.

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -205,23 +208,27 @@ extension HomeMapView {
         })
         let stopSequence = trip?.stopSequence?.intValue
 
-        guard let routeId = vehicle.routeId ?? trip?.trip.id else {
+        guard let routeId = trip?.trip.routeId ?? vehicle.routeId else {
             // TODO: figure out something to do if this is nil
             return
         }
+        let routeType = globalData?.routes[routeId]?.type
```
EmmaSimon, line 215: Oh, sorry, I was unclear. I was thinking that it would be better to do this by passing the route ID into `TripDetailsPage`, then get the route from global to check the route type there, so that we wouldn't need to include the route type in the navigation entry at all. It seems redundant to include a piece of information that's derived from a different piece of information that's already included. And since changing the nav items has a relatively broad reach, I think it's best to only put strictly necessary and unique filter information in there.

BrandonTR: Oh, got it! I'll make that change 😄

EmmaSimon (approved):
Thanks for making that change and fixing my bug!

PR #409, opened 2024-09-13
feat: rewrite trips and collapse route patterns during temporary terminals

boringcactus:
### Summary

_Ticket:_ [Show only temporary terminals as headsigns during subway shuttles/suspensions](https://app.asana.com/0/1205732265579288/1208057797841305/f)

I really wish we just had correct data upstream. This was initially some of the ugliest code I've written in my life, and it took longer to refactor than it did to write in the first place.

### Testing

Added tests with real data taken from the recent partial RL closure.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -595,6 +595,19 @@ private fun NearbyStaticData.rewrittenForTemporaryTerminals(
         return isSubway && routeHasAlert && schedulesNeverTypical && predictionsAlwaysTypical
     }
 
+    fun RoutePattern.isTruncationOf(
+        fullPattern: RoutePattern,
+        fullPatternStopIds: List<String>,
+        stopId: String
+    ): Boolean {
+        if (typicality == RoutePattern.Typicality.Typical) return false
+        if (directionId != fullPattern.directionId) return false
+        val truncatedPatternTrip = globalData.trips[representativeTripId] ?: return false
+        checkNotNull(truncatedPatternTrip.stopIds) { fetchedWithoutStopIds(this) }
+        return truncatedPatternTrip.stopIds.contains(stopId) &&
+            fullPatternStopIds.containsSubsequence(truncatedPatternTrip.stopIds)
+    }
+
     for (routeId in relevantRouteIds) {
         val route = globalData.routes[routeId] ?: continue
 
@@ -621,18 +634,7 @@ private fun NearbyStaticData.rewrittenForTemporaryTerminals(
                         // consider all the patterns that could be in the schedule
                         val plausibleTruncatedPatterns =
                             routePatternsByRoute[routeId].orEmpty().filter {
-                                if (it.typicality == RoutePattern.Typicality.Typical)
-                                    return@filter false
-                                if (it.directionId != fullPattern.directionId) return@filter false
-                                val truncatedPatternTrip =
-                                    globalData.trips[it.representativeTripId] ?: return@filter false
-                                checkNotNull(truncatedPatternTrip.stopIds) {
-                                    fetchedWithoutStopIds(it)
-                                }
-                                truncatedPatternTrip.stopIds.contains(stopId) &&
-                                    fullPatternTrip.stopIds.containsSubsequence(
-                                        truncatedPatternTrip.stopIds
-                                    )
+                                it.isTruncationOf(fullPattern, fullPatternTrip.stopIds, stopId)
                             }
 
                         val truncatedPatternId =
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -565,9 +565,6 @@ private fun NearbyStaticData.rewrittenForTemporaryTerminals(
     val predictionsByRoute = predictions.predictions.values.groupBy { it.routeId }
     val routePatternsByRoute = globalData.routePatterns.values.groupBy { it.routeId }
 
-    val relevantRouteIds =
-        predictions.predictions.values.mapTo(mutableSetOf()) { it.routeId }.toSet()
-
     val rewrittenTrips = predictions.trips.toMutableMap()
     // from Pair(fullPattern.id, stopId) to truncatedPattern.id
     val truncatedPatternByFullPatternAndStop = mutableMapOf<Pair<String, String>, String?>()
@@ -582,15 +579,19 @@ private fun NearbyStaticData.rewrittenForTemporaryTerminals(
                     alert.anyInformedEntity { it.appliesTo(routeId = routeId) }
             }
 
+        val routeSchedules = schedulesByRoute[routeId].orEmpty()
         val schedulesNeverTypical =
-            schedulesByRoute[routeId].orEmpty().all {
-                it.routePattern()?.typicality != RoutePattern.Typicality.Typical
-            }
+            routeSchedules.isNotEmpty() &&
+                routeSchedules.all {
+                    it.routePattern()?.typicality != RoutePattern.Typicality.Typical
+                }
 
+        val routePredictions = predictionsByRoute[routeId].orEmpty()
         val predictionsAlwaysTypical =
-            predictionsByRoute[routeId].orEmpty().all {
-                it.routePattern()?.typicality == RoutePattern.Typicality.Typical
-            }
+            routePredictions.isNotEmpty() &&
+                routePredictions.all {
+                    it.routePattern()?.typicality == RoutePattern.Typicality.Typical
+                }
 
         return isSubway && routeHasAlert && schedulesNeverTypical && predictionsAlwaysTypical
     }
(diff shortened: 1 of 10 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -466,6 +533,214 @@ fun NearbyStaticData.withRealtimeInfo(
...
+ * canonical route patterns. The way we work around this is that for
+ * 1. subway routes
+ * 2. with alerts somewhere on the line
+ * 3. where the schedule has non-typical patterns instead of typical patterns
+ * 4. but the predictions all have typical patterns instead of non-typical patterns
+ *
+ * we check the stops of the scheduled patterns against both the predicted pattern and the
+ * prediction's current stop to override the prediction's route pattern with a correct scheduled
+ * pattern, and then we load the headsign from the new route pattern's representative trip. This
+ * should be narrowly scoped enough to apply to temporary terminals but no other trips. Hopefully.
```
KaylaBrady, line 548: praise: I appreciate the detailed comment here! Is there any relevant code from screenplay that is relevant to link here?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -466,6 +533,214 @@ fun NearbyStaticData.withRealtimeInfo(
...
+ */
+private fun NearbyStaticData.rewrittenForTemporaryTerminals(
+    predictions: PredictionsStreamDataResponse?,
+    globalData: GlobalResponse?,
+    alerts: List<Alert>?,
+    schedules: ScheduleResponse?,
+): Pair<NearbyStaticData, PredictionsStreamDataResponse?> {
```
KaylaBrady, line 555: suggestion: This is some pretty complicated logic - I think it would be worth breaking this out into named & documented helper functions.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -466,6 +533,214 @@ fun NearbyStaticData.withRealtimeInfo(
...
 }
 
+/**
+ * For reasons, for mid-line subway shuttles/suspensions, schedules will correctly reflect the
+ * closure and use route patterns that stop at temporary terminals, but predictions will use the
+ * canonical route patterns. The way we work around this is that for
+ * 1. subway routes
+ * 2. with alerts somewhere on the line
+ * 3. where the schedule has non-typical patterns instead of typical patterns
+ * 4. but the predictions all have typical patterns instead of non-typical patterns
```
KaylaBrady, line 543: question: How does this perform is there is an alert on only a couple of stops on one of the RL branches?

KaylaBrady (commented):
This is complicated, I also hope this can be fully resolved upstream. Leaving a few initial comments

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: Not that I know of. Why would there be? Am I missing something?

boringcactus: It's [not](https://mbta.slack.com/archives/C03K6NLKKD1/p1726160321566569?thread_ts=1726096049.324989&cid=C03K6NLKKD1) - they're scanning the trip's full predictions to get its actual last stop and using that as the headsign, but we don't have the trip's full predictions, and I don't think we want them just for this.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: Is the path I'm on now of having nested helper functions inside `NearbyStaticData.rewrittenForTemporaryTerminals` reasonable, or would you rather have non-nested helper functions (perhaps even with their own unit tests)? The main advantage of nested helper functions is they can refer to the `globalData` without needing to take it as a parameter, but that may not be worth it.

KaylaBrady: I think non-nested helpers would be easier to read, and unit tests for meaningfully complex helpers would be nice too.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: Correctly now, I think. (I'm beginning to appreciate the value of unit testing the individual helper functions.)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: I have moved all the helpers into a separate `TemporaryTerminalRewriter` class which also encapsulates all the state, and I've written unit tests for all the individual pieces.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
```diff
@@ -0,0 +1,254 @@
...
+    val alerts: List<Alert>,
+    val schedules: ScheduleResponse
+) {
+    private val schedulesByRoute = schedules.schedules.groupBy { it.routeId }
+    // predictions that reflect auto-cancelled schedules will throw everything off, so filter out
+    // cancellations
+    val predictionsByRoute =
+        predictions.predictions.values
+            .filterNot { it.scheduleRelationship == Prediction.ScheduleRelationship.Cancelled }
+            .groupBy { it.routeId }
```
KaylaBrady, line 30: question: why will these throw everything off? Don't we want to surface cancellations? It seems a bit risky to do this filtering outside of checking `appliesToRoute`

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /**
+     * Finds a truncated pattern for [fullPattern] that includes [stopId]. Since at time of writing
+     * we only fetch scheduled trips for the remainder of the day, does not require that the
+     * truncated pattern actually appear in the schedule unless there are multiple plausible
+     * patterns (which could happen if the rating includes both a current disruption and a future
+     * disruption on the same line). If there is no plausible truncated pattern, or if there are
+     * several, does not truncate [fullPattern].
+     */
```
KaylaBrady, line 111: This has taken me a few reads to grasp, the unit tests help. Perhaps the name `findTruncatedPattern` or `matchTruncatedPattern` for this and `truncatePatternsAtStop` would be clearer that this isn't doing any truncation itself

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
```diff
@@ -0,0 +1,254 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * Checks each typical pattern and child stop at [stopPatterns] for truncation, and merges full
+     * patterns into their truncated [NearbyStaticData.StaticPatterns].
+     */
+    fun truncatePatternsAtStop(
```
KaylaBrady, line 141: I'm having a hard time wrapping my head around this function - "merges full patterns into their truncated patterns" seems somewhat simpler than what this entails. I think it would help the readability of this and the flow through these functions if it were more functional - instead of mutating truncatedPatternByFullPatternAndStop, could it return the map?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
```diff
@@ -0,0 +1,254 @@
...
+                                    truncatedPatternByFullPatternAndStop.containsValue(pattern.id)
+                                }
+                            if (isTruncated) 0 else 1
+                        }
+                    patternsCorrectFirst
+                        .reduce { acc, nextPatterns ->
+                            acc.copy(patterns = acc.patterns + nextPatterns.patterns)
+                        }
+                        .let { it.copy(patterns = it.patterns.sorted()) }
+                }
```
KaylaBrady, line 186: I'm confused by what this is doing - could it be broken down differently or commented?

KaylaBrady (commented):
Thanks for breaking this down and adding tests! I'm still struggling a bit with the complexity of it, a few small comments for things that I think would help.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
boringcactus: Renamed this function.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
boringcactus: I genuinely can't think of a simpler way to merge the full patterns into their truncated patterns than what I'm doing here. Maybe I've just been working on this task for too long; what were you expecting instead?

boringcactus: Returning the map makes things a little bit more complicated, because `truncatePatternsAtStop` is called once per stop but `rewritePredictions` is called once per route, but if you think that complexity would make the code more legible I can do that.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
boringcactus: Broke this down slightly differently and gave the `groupBy` lambda a name.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
KaylaBrady: Maybe I'm confused by what merge is doing here. It seems like we are replacing full patterns with the truncated one, is that right? But I think I'm still not getting the purpose of collapsedPatterns and the relevance of the sorting within it

boringcactus: Technically, I guess with the features we have right now, we could be just discarding the full patterns completely, since we're rewriting the predictions' trips to point to the truncated patterns. However, if we did that, when we do something in the future to represent downstream alerts, they won't appear on the temporary terminal headsign, and I think that'd be incorrect. What we actually want to do is merge the full `RoutePattern`s from the full pattern `StaticPatterns` into the truncated `RoutePattern`s' `StaticPatterns`, so that we keep the truncated `StaticPatterns.headsign` but we still have the full `RoutePattern`s. If we don't sort the truncated `StaticPatterns` to the front, then a full `StaticPatterns` might wind up at the front, in which case we'd use that headsign instead, and that's not what we want.

That said, I think if we're merging the full patterns into the truncated `StaticPatterns`, we may not actually need to be also rewriting the predictions, since they'll still be grouped into the truncated headsign. I'm not quite sure if I want to tear that code out, though - rewriting both the `StaticPatterns` and the prediction trip feels like a belt-and-suspenders approach, and since there isn't a temporary terminal right now, we can't easily validate that we only need the belt.

boringcactus: If we don't rewrite the trips, then at JFK, Alewife gets discarded as an arrival-only pattern, which is incorrect. (In my tests, the northbound-to-JFK patterns should be being discarded as arrival-only, but the schedules there have slightly inaccurate default data from the `ObjectCollectionBuilder`, which I'm about to fix.)

KaylaBrady (approved):
Nice work on a gnarly workaround!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
boringcactus: I thought I had finally come up with a cleaner approach for merging things, but it's breaking keeping the ordinary terminal visible and alerting at the disruption boundary, and since the PR is already approved I'm not going to sink even more time into getting that approach working.

PR #412, opened 2024-09-17
fix: handle alert boundaries correctly

boringcactus:
### Summary

_Ticket:_ [Correctly handle shuttle/suspension boundaries](https://app.asana.com/0/1205732265579288/1207247862893776/f)

I guess it makes sense that moving the child stop IDs from the `StopPatterns` to the `StaticPatterns` would be enough to fix everything here, but I'm still pleasantly surprised.

### Testing

Added a unit test and checked that JFK during the current disruption shows the alert only to Braintree and not also to Alewife or Ashmont.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -395,6 +382,19 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
             )
         }
 
+        private fun filterStopsByPatterns(
+            routePatterns: List<RoutePattern>,
+            global: GlobalResponse,
+            localStops: Set<String>
+        ): Set<String> {
+            val patternsStops =
+                routePatterns.flatMapTo(mutableSetOf()) {
```
JackVCurtis, line 391: Haven't seen `flatMapTo`, very clean!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
```diff
@@ -141,23 +141,25 @@ class TemporaryTerminalRewriter(
     fun truncatePatternsAtStop(
         stopPatterns: NearbyStaticData.StopPatterns
     ): NearbyStaticData.StopPatterns {
-        for (fullPattern in stopPatterns.patterns.flatMap { it.patterns }) {
-            if (fullPattern.typicality != RoutePattern.Typicality.Typical) continue
-            val fullPatternStopIds =
-                globalData.trips[fullPattern.representativeTripId]?.stopIds ?: continue
+        for (fullStaticPatterns in stopPatterns.patterns) {
```
JackVCurtis, line 144: This amount of nested `for` loops is kind of a yellow flag for me - it's probably not worth optimizing, but I wouldn't say no to some comments for clarity.

JackVCurtis (approved):
Looks good, one suggestion but nothing blocking

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: I wish sometimes that Kotlin iterable methods were lazy like Rust iterator methods are, so that `.flatMap().toSet()` wouldn't construct the full list with duplicates, but since they aren't and it would, I thought it made sense to keep this slightly more efficient.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TemporaryTerminalRewriter.kt
boringcactus: Added a comment.

PR #413, opened 2024-09-17
fix: don't draw child stops on map

boringcactus:
### Summary

_Ticket:_ [Don't show child stops and entrances on map](https://app.asana.com/0/1205732265579288/1208288117643287/f)

Removes calls to the child stop drawing logic, along with the placeholder icons.

### Testing

Checked that the child stops are no longer drawn.


JackVCurtis (approved):
Well that certainly simplifies things.

EmmaSimon (approved):

PR #418, opened 2024-09-20
feat: Add "No service today" departure status

EmmaSimon:
### Summary

_Ticket:_ [Handle no service for the entire service day](https://app.asana.com/0/1205732265579288/1208217009015138/f)

Add a new state for departure rows to note when a route pattern has no service in that day. This is a relatively rare state, because we still hide non-typical route patterns, and many patterns which have no trips today are also not typical. This also extends to some entire routes, which _only_ have non-typical patterns, like the CapeFlyer. A notable exception is the Foxboro Event Service, which has a typical pattern and will show up in nearby transit and stop pages.

Screenshot_20240922-092639 Screenshot_20240922-092750

### Testing

Did some manual testing over the weekend and saw many weekday only typical routes with no service, properly sorted to the bottom of nearby transit when all patterns on a route had no service.

Added unit tests for all pieces of functionality, including schedule checking, formatting, and sorting.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -546,8 +564,21 @@ fun NearbyStaticData.withRealtimeInfo(
...
+                    it.patternsByStop.all { byStop ->
+                        byStop.patterns.all { patterns ->
+                            patterns.upcomingTrips.isNullOrEmpty() && !patterns.hasSchedulesToday
+                        }
+                    }
+                }
+                .thenBy(Route.subwayFirstComparator) { it.sortRoute() }
+                .thenBy { it.distanceFrom(sortByDistanceFrom) }
+                .thenBy { it.sortRoute() }
+        )
```
EmmaSimon, line 581: I got stuck on the sorting for a bit, until I realized that I needed to split up the `Route.relevanceComparator`, because we needed to sort the non-scheduled routes after the pinning but before putting subway first. This is risky in case we add more steps to the `relevanceComparator`, it would be awkward to get the schedule sorting to conditionally happen within there.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
```diff
@@ -109,11 +111,17 @@ data class UpcomingTrip(
...
+                            schedulesHere.orEmpty(),
+                            predictionsHere.orEmpty(),
+                            trips,
+                            predictions?.vehicles.orEmpty()
+                        )
+                        .filter { upcomingTrip ->
+                            if (upcomingTrip.prediction != null) return@filter true
+                            val scheduleTime =
+                                upcomingTrip.schedule?.scheduleTime ?: return@filter true
+                            scheduleTime.minus(filterAtTime) >= ZERO
```
boringcactus, line 123: Would `scheduleTime >= filterAtTime` work here?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
```diff
@@ -33,13 +35,15 @@ sealed class RealtimePatterns : Comparable<RealtimePatterns> {
         override val patterns: List<RoutePattern>,
         override val upcomingTrips: List<UpcomingTrip>? = null,
         override val alertsHere: List<Alert>? = null,
+        override val hasSchedulesToday: Boolean,
```
boringcactus, line 38: For tests, it would be convenient if `hasSchedulesToday` defaulted to `true` - that's why `upcomingTrips` and `alertsHere` default to `null` - and if `hasSchedulesToday` doesn't have a default and isn't moved before `upcomingTrips` and `alertsHere` then the defaults for `upcomingTrips` and `alertsHere` will never be usable and should be removed. A lot of the files touched here are just catching up to the lack of defaults here.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/NearbyResponseTest.kt
```diff
@@ -1316,6 +1341,190 @@ class NearbyResponseTest {
...
+                latitude = closeBusStop.latitude + 0.3
+                longitude = closeBusStop.longitude + 0.3
+            }
+        val farSubwayStop =
+            objects.stop {
+                latitude = closeBusStop.latitude + 0.5
+                longitude = closeBusStop.longitude + 0.5
+            }
+
+        val closeBusRoute =
```
boringcactus, line 1375: Comments on these routes explaining why they'll sort where they do (pinned, subway, no schedules) might help readability here.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
EmmaSimon: Yeah that's fair. It seemed wrong to assume that there were schedules by default, but that's what we're doing anyway before schedules load, so I'm fine with just having the default, and it definitely cuts down on the annoying test changes.

boringcactus (approved):

PR #421, opened 2024-09-23
fix: match schedules to predictions by parent

boringcactus:
### Summary

_Ticket:_ [Duplicate departure rows for Commuter Rail trips](https://app.asana.com/0/1205425564113216/1207114848419521/f)
[Slack thread](https://mbta.slack.com/archives/C05QMG9GS9M/p1727112428287779) from different symptoms of the same bug

Matching trips by parent station ID is so clearly the right thing to do that I have no idea why we ever put up with the `flatMap()` in the `RealtimePatterns` constructors.

### Testing

Added a unit test that would've caught both the new symptom and the old symptom.


JackVCurtis (approved):
👍

PR #422, opened 2024-09-24
fix: show cancelled trips in departure lists

BrandonTR:
### Summary

_Ticket:_ [Show explicitly cancelled bus, CR, ferry trips in departure lists](https://app.asana.com/0/1205425564113216/1207678190799459/f)

Show cancelled trips for non-subway stops when there are more than 2 departures. You'll see I added a lot of `@MainActor` annotations to tests, this was because SwiftUI previews would not work until I did so.

### Testing

Added some unit tests around new filtering and trip display formatting. Also added new previews to Android and iOS.

### DIFF
--- a/iosApp/iosAppTests/Pages/Map/HomeMapViewTest.swift
+++ b/iosApp/iosAppTests/Pages/Map/HomeMapViewTest.swift
@@ -19,7 +19,7 @@ final class HomeMapViewTest: XCTestCase {
         executionTimeAllowance = 60
     }
 
-    func testAppears() throws {
+    @MainActor func testAppears() throws {
         let viewportProvider: ViewportProvider = .init(viewport: .followPuck(zoom: 1))
         let sheetHeight: Binding<CGFloat> = .constant(100)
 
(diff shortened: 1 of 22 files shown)


iosApp/iosAppTests/Pages/Map/HomeMapViewTest.swift
```diff
@@ -19,7 +19,7 @@ final class HomeMapViewTest: XCTestCase {
         executionTimeAllowance = 60
     }
 
-    func testAppears() throws {
+    @MainActor func testAppears() throws {
```
JackVCurtis, line 22: Is there a reason related to this change that we need this macro now?

JackVCurtis (approved):
Looks solid, nothing blocking!

iosApp/iosAppTests/Pages/Map/HomeMapViewTest.swift
BrandonTR: I'm not sure. It might have been broken for a bit and we didn't notice. When running just the tests it doesn't seem to matter but SwiftUI previews seem to also compile the test suite and throw an error if this macro isn't there (even though it's not actually running the tests).

PR #424, opened 2024-09-24
fix: Don't sort disrupted routes without schedules to the bottom of nearby transit

EmmaSimon:
### Summary

_Ticket:_ Follow up fix for [Handle no service for the entire service day](https://app.asana.com/0/1205732265579288/1208217009015138/f)

In response to [this slack thread](https://mbta.slack.com/archives/C05QMG9GS9M/p1727187298728979), routes that have major disruptions should still show up in their usual place in nearby transit.

### Testing

Added a test for sorting behavior with alerts

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -572,13 +572,7 @@ fun NearbyStaticData.withRealtimeInfo(
             compareBy<StopsAssociated, Route>(Route.pinnedRoutesComparator(pinnedRoutes)) {
                     it.sortRoute()
                 }
-                .thenBy {
-                    it.patternsByStop.all { byStop ->
-                        byStop.patterns.all { patterns ->
-                            patterns.upcomingTrips.isNullOrEmpty() && !patterns.hasSchedulesToday
-                        }
-                    }
-                }
+                .thenBy { !it.hasSchedulesToday }
                 .thenBy(Route.subwayFirstComparator) { it.sortRoute() }
                 .thenBy { it.distanceFrom(sortByDistanceFrom) }
                 .thenBy { it.sortRoute() }
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -572,7 +572,7 @@ fun NearbyStaticData.withRealtimeInfo(
             compareBy<StopsAssociated, Route>(Route.pinnedRoutesComparator(pinnedRoutes)) {
                     it.sortRoute()
                 }
-                .thenBy { !it.hasSchedulesToday }
+                .thenBy { !it.hasServiceOrDisruptionToday }
                 .thenBy(Route.subwayFirstComparator) { it.sortRoute() }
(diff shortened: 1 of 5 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -572,13 +572,7 @@ fun NearbyStaticData.withRealtimeInfo(
...
                     it.sortRoute()
                 }
-                .thenBy {
-                    it.patternsByStop.all { byStop ->
-                        byStop.patterns.all { patterns ->
-                            patterns.upcomingTrips.isNullOrEmpty() && !patterns.hasSchedulesToday
-                        }
-                    }
-                }
+                .thenBy { !it.hasSchedulesToday }
```
KaylaBrady, line 575: nitpick (non-blocking): this name feels a bit of a misnomer - docs would help, maybe something like `it.ordinarilyWouldRunToday`

KaylaBrady (approved):
Nice! A few minor suggestions

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
```diff
@@ -23,6 +23,11 @@ sealed class RealtimePatterns : Comparable<RealtimePatterns> {
     abstract val alertsHere: List<Alert>?
     abstract val hasSchedulesToday: Boolean
 
+    val hasMajorAlerts
+        get() = run {
```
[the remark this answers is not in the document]
EmmaSimon: Since we can have computed properties in kotlin, I'm still a bit unclear on when and where a function is more or less appropriate. Lately I've been leaning more in the direction of, if it only uses data internal to the class and doesn't need any arguments, make it a computed prop. But if you think a function makes more sense, I wouldn't be opposed to that. I can do a test for sure though.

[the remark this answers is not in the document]
KaylaBrady: 👍 that reasoning makes sense to me! As the person with the least hands-on kotlin experience, happy to adopt that convention.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
EmmaSimon: Yeah that's fair. That's just the only use case we're using it in, and I don't have a great idea for a name to encompass all of these. I don't know how likely/possible it is, but I could see `ordinarilyWouldRunToday` also becoming a misnomer if there are no scheduled trips but for some reason we do start running service and have upcoming trips, then it ordinarily wouldn't be running service but abnormally are. Or if we have a disruption on a route with no weekend service, if the alert isn't specifically set to exclude the weekend, and this is true, that's not ordinarily running either. But I'm being pedantic and those are very contrived examples. 😅 Uhhh... Maybe, `hasServiceOrDisruptionToday`? Orrr.. for very vague but more literal, something like `hasNearbyDisplayPriority`?

KaylaBrady: Yeah I see what you mean 😃  I'd be fine with either of those two, I think the doc will be more important for conveying the nuance anyway.

KaylaBrady (approved):

PR #427, opened 2024-09-25
fix: goBack crash

BrandonTR:
### Summary

_Ticket:_ [Crash on goBack](URL)

This should at least prevent this specific crash. `popLast` can be called on empty collections while `removeLast` cannot without throwing an exception.

### Testing

No automated tests, functionally this should be identical. (Aside from the crashing)


iosApp/iosApp/Pages/NearbyTransit/NearbyViewModel.swift
```diff
@@ -63,7 +63,7 @@ class NearbyViewModel: ObservableObject {
     }
 
     func goBack() {
-        navigationStack.removeLast()
+        _ = navigationStack.popLast()
```
JackVCurtis, line 66: Where will the user be directed when the navigation stack is empty? Is it worth having a default route we use when the user triggers this behavior?

KaylaBrady, line 66: I finally encountered this in the wild from hitting the back button twice quickly! Excited for this fix 🎉 
Consider adding a test for calling goBack on an empty list to prevent a regression

BrandonTR: There kinda is a default route already, in `ContentView` line 163 we use `lastSafe()` which will return nearby route if there is no route.

KaylaBrady (approved):

PR #428, opened 2024-09-26
filter stop layers based on zoom level

JackVCurtis:
### Summary

_Ticket:_ [Hide stops based on map zoom](https://app.asana.com/0/1205732265579288/1208288117643285/f)

What is this PR for?

Hides the bus stop layers and eventually all stop layers as the user zooms out. 

### Testing

What testing have you done?

I've updated the shared tests, but we seem to be lacking test coverage for the map code I've updated. Is that worth filling in as part of this PR?

### DIFF
--- a/iosApp/iosApp/Pages/Map/MapLayerManager.swift
+++ b/iosApp/iosApp/Pages/Map/MapLayerManager.swift
@@ -15,7 +15,7 @@ protocol IMapLayerManager {
     var currentScheme: ColorScheme? { get }
     func addIcons(recreate: Bool)
     func addLayers(colorScheme: ColorScheme, recreate: Bool)
-
+    func updateVisibleLayers(colorScheme: ColorScheme)
     func updateSourceData(routeData: MapboxMaps.FeatureCollection)
     func updateSourceData(stopData: MapboxMaps.FeatureCollection)
 }
@@ -58,15 +58,10 @@ class MapLayerManager: IMapLayerManager {
     }
 
     func addLayers(colorScheme: ColorScheme, recreate: Bool = false) {
-        let colorPalette = switch colorScheme {
-        case .light: ColorPalette.companion.light
-        case .dark: ColorPalette.companion.dark
-        @unknown default: ColorPalette.companion.light
-        }
+        let colorPalette = getColorPalette(colorScheme: colorScheme)
         currentScheme = colorScheme
-        let layers: [MapboxMaps.Layer] = RouteLayerGenerator.shared.createAllRouteLayers(colorPalette: colorPalette)
-            .map { $0.toMapbox() }
-            + StopLayerGenerator.shared.createStopLayers(colorPalette: colorPalette).map { $0.toMapbox() }
+        let layers = generateLayers(colorScheme: colorScheme)
+
         for layer in layers {
             do {
                 if map.layerExists(withId: layer.id) {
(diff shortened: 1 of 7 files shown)


iosApp/iosApp/Pages/Map/MapLayerManager.swift
```diff
@@ -88,6 +84,49 @@ class MapLayerManager: IMapLayerManager {
         }
     }
 
+    func updateVisibleLayers(colorScheme: ColorScheme) {
+        let newLayers = generateLayers(colorScheme: colorScheme)
+        let newLayerIds = newLayers.map(\.id)
+        for layerIdentifier in map.allLayerIdentifiers {
+            do {
+                if layerIdentifier.id.contains(StopLayerGenerator.shared.stopLayerId),
```
JackVCurtis, line 92: I don't love this, it creates an implicit convention that all stop layers will include `stop-layer` in the ID. However since the transfer and alert layer ids are indexed, I couldn't see an easy way of identifying stop layers without doing some kind of pattern match

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/StopLayerGenerator.kt
```diff
@@ -9,6 +9,8 @@ import com.mbta.tid.mbta_app.map.style.downcastToColor
 object StopLayerGenerator {
     val maxTransferLayers = 3
     val stopZoomThreshold = 8.0
+    val showBusStopZoomThreshold = 12.0
+    val showAnyStopZoomThreshold = 11.0
```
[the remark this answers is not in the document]
JackVCurtis: Thanks for the correction, I really biffed the implementation on this one. I've updated the PR accordingly, thanks for the clear alternative!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/StopLayerGenerator.kt
```diff
@@ -38,14 +39,22 @@ object StopLayerGenerator {
                 Exp.image(Exp(""))
             )
         stopSelectedPinLayer.iconOffset = offsetPinValue()
-        includeSharedProps(stopSelectedPinLayer)
+        stopTouchTargetLayer.filter =
+            Exp.ge(
+                Exp.zoom(),
+                Exp.case(
+                    MapExp.isBusExp to Exp(busStopZoomThreshold + 1),
+                    Exp(stopZoomThreshold + 1)
```
boringcactus, line 47: These +1s feel counterintuitive to me - can you add a comment explaining why they're necessary?

boringcactus (commented):
I probably would've mentioned this approach somewhere on the task if I'd known you were going to be the one picking it up - I think all the rest of us have had enough map experience to already know that this approach is possible.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/StopLayerGenerator.kt
JackVCurtis: Double checked my logic and I think we don't need the + 1's at all - minZoom will hide the layer if zoom < threshold, and filter will show the layer if zoom >= threshold, so the threshold shouldn't need the adjustment

boringcactus (approved):

PR #429, opened 2024-09-26
feat: Add search input and overlay

EmmaSimon:
### Summary

_Ticket:_ [Search UI: style input](https://app.asana.com/0/1205732265579288/1208355161933385/f)

Add styling for the search input. I briefly looked into animating the text field to slide in and out when the cancel button appeared, but it was difficult enough that I stopped trying. It would also be good for the input to animate away when moving from nearby to stop details.

 Simulator Screenshot - iPhone 15 Pro - 2024-09-26 at 11 53 34 Simulator Screenshot - iPhone 15 Pro - 2024-09-26 at 11 35 08

### Testing

Manual testing of all behavior.
Unit test for search overlay component.


iosApp/iosApp/ComponentViews/ActionButton.swift
```diff
@@ -13,21 +13,24 @@ struct ActionButton: View {
     enum Kind {
         case back
         case close
+        case clear
```
KaylaBrady, line 16: question(non-blocking): It seems like ~most of the properties are different for the clear button from the other two - is there a benefit to keeping it an ActionButton? I wonder if there is a different composition of Views here that could help reusability, but definitely not blocking for this PR.

iosApp/iosApp/ContentView.swift
```diff
@@ -97,16 +97,19 @@ struct ContentView: View {
...
+            ZStack(alignment: .top) {
+                mapWithSheets
+                VStack(alignment: .trailing, spacing: 0) {
+                    if contentVM.searchEnabled, nearbyVM.navigationStack.lastSafe() == .nearby {
+                        SearchOverlay(searchObserver: searchObserver, nearbyVM: nearbyVM, searchVM: searchVM)
+                    }
+                    if !searchObserver.isSearching, !viewportProvider.viewport.isFollowing,
+                       locationDataManager.currentLocation != nil {
+                        RecenterButton { Task { viewportProvider.follow() } }
+                    }
```
KaylaBrady, line 110: question: Could this be kept within the HomeMapView if there is a new prop like `hideRecenter`? I think it would be nice to keep the recenter button localized with the rest of the map if possible

iosApp/iosApp/Pages/Search/SearchOverlay.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    var body: some View {
+        ZStack(alignment: .top) {
+            if searchObserver.isSearching {
+                Color.fill2
+            }
+            VStack {
+                SearchField(searchObserver: searchObserver)
+                if searchObserver.isSearching {
+                    SearchView(
```
KaylaBrady, line 42: suggestion (non-blocking): Looks like `SearchResultView` is already taken, but I think that would be an appropriate name here now that the query check has been pulled out of SearchView

iosApp/iosApp/Pages/Search/SearchResultView.swift
```diff
@@ -75,15 +59,11 @@ struct SearchView: View {
...
-        VStack {
-            if !query.isEmpty {
-                SearchResultView(
-                    results: searchResults,
-                    handleStopTap: handleStopTap,
-                    showRoutes: searchVM.routeResultsEnabled
-                )
-            }
-        }
+        SearchResultView(
```
KaylaBrady, line 62: question (non-blocking): Is there an accessibility label for this section that would be helpful to introduce it as search results?

iosApp/iosApp/Pages/Search/TextFieldObserver.swift
```diff
@@ -0,0 +1,33 @@
+//
```
KaylaBrady, line 1: question: this looks like a new file but I think it was just moved - is there an old file that needs to be deleted?

iosApp/iosAppTests/Views/SearchResultViewTests.swift
```diff
@@ -55,6 +55,61 @@ final class SearchResultViewTests: XCTestCase {
...
+        var sut = SearchOverlay(
+            searchObserver: searchObserver,
+            nearbyVM: NearbyViewModel(),
+            searchVM: SearchViewModel(),
+            searchResultsRepository: FakeRepo(getSearchResultsExpectation: getSearchResultsExpectation)
+        )
+
+        ViewHosting.host(view: sut)
+
+        // On init, only the search field should be displayed
```
KaylaBrady, line 84: praise: appreciate the comments here!

KaylaBrady (commented):
Looking good! Appreciate the re-organization of these views. A few minor thoughts!

iosApp/iosApp/ComponentViews/ActionButton.swift
EmmaSimon: Yeah, I felt weird about doing this, it's the exact same style as the clear button except half the size and different colors, so it seemed awkward to copy all of that into a new component, but it's also awkward to fit it into here. Hm, it might be ideal to add modifiers or props to let you recolor the circle and icon, then just resize the entire action button. Though then the accessibility label would also be off, I'm not sure if adding another label higher up overrides any lower down?

iosApp/iosApp/ContentView.swift
EmmaSimon: No, I don't think so. It needs to be in this VStack so that it will layout properly below the search bar when it exists, or in the top corner when search is gone. I think it makes sense separate since it is a map overlay. If we have other buttons or need to clean this up in the future, I think having a general map overlay view which includes search and any buttons could make sense.

iosApp/iosApp/Pages/Search/SearchOverlay.swift
EmmaSimon: I was trying to avoid significant changes to anything in `SearchResultView.swift` because @BrandonTR and I started working on the result styling changes in parallel. I agree, either I could do this and possibly cause merge conflicts, or Brandon could handle it?

iosApp/iosApp/Pages/Search/TextFieldObserver.swift
EmmaSimon: It was pulled out of `SearchResultView.swift`, there's still a handful of different views in there, and it used to also include this.

iosApp/iosApp/Pages/Search/SearchOverlay.swift
BrandonTR: I don't mind taking that on!

KaylaBrady (approved):

PR #434, opened 2024-09-30
refactor: flatten ErrorDetails into ApiResult.Error

boringcactus:
### Summary

If we create `ApiResult.Error`s for every failure and not just non-success HTTP responses from the backend, we will benefit from it being more straightforward to create `ApiResult.Error`s from things that are not HTTP responses.

### Testing

Ran unit tests.


KaylaBrady (approved):
Nice!

PR #435, opened 2024-09-30
refactor: replace Outcome with ApiResult

boringcactus:
### Summary

Stacked on #434. Joins our two error types into one.

### Testing

Ran the unit tests and checked both iOS and Android apps for surface-level functionality.


iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -82,8 +82,8 @@ extension HomeMapView {
     func joinVehiclesChannel(routeId: String, directionId: Int32) {
         leaveVehiclesChannel()
         vehiclesRepository.connect(routeId: routeId, directionId: directionId) { outcome in
-            if let data = outcome.data {
-                vehiclesData = Array(data.vehicles.values)
+            if case let .ok(data) = onEnum(of: outcome) {
+                vehiclesData = Array(data.data.vehicles.values)
```
EmmaSimon, line 86: Sorry for the nitpick, but all the `data.data`s here and throughout make me sad and are a bit unreadable. Could we do something like `result.data` instead?

EmmaSimon (commented):
Thanks for doing this refactor 🙏

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
boringcactus: Turned every `data.data` into a `result.data`.

EmmaSimon (approved):

PR #436, opened 2024-09-30
refactor: use ApiResult for all static data fetching

boringcactus:
### Summary

Stacked on #435. Moves all static data fetching to return `ApiResult`s rather than throw errors if the data is unavailable. Also changes the default HTTP client behavior to verify that response codes indicate success.

### Testing

Updated unit tests and checked iOS and Android apps for surface-level correctness. I have not added unit tests for `ApiResult.runCatching` - I can if it'd be worth it.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
```diff
@@ -119,26 +119,47 @@ class ResponseCache<T>(
...
 
-            val httpResponse = fetch(this.data?.metadata?.etag)
-            return when (httpResponse.status) {
-                HttpStatusCode.NotModified -> {
-                    val data = this.data ?: throw RuntimeException("Failed to update cached data")
+            try {
+                val httpResponse = fetch(this.data?.metadata?.etag)
+                return when (httpResponse.status) {
+                    HttpStatusCode.NotModified -> {
+                        // since MobileAppClient defaults to expectSuccess = true, this shouldn't
```
JackVCurtis, line 133: Sorry, not quite following this comment. What's the behavior of expectSuccess? Why are we throwing a RedirectResponseException if this block shouldn't happen?

boringcactus: When [`expectSuccess`](https://ktor.io/docs/client-response-validation.html#default) is `true`, Ktor will throw an exception if the HTTP response is not 2xx, which is why the 304 handling has been moved to the `catch` block. However, if in the future for some reason we explicitly set `expectSuccess = false` in a call to `ResponseCache.getOrFetch`, it would be nice if `ResponseCache.getOrFetch` still behaved correctly, and throwing the exception here that would've been thrown as a result of `expectSuccess = true` seemed like the easiest way to do that.

KaylaBrady: suggestion (non-blocking): Instead of setting `expectSuccess` to true, could we handle 304 as a normal response with [custom validation](https://ktor.io/docs/client-response-validation.html#custom)? It seems somewhat unintuitive for an expected 304 response to be caught as an exception.

KaylaBrady (approved):
👍 one small suggestion, but doesn't need to block this PR & the error banner work stacked on top of fit.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
boringcactus: Ah, yeah, that makes more sense.

KaylaBrady (approved):

PR #423, opened 2024-09-24
feat: Use new predictions V2 channel behind feature flag

KaylaBrady:
### Summary

_Ticket:_ [Predictions Scalability: use new predictions channel](https://app.asana.com/0/1205732265579288/1207791938443981/f)

What is this PR for?

This switches nearby transit & stop details to use the new predictions channel, currently behind a cutover. Temporarily pointing to dev-orange to pick up https://github.com/mbta/mobile_app_backend/pull/203, will remove that before merging.

### Testing

What testing have you done?

Added a few unit tests & ran locally to confirm predictions continued to update as expected.

PR #437, opened 2024-10-01
fix: iOS 16 sheet navigation crash fix

KaylaBrady:
### Summary

_Ticket:_ [Fix iOS 16 sheet bug](https://app.asana.com/0/1205732265579288/1208295417175128/f)

What is this PR for?

This PR resolves the iOS 16 sheet bug by maintaining a single sheet whose contents change when navigation changes rather than opening / closing a new sheet for each navigation change.

After speaking with design, navigating within the same view type (stop details => stop details, trip details => trip details) is treated as a change within the same view rather than opening a new view to avoid race conditions between the opening and closing views. 

## Note
* When using the existing predictions for stop channel, this maintains an existing race condition around leaving / joining predictions when navigating from nearby transit => stop details. That race condition will be resolved by the switchover to use [the new predictions channel](https://github.com/mbta/mobile_app/pull/423) with unique topic names. 

### Testing

What testing have you done?

Added unit tests for the onChange behavior in stop & trip details. Ran locally & in simulator with slow animations turned on (I turned it off partway in this video because I got impatient)

https://github.com/user-attachments/assets/c2e5a997-db4e-4640-92fd-946ed8d75615


iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -74,28 +74,44 @@ struct StopDetailsPage: View {
...
+
+                updateDepartures(stop)
+            }
+            .onChange(of: schedulesResponse) { _ in
+
+                updateDepartures(stop)
+            }
             .onReceive(inspection.notice) { inspection.visit(self, $0) }
-            .task {
+            .task(id: stop.id) {
```
KaylaBrady, line 103: This was surprising to me: without passing `stop.id` here, the contained call to `updateDepartures()` was always using the original stop from when the view was created, resulting in the old stop pills flashing in.

iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -134,6 +149,11 @@ struct StopDetailsPage: View {
     }
 
     func changeStop(_ stop: Stop) {
+        leavePredictions()
+        fetchStopData(stop)
```
KaylaBrady, line 153: I could also update the call to `connect` in the repository to leave the channel first. We haven't consistently taken that approach, but it might help to standardize on it.

EmmaSimon: Are there any downsides to doing this? It seems like it makes sense for the repo to handle it.

KaylaBrady: I guess that is the approach that the predictions V2 channel takes, so changing it for V1 doesn't really matter, depending on when this change lands relative to the V2 cutover.

EmmaSimon: Ahh, gotcha, yeah if it's not relevant for v2 then I wouldn't bother.

iosApp/iosApp/ContentView.swift
```diff
@@ -157,108 +160,118 @@ struct ContentView: View {
...
-                    visibleNearbySheet = entry
-                    screenTracker.track(
-                        screen: filtered ? .stopDetailsFiltered : .stopDetailsUnfiltered
-                    )
-                }
+                case let .stopDetails(stop, filter):
+                    // Wrapping in a TabView helps the page to animate in as a single unit
+                    // Otherwise only the header animates
+                    TabView {
+                        StopDetailsPage(
```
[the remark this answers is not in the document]
KaylaBrady: @EmmaSimon These issues should be resolved now!

I took the approach of each stop details page (outgoing & incoming) owning their own filter and departure state and updating the shared state when either of these values change _only if that stop details page is the incoming one_. This doesn't feel totally ideal to me but seems like the most reliable way to ensure there is no shared data between the two pages.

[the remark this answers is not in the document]
KaylaBrady: Actually looking into an issue this introduced when tapping on a row from unfiltered stop details 👀 but I expect this general approach to still work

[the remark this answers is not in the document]
EmmaSimon: Nice, the current approach is solid, this all looks good! I did notice a few transition animation quirks now that I want to mention, but this is definitely something we can release now and tweak later with design input.
* Transitions between different stop pages on the map have the slide up animation, while transitions between vehicles don't really seem to fade, just pop to the new vehicle.
* Updating the stop filter now has a fade transition (and even a lil slide up/down of the direction picker if the size changes). I don't hate this, but not sure if it was intentional. It's also a bit funky because of how the components are colored, the route header has its own colored background in addition to the colored background fill, which becomes visible during the fade.

EmmaSimon (approved):

iosApp/iosApp/ContentView.swift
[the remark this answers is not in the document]
KaylaBrady: Thanks! I spoke with design about the lack of animation between trip details pages - I avoided implementing it on this pass since supporting slide would involve resolving [other race condition issues](https://mbta.slack.com/archives/C062QNAJZ2M/p1727791334641539?thread_ts=1727789603.100029&cid=C062QNAJZ2M), and it is a lower priority page. I'll definitely make a ticket for it! 

The stop filter animation was an unintentional side effect a side effect of breaking the relationship between stop details page and the last filter binding, but it is actually very straightforward to remove so I can do that now.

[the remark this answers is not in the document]
EmmaSimon: I'm totally in favor of not doing anything to get trip-trip animations working before we get rid of the separate trip page entirely, but it's probably still good to have a ticket somewhere marking it as known. Thanks!

iosApp/iosApp.xcodeproj/project.pbxproj
```diff
@@ -1551,7 +1551,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 16.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 16.4;
```
KaylaBrady, line 1554: spoke with product IRL after raising [in slack ](https://mbta.slack.com/archives/C062QNAJZ2M/p1727883523066669) - opting to bump the min version.

PR #438, opened 2024-10-01
feat: Save visited stops

EmmaSimon:
### Summary

_Ticket:_ [Search UI: store recently visited stops](https://app.asana.com/0/1205732265579288/1208355161933389/f)

We want to show recently visited stops in the empty state of the search overlay. This records all visits to stop pages and persists them to disk, along with displaying a barebones, non-interactive and unstyled list of stop names on the empty search page.

### Testing

Added tests for the usecase (which also covers testing the repository and the VisitHistory class), along with a swift test asserting that updates to the nav stack result in saving visits.


iosApp/iosApp/Pages/NearbyTransit/NearbyViewModel.swift
```diff
@@ -20,12 +22,28 @@ class NearbyViewModel: ObservableObject {
...
+        didSet { Task {
+            let navEntry = navigationStack.lastSafe()
+            do {
+                switch navEntry {
+                case let .stopDetails(stop, _):
+                    try await visitHistoryUsecase.addVisit(visit: .StopVisit(stopId: stop.id))
+                default: break
+                }
+            } catch {
+                let navLabel = navEntry.sheetItemIdentifiable().debugDescription
```
JackVCurtis, line 35: Will this make the raw error description visible to the user?

JackVCurtis (approved):
One question about how we want to handle an error case, but otherwise great!

iosApp/iosApp/Pages/NearbyTransit/NearbyViewModel.swift
EmmaSimon: Oh, nope this is just a warn log for us, nothing that would be visible to users. Unless they tried really hard to read their system logs lol

PR #441, opened 2024-10-02
feat: Prevent map zooming when navigating between stop details and nearby transit

EmmaSimon:
### Summary

_Ticket:_ [Don't zoom map when navigating to Nearby Transit or Stop Details](https://app.asana.com/0/1205732265579288/1208389938532288/f)

There was too much map movement when transitioning between nearby transit, stop details, and trip details. We don't want to zoom in and out so much on every page transition. This change makes it so that the camera simply pans at the current zoom level to the selected stop on the stop page, or the previous location in nearby transit.

### Testing

Updated existing tests of this behavior to also assert zoom level, and added a new test for returning to a saved camera state.


JackVCurtis (approved):
Looks good!

PR #445, opened 2024-10-07
refactor: centralize route pattern sorting

boringcactus:
### Summary

_Ticket:_ [Handle service ended for the day](https://app.asana.com/0/1205425564113216/1208217009015140/f)

Centralizes route pattern sorting so we get consistent behavior at each layer and so it's easier to add or change the sort criteria.

### Testing

Checked that the one existing unit test where logic is changing is producing more correct behavior (sorting an alert pattern above a no-service pattern within a stop, which previously wasn't happening).


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternSorting.kt
```diff
@@ -0,0 +1,83 @@
...
+            {
+                when (it) {
+                    is NearbyStaticData.StaticPatterns.ByDirection -> -1
+                    is NearbyStaticData.StaticPatterns.ByHeadsign -> 1
+                }
+            },
+            { it.patterns.first() }
+        )
+
+    fun comparePatterns(): Comparator<RealtimePatterns> =
```
KaylaBrady, line 40: nitpick (non-blocking): consider calling this `compareRealtimePatterns` for consistency with `compare static patterns`

KaylaBrady (approved):
Great approach!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternSorting.kt
boringcactus: Renamed.

PR #454, opened 2024-10-08
fix: Prevent Government Center direction from displaying

EmmaSimon:
### Summary

_Ticket:_ [GL showing duplicate Eastbound @ Gov Center with no data](https://app.asana.com/0/1205732265579288/1208378317839446/f)

We got a couple bug reports about a "Government Center" direction showing up at Gov Ctr and further eastbound stops. This was happening because `StopPatterns.ForLine` was only creating directions based on the typical pattern for each route, which meant that any time there was a trip on an atypical route like GL C to Union Square, it was being displayed with the typical "Government Center" direction label. This label should never be displayed because it should always be overridden with a special case, and atypical trips should have direction labels based on their actual destinations, not the typical one.

| Before | After |
| - | - |
| image (3) | Simulator Screenshot - iPhone 15 Pro - 2024-10-07 at 21 47 18 |

### Testing

Added a test to ensure that the "Government Center" direction is removed in the static data creation. Also saw a handful of cases live where an atypical C train going to and from Union was displayed under the correct grouped direction label.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -344,7 +372,16 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
...
                 routes = routes,
                 stop = stop,
-                patterns = linePatterns,
+                patterns =
+                    linePatterns.filter {
+                        when (it) {
+                            is StaticPatterns.ByDirection ->
+                                // Remove all directions terminating mid-line at Gov Center,
+                                // this direction label should never be displayed.
+                                it.direction.destination != "Government Center"
```
boringcactus, line 381: I'm confused. Shouldn't even the patterns that end at Government Center be sorted into a "Gov Ctr & North" direction?

EmmaSimon: Yeah, but by this point they will have already been sorted into the overridden "Gov Ctr & North" direction. The bare "Government Center" direction _only_ shows up when the direction has not been overridden by a grouped direction, which can only occur at or after the Gov Center stop. It's normal for it to happen at Gov Center for C and B trains, since the list of served stop IDs don't include any upstream stops that would override their default destination. After these changes, I think that should be the only place it occurs.

boringcactus: Ah, if it's the presence of downstream stops that causes the override to kick in then that'd do it. It may be worth clarifying in the comment that trips continuing past Government Center will be grouped into "Gov Ctr & North" even at Government Center but trips terminating at Government Center will still say Government Center at Government Center. That's still confusing, because at Government Center shouldn't the trips that terminate at Government Center be hidden for being arrival-only?

EmmaSimon: Yes, the trips are hidden, but this resolution is happening in the static data, so until we load the actual upcoming trips, we don't know whether or not there will be trips or if they should be hidden. And since these are typical patterns, this was causing the "Government Center" direction to flash in with "No realtime data" before any trips were loaded, at which point it would disappear. But yeah I'll add to the comment.

boringcactus: Ah, that makes sense, since we know things are arrival-only by the schedules or the predictions. I think we could also determine that things are arrival-only in static data by checking the stop lists of representative trips, but if that's not simpler than this then there's no point doing that instead.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/NearbyResponseTest.kt
```diff
@@ -509,6 +509,211 @@ class NearbyResponseTest {
...
+
+                typicality = RoutePattern.Typicality.Typical
+                directionId = 1
+                sortOrder = 2
+            }
+
+        val global =
+            GlobalResponse(
+                objects,
+                patternIdsByStop =
```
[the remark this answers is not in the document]
EmmaSimon: There might be a more concise way to do it with some arcane kotlin wizardry, but I think I've got it working in the latest change. I only updated this test though, not sure if there are any specific existing tests you had in mind, but I didn't seek out any other places that could be retrofitted to use it.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
EmmaSimon: I took a shot at expanding the comment, it's uhhhh, pretty verbose and potentially incoherent, but is hopefully more helpful.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/GlobalResponse.kt
```diff
@@ -21,6 +21,25 @@ data class GlobalResponse(
...
+        objects.lines,
+        objects.routePatterns
+            .flatMap {
+                objects.trips[it.value.representativeTripId]?.stopIds?.map { stopId ->
+                    stopId to it.key
+                }
+                    ?: emptyList()
+            }
+            .groupBy { it.first }
+            .mapValues { it.value.map { stopAndPattern -> stopAndPattern.second } },
```
boringcactus, line 36:
```suggestion
            .groupBy({ it.first }, { it.second }),
```

The Kotlin standard library is [pretty nice](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/group-by.html). (The Elixir standard library [also has this](https://hexdocs.pm/elixir/Enum.html#group_by/3).)

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/NearbyResponseTest.kt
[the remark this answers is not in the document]
boringcactus: I may toss a follow-up PR together to do this everywhere - that certainly doesn't need to be your problem here.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
boringcactus: I think that helps.

PR #455, opened 2024-10-08
fix: show loading instead of service ended when partially loaded

boringcactus:
### Summary

_Ticket:_ [Handle service ended for the day](https://app.asana.com/0/1205732265579288/1208217009015140/f), kinda
[Slack thread](https://mbta.slack.com/archives/C05QMG9GS9M/p1728334936690569)

If predictions load before schedules, hasSchedulesToday defaults to true while schedules are loading, so we'll say "Service ended" in places where there are no predictions. Making that default to null won't help if schedules load before predictions; we'll still flash "predictions unavailable" on subway. The most direct way to ensure that we don't make claims about absent service until all the data has loaded is to keep track directly of whether or not all the data has loaded.

### Testing

Manually verified that everything looks like it's working correctly, updated unit tests, added a new unit test.


JackVCurtis (approved):
👍 I like the approach, very straightforward.

KaylaBrady (approved):

PR #456, opened 2024-10-09
feat: search results styling

BrandonTR:
### Summary

_Ticket:_ [Search UI: style results](https://app.asana.com/0/1205425564113216/1208355161933387/f)

Search result styling.

### Testing

Added unit tests for new RoutePillSpec changes, fixed and added new tests for refactored SearchResultsContainer/View.


iosApp/iosApp/Pages/Search/SearchResultsContainer.swift
```diff
@@ -0,0 +1,382 @@
+//
```
JackVCurtis, line 1: This is a lot of top-level structs for a single file, I think. What do you think of breaking this up a bit?

JackVCurtis (approved):
👍 Could use a little more organization but nothing blocking

iosApp/iosApp/Pages/Search/SearchResultsContainer.swift
BrandonTR: I agree! I'll go ahead and break it up.

PR #457, opened 2024-10-09
chore: upgrade versions to support xcode 16

KaylaBrady:
### Summary

No ticket.

What is this PR for?

This bumps dependency versions to be compatible with Xcode 16 ([slack thread](https://mbta.slack.com/archives/C062QNAJZ2M/p1728488271210059)).

### Testing

What testing have you done?

Confirmed I was able to build & run the app locally. However, I do now see a warning: `empty dSYM file detected, dSYM was created with an executable with no debug info.`. 

There is a [developer forum thread](https://developer.apple.com/forums/thread/761589) on this new warning, with Sentry 8.36.0 specifically mentioned.

Apple developer response:

> This is a new warning in Xcode 16 where if you're missing a dSYM, Xcode lets you know in case you aren't expecting that — this is to reduce scenarios where you won't are able to fully symbolicating frames coming from your app's code in crash reports. In some cases, you may not have a dSYM, such as a closed source library from a vendor, and in those cases, the warning is safe to discard.

I built an archive locally and confirmed dSYM files are present in the build.
image


boringcactus (approved):

KaylaBrady:
There are some ongoing XCode 16 issues with ViewInspector that aren't trivially resolved, including `callOnChange` failing where it wouldn't before (`ProxyModifiedMap does not have 'onChange' modifier`). Relevant issue: https://github.com/nalexn/ViewInspector/issues/327

Going to hold off on trying to further complete this upgrade for now.

KaylaBrady (commented):
@boringcactus your changes look great to me, thanks for getting this over the line! I'm not able to formally approve since I started the PR, but your changes get a ✅  from me!

PR #459, opened 2024-10-10
Jvc trip details screen reader text

JackVCurtis:
### Summary

_Ticket:_ [Optimize Trip Details screen reader text](https://app.asana.com/0/1205732265579288/1208295417175124/f)

What is this PR for?

Updates the Trip Details page to use accessible elements and headings for improved VoiceOver access.

### Testing

What testing have you done?

Manual testing in the Accessibility inspector. Unfortunately I can't find a way to test accessibility labels and elements through the unit tests.

### DIFF
--- a/iosApp/iosApp/ComponentViews/SheetHeader.swift
+++ b/iosApp/iosApp/ComponentViews/SheetHeader.swift
@@ -23,6 +23,7 @@ struct SheetHeader: View {
                 Text(title)
                     .font(Typography.title3Semibold)
                     .padding([.top], 1)
+                    .accessibilityAddTraits(.isHeader)
                     .accessibilityHeading(.h1)
             }
             if let onClose {
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
@@ -28,7 +28,7 @@ struct TripDetailsHeader: View {
                 .accessibilityElement()
                 .accessibilityAddTraits(.isHeader)
                 .accessibilityHeading(.h1)
-                .accessibilityLabel("\(line?.shortName ?? "") \(route.type.typeText(isOnly: true)) to \(trip.headsign)")
+                .accessibilityLabel("\(route.label) \(route.type.typeText(isOnly: true)) to \(trip.headsign)")
             }
             Spacer()
             ActionButton(kind: .close, action: onClose)
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
@@ -47,7 +47,6 @@ struct TripDetailsHeader: View {
         }
         .accessibilityElement()
         .accessibilityAddTraits(.isHeader)
-        .accessibilityHeading(.h1)
         .accessibilityLabel("Real-time arrivals updating live")
     }
 
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
@@ -25,6 +25,10 @@ struct TripDetailsHeader: View {
                     RoutePill(route: route, line: line, type: .fixed)
                     toHeadsign(trip.headsign)
                 }
+                .accessibilityElement()
+                .accessibilityAddTraits(.isHeader)
+                .accessibilityHeading(.h1)
+                .accessibilityLabel("\(line?.shortName ?? "") \(route.type.typeText(isOnly: true)) to \(trip.headsign)")
             }
             Spacer()
             ActionButton(kind: .close, action: onClose)
@@ -41,12 +45,15 @@ struct TripDetailsHeader: View {
             Text("Live")
                 .font(Typography.footnote)
         }
+        .accessibilityElement()
+        .accessibilityAddTraits(.isHeader)
+        .accessibilityHeading(.h1)
+        .accessibilityLabel("Real-time arrivals updating live")
     }
 
     func toHeadsign(_ headsign: String) -> some View {
         Text("to \(headsign)")
             .font(Typography.headlineBold)
-            .accessibilityHeading(.h1)
     }
 }
 
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
@@ -24,16 +24,14 @@ struct TripDetailsStopListSplitView: View {
                     }
                 }
                 .accessibilityElement()
-                .accessibilityAddTraits(/*@START_MENU_TOKEN@*/ .isHeader/*@END_MENU_TOKEN@*/)
+                .accessibilityAddTraits(.isHeader)
                 .accessibilityHeading(.h2)
                 .accessibilityLabel(
                     LocalizedStringKey(
                         "\(routeType?.typeText(isOnly: true) ?? "") is \(splitStops.collapsedStops.count, specifier: "%ld") stops away from \(splitStops.targetStop.stop.name)"
                     )
                 )
-                .accessibilityHint(
-                    splitStops.collapsedStops.map(\.stop.name).joined(separator: ", ")
-                )
+                .accessibilityHint("List remaining stops")
             }
             TripDetailsStopView(stop: splitStops.targetStop, now: now, onTapLink: onTapLink, routeType: routeType)
                 .listRowBackground(Color.keyInverse.opacity(0.15))
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
@@ -23,6 +23,17 @@ struct TripDetailsStopListSplitView: View {
                         TripDetailsStopView(stop: stop, now: now, onTapLink: onTapLink, routeType: routeType)
                     }
                 }
+                .accessibilityElement()
+                .accessibilityAddTraits(/*@START_MENU_TOKEN@*/ .isHeader/*@END_MENU_TOKEN@*/)
+                .accessibilityHeading(.h2)
+                .accessibilityLabel(
+                    LocalizedStringKey(
+                        "\(routeType?.typeText(isOnly: true) ?? "") is \(splitStops.collapsedStops.count, specifier: "%ld") stops away from \(splitStops.targetStop.stop.name)"
+                    )
+                )
+                .accessibilityHint(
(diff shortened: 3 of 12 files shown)


iosApp/iosApp/ComponentViews/SheetHeader.swift
```diff
@@ -23,6 +23,7 @@ struct SheetHeader: View {
                 Text(title)
                     .font(Typography.title3Semibold)
                     .padding([.top], 1)
+                    .accessibilityAddTraits(.isHeader)
```
KaylaBrady, line 26: 👍 great find on this!

iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
```diff
@@ -23,6 +23,17 @@ struct TripDetailsStopListSplitView: View {
                         TripDetailsStopView(stop: stop, now: now, onTapLink: onTapLink, routeType: routeType)
                     }
                 }
+                .accessibilityElement()
+                .accessibilityAddTraits(/*@START_MENU_TOKEN@*/ .isHeader/*@END_MENU_TOKEN@*/)
```
KaylaBrady, line 27: nitpick: some tokens slipped in

iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
```diff
@@ -41,12 +45,15 @@ struct TripDetailsHeader: View {
             Text("Live")
                 .font(Typography.footnote)
         }
+        .accessibilityElement()
+        .accessibilityAddTraits(.isHeader)
+        .accessibilityHeading(.h1)
```
KaylaBrady, line 50: question: Is h1 necessary here? My expectation from web headings is that there is [typically only one h1 on a page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page).

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
```diff
@@ -26,12 +26,36 @@ struct TripDetailsStopView: View {
                         Spacer()
                         UpcomingTripView(prediction: upcomingTripViewState, routeType: routeType)
                     }
+                    .accessibilityElement()
```
KaylaBrady, line 29: suggestion: If `.accessibilityElement` is removed here, or used with [.combine](https://developer.apple.com/documentation/swiftui/accessibilitychildbehavior/combine), then I think this reads as-intended without needing to specify a separate `accessibilityLabel` since the inner components of the view already have accessibility labels.

The connection piece of the accessibilityLabel could be added directly to `scrollRoutes`.

iosApp/iosApp/Pages/TripDetails/VehicleCardView.swift
```diff
@@ -78,10 +78,18 @@ struct VehicleOnTripView: View {
                     .font(Typography.headlineBold)
                     .foregroundColor(textColor)
             }
+            .accessibilityElement()
+            .accessibilityAddTraits(.isHeader)
+            .accessibilityHeading(.h2)
+            .accessibilityLabel(
+                "\(route.type.typeText(isOnly: true)) \(vehicleStatusDescription(vehicle.currentStatus)) \(stop.name)"
+            )
```
KaylaBrady, line 86: suggestion: if `accessibilityElement` is used with [.combine](https://developer.apple.com/documentation/swiftui/accessibilitychildbehavior/combine), then the vehicle type accessibility label can be moved to the vehicle icon and the status + stop name will be read as specified by those views. As-is, the vehicle icon is read out loud as "small subway image"

iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
```diff
@@ -23,6 +23,17 @@ struct TripDetailsStopListSplitView: View {
...
+                .accessibilityHeading(.h2)
+                .accessibilityLabel(
+                    LocalizedStringKey(
+                        "\(routeType?.typeText(isOnly: true) ?? "") is \(splitStops.collapsedStops.count, specifier: "%ld") stops away from \(splitStops.targetStop.stop.name)"
+                    )
+                )
+                .accessibilityHint(
+                    splitStops.collapsedStops.map(\.stop.name).joined(separator: ", ")
```
KaylaBrady, line 35: suggestion: this may be worth confirming with design, but I'd expect this to read "Lists remaining stops" rather than the list of remaining stops. The [hint](https://developer.apple.com/documentation/swiftui/view/accessibilityhint(_:)-3rdgs) is intended to describe to users what would happen if they were to click the button.

KaylaBrady (commented):
This is headed in the right direction! Added a few suggestions about relying on the existing child view accessibility labels where possible. I'd be happy to do a huddle on this to go over any questions synchronously!

iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
JackVCurtis: Aaaaagh. Thank you.

iosApp/iosApp/Pages/TripDetails/TripDetailsHeader.swift
JackVCurtis: Yup, I think this was a copy/paste error that grabbed too many lines.

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
JackVCurtis: I had that approach initially but [this post ](https://www.hackingwithswift.com/books/ios-swiftui/hiding-and-grouping-accessibility-data) told me that the `.combine` option will cause the text to be read with pauses between elements, rather than as a single statement. I figured using a separate label would be a smoother experience.

iosApp/iosApp/Pages/TripDetails/TripDetailsStopListSplitView.swift
JackVCurtis: Yeah that makes sense. I don't think we need to check with design, looking at the Figma I think I mistook the literal text for instructions on what that text should be.

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
KaylaBrady: Ah got it, thanks for that resource! I'm not sure if that is necessary in this case since there is already punctuation separating each constituent element in this custom accessibility Label - comma after stop name, period after arriving at. 

However, if this does all need to be one label, I think it would be worth using the logic from `UpcomingTripAccessibilityFormatters` for the time piece since that encompasses more scenarios than "arriving at"

iosApp/iosApp/Pages/TripDetails/VehicleCardView.swift
KaylaBrady: I see how your point above about creating a new combined label applies in this case, but something funky seems to be going on - using voice over I hear "train underscore conditional content..."

JackVCurtis: Hmm, thanks for that. I'll look into it.

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
```diff
@@ -26,9 +26,28 @@ struct TripDetailsStopView: View {
                         Spacer()
                         UpcomingTripView(prediction: upcomingTripViewState, routeType: routeType)
                     }
+                    .accessibilityElement(children: .combine)
+                    .accessibilityAddTraits(.isHeader)
+                    .accessibilityHeading(.h2)
                 }
             )
             scrollRoutes
+                .accessibilityLabel(scrollRoutesAccessibilityLabel)
```
KaylaBrady, line 35: suggestion: I think this should be marked with `accessibilityElement()` as well - after the accessibility label is read, each route number is read again.

If this could be moved into the SheetNavigationLink label as well, I think that would be preferable so that it is all read together. If that is presents other challenges though then I think this is OK for now!

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
```diff
@@ -26,9 +26,28 @@ struct TripDetailsStopView: View {
...
+
+    var scrollRoutesAccessibilityLabel: String {
+        if stop.routes.isEmpty {
+            return ""
+        } else if stop.routes.count == 1 {
+            return "Connection to \(stop.routes.first!.shortName) \(stop.routes.first!.type.typeText(isOnly: true))"
+        } else {
+            let lastConnection = stop.routes.last
+            return stop.routes.prefix(stop.routes.count - 1).map {
+                "\($0.shortName) \($0.type.typeText(isOnly: true))"
```
KaylaBrady, line 47: suggestion: use `$0.label` here as well, and include "Connections to"

iosApp/iosApp/Pages/TripDetails/VehicleCardView.swift
KaylaBrady, line 1: suggestion: mark the routeVehicle as `.accessibilityHidden(true)` so that the image information is not read aloud

KaylaBrady (commented):
Nice improvements! Main outstanding comment is about switching the route label field, others can be addressed separately / as follow ups if needed!

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
KaylaBrady: Looks like the only piece left is including the "Connections to " prefix when there are plural connections

KaylaBrady (approved):
Sounds great! Just one small piece left for the connections list

iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift
```diff
@@ -26,9 +26,29 @@ struct TripDetailsStopView: View {
...
+            return ""
+        } else if stop.routes.count == 1 {
+            return "Connection to \(stop.routes.first!.label) \(stop.routes.first!.type.typeText(isOnly: true))"
+        } else {
+            let lastConnection = stop.routes.last
+            return stop.routes.prefix(stop.routes.count - 1).map {
+                "\($0.label) \($0.type.typeText(isOnly: true))"
+            }
+            .joined(separator: ", ") +
+            "and \(lastConnection!.label) \(lastConnection!.type.typeText(isOnly: true))"
```
KaylaBrady, line 51: Oh I missed this: I think we'll want a space before "and ...", otherwise it reads "bus and" like one word

EmmaSimon:
Just fyi, I resolved the test failures happening here in #472, so don't worry about fixing them.

PR #460, opened 2024-10-10
fix: Display GL vehicles on filtered stop page

EmmaSimon:
### Summary

_Ticket:_ [Green Line vehicles don't load in filtered Stop Details](https://app.asana.com/0/1205732265579288/1207813918309501/f)

We needed to be able to load vehicles for multiple routes in order to display GL vehicles before, which wasn't possible on the backend before https://github.com/mbta/mobile_app_backend/pull/214. This switches the vehicles channel to use that new multi-vehicle topic, and adds some additional filtering to only display the vehicles on routes which match the displayed route shapes on the map. 

Simulator Screenshot - iPhone 15 Pro - 2024-10-10 at 11 00 22

### Testing

Added unit tests in shared code.

### DIFF
--- a/iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
+++ b/iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
@@ -101,7 +101,9 @@ extension HomeMapView {
         leaveVehiclesChannel()
         vehiclesRepository.connect(routeId: routeId, directionId: directionId) { outcome in
             if case let .ok(result) = onEnum(of: outcome) {
-                vehiclesData = Array(result.data.vehicles.values)
+                if let departures = nearbyVM.departures {
+                    vehiclesData = Array(departures.filterVehiclesByUpcoming(vehicles: result.data).values)
+                }
             }
         }
     }
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
@@ -4,6 +4,7 @@ import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.ScheduleResponse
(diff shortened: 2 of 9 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
```diff
@@ -88,6 +89,15 @@ data class StopDetailsDepartures(val routes: List<PatternsByStop>) {
...
     )
 
+    val allUpcomingTrips = routes.flatMap { it.allUpcomingTrips() }
+
+    val upcomingPatternIds = allUpcomingTrips.mapNotNull { it.trip.routePatternId }.toSet()
+
+    fun filterVehiclesByUpcoming(vehicles: VehiclesStreamDataResponse): Map<String, Vehicle> {
+        val routeIds = allUpcomingTrips.map { it.trip.routeId }.toSet()
+        return vehicles.vehicles.filter { routeIds.contains(it.value.routeId) }
+    }
```
KaylaBrady, line 99: praise: I like the refactor to move & add this logic to StopDetailsDepartures!

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -101,7 +101,9 @@ extension HomeMapView {
         leaveVehiclesChannel()
         vehiclesRepository.connect(routeId: routeId, directionId: directionId) { outcome in
             if case let .ok(result) = onEnum(of: outcome) {
-                vehiclesData = Array(result.data.vehicles.values)
+                if let departures = nearbyVM.departures {
+                    vehiclesData = Array(departures.filterVehiclesByUpcoming(vehicles: result.data).values)
+                }
```
KaylaBrady, line 106: suggestion(if-minor): This looks good but seems pretty hard to test as-is. That may be a case for moving this logic into the VM - imagining a func like `handleNewVehicles(departures, vehicles)` that could be tested independently and confirms that `vehiclesData` is set to the filtered list.

KaylaBrady (approved):
Looks good!

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
EmmaSimon: I'm gonna hold off on doing this right now 🙏  It might make sense to roll into the changes when we split up the nearby VM?

PR #461, opened 2024-10-10
feat(iOS): forget predictions if backgrounded for over ten minutes

boringcactus:
### Summary

_Ticket:_ [[Loading states] Clear predictions when older than 10 minutes](https://app.asana.com/0/1205732265579288/1208468599574698/f)

My first draft had all the logic in the `onActive` callbacks, but there's enough of it that it makes sense to move it into the shared code.

### Testing

Manually verified that predictions are forgotten after ten minutes but not after nine minutes. Added unit tests for the logic in shared code; I'm not sure how straightforward it'd be to test the scene phase handlers, but if it's worth a look I can try.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/TripPredictionsRepository.kt
```diff
@@ -27,6 +30,10 @@ class TripPredictionsRepository(private val socket: PhoenixSocket) :
 
     override var lastUpdated: Instant? = null
 
+    override fun shouldForgetPredictions(predictionCount: Int) =
```
JackVCurtis, line 33: Small thing, but I think it might be more readable to make this a boolean with a more informative variable name, then passing `predictionCount > 0` as the argument in the caller.

JackVCurtis (approved):
👍

PR #464, opened 2024-10-10
fix(HomeMapView): Improve map layer backgrounding behavior

KaylaBrady:
### Summary

_Ticket:_ [Fix "map tile crash"](https://app.asana.com/0/1205732265579288/1208436806470885/f)

What is this PR for?
This PR intends to address the elusive map tile crash. As documented in the [investigation notes](https://www.notion.so/mbta-downtown-crossing/Map-tile-crash-119f5d8d11ea80e2bde7ca7f51b1abb5), this crash appears to be caused by the [Mapbox CleanupManager crash](https://github.com/mapbox/mapbox-maps-ios/issues/2174).

This PR attempts to address the issue in two ways:
~1. Bump the mapbox version to the latest~ location permission popup fails to appear when on the latest version, backed out that change
2. Improve map layer backgrounding behavior. In testing the version bump I noticed layers intermittently disappearing from the map after backgrounding & returning. Since the crash appears related to backgrounding / reopening the app, ensuring that map activity during backgrounding is minimal may help.

### Testing

What testing have you done?

Ran locally and in simulator, backgrounding & re-opening app to confirm layers persist. 
Also tested changing color scheme & re-opening app

### DIFF
--- a/iosApp/iosApp/Pages/Map/HomeMapViewLayerExtension.swift
+++ b/iosApp/iosApp/Pages/Map/HomeMapViewLayerExtension.swift
@@ -59,7 +59,10 @@ extension HomeMapView {
     }
 
     func refreshMap() {
+        print("KB: refreshing map")
         if let layerManager = mapVM.layerManager {
+            print("KB: refreshing map - layer manager present")
+
             if layerManager.currentScheme != colorScheme {
                 layerManager.addIcons(recreate: true)
                 addLayers(layerManager, recreate: true)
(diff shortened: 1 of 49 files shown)


iosApp/iosApp/Pages/Map/HomeMapViewLayerExtension.swift
```diff
@@ -60,13 +60,13 @@ extension HomeMapView {
 
     func refreshMap() {
         if let layerManager = mapVM.layerManager {
+            updateGlobalMapDataSources()
```
KaylaBrady, line 63: At one point I was getting warning messages about the sources for layers not being present. Updating the sources above adding the layers fixed the issue. This may be rendered a moot change by switching `addLayers` to `addPersistedLayers` below, but seemed worth keeping around.

iosApp/iosApp/Pages/Map/MapLayerManager.swift
```diff
@@ -76,9 +83,9 @@ class MapLayerManager: IMapLayerManager {
                 }
 
                 if map.layerExists(withId: "puck") {
-                    try map.addLayer(layer, layerPosition: .below("puck"))
+                    try map.addPersistentLayer(layer, layerPosition: .below("puck"))
```
KaylaBrady, line 86: persistentLayer keeps the map layers from flashing in every time the map returns from background

EmmaSimon (approved):

PR #472, opened 2024-10-15
fix: Show correct headsigns and groupings on all GL branches

EmmaSimon:
### Summary

_Ticket:_ [GL showing duplicate Eastbound @ Gov Center with no data](https://app.asana.com/0/1205732265579288/1208378317839446/f)

Most of the added lines here are in the tests, sorry for the large PR! 😅 

After #454, there were issues introduced where we would only ever display grouped nearby headsigns on the Medford/Tufts branch. This changes the headsign grouping to split grouped routes into separate headsigns if only a single route pattern is predicted in one direction. This means that it's technically possible (but very unlikely) for a grouping to be displayed when an atypical trip is predicted but has multiple other predictions before it and so wouldn't be displayed in nearby, but since this only applies to a stretch of 5 stops that rarely have more than 2 trains headed inbound at any time, this shouldn't happen often.

This also includes changes to use the pattern headsign to generate directions, since the destination names in the route only apply to the route's typical pattern, and includes a default null `Direction` object in the `StaticPatterns.ByHeadsign`, because we still want the special cased `Direction` object (like "Park St & North") used in `groupedDirection` to set the overall directions in `StopPatterns.ForLine` for displaying in the filtered stop details direction toggle.

Screenshots of correct behavior after these changes:

Simulator Screenshot - iPhone 15 Pro - 2024-10-15 at 11 42 17Simulator Screenshot - iPhone 15 Pro - 2024-10-15 at 11 42 22Simulator Screenshot - iPhone 15 Pro - 2024-10-15 at 11 46 04Simulator Screenshot - iPhone 15 Pro - 2024-10-15 at 11 46 09

### Testing

Added tests for newly added functionality and updated Direction behavior.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -133,6 +133,10 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
                     } else {
                         when (val it = typicalHere.first()) {
                             is StaticPatterns.ByDirection -> it.direction
+                            // When buildStopPatternsForLine creates a StaticPatterns.ByHeadsign,
+                            // it should always set it.direction to a non-null value, this fallback
+                            // is only here to make the compiler happy and do something sensible if
+                            // there are possible edge cases where one is set to null.
                             is StaticPatterns.ByHeadsign -> it.direction
                                     ?: Direction(
                                         it.route.directionNames[directionId] ?: "",
@@ -326,7 +330,7 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
             )
         }
 
-        fun buildStopPatternsForLine(
+        private fun buildStopPatternsForLine(
             stop: Stop,
             patterns: Map<Route, List<RoutePattern>>,
             line: Line,
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -683,10 +683,18 @@ class NearbyStaticDataBuilder {
             route: Route,
             headsign: String,
             patterns: List<RoutePattern>,
-            stopIds: Set<String> = allStopIds
+            stopIds: Set<String> = allStopIds,
+            direction: Direction? = null
         ) {
             data.add(
-                NearbyStaticData.StaticPatterns.ByHeadsign(route, headsign, line, patterns, stopIds)
+                NearbyStaticData.StaticPatterns.ByHeadsign(
+                    route,
+                    headsign,
+                    line,
+                    patterns,
+                    stopIds,
+                    direction
+                )
             )
         }
 
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
@@ -1,6 +1,7 @@
 package com.mbta.tid.mbta_app.model
 
 import co.touchlab.skie.configuration.annotations.DefaultArgumentInterop
+import com.mbta.tid.mbta_app.model.NearbyStaticData.StopPatterns.ForLine.Companion.groupedDirection
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.model.response.NearbyResponse
@@ -33,7 +34,8 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
             val headsign: String,
             val line: Line?,
             override val patterns: List<RoutePattern>,
-            override val stopIds: Set<String>
+            override val stopIds: Set<String>,
+            val direction: Direction? = null,
         ) : StaticPatterns() {
             override fun copy(patterns: List<RoutePattern>, stopIds: Set<String>) =
                 copy(route = route, patterns = patterns, stopIds = stopIds)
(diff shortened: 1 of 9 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -99,4 +101,143 @@ data class PatternsByStop(
             }
             .distinct()
     }
+
+    companion object {
+        // Even if a direction can serve multiple routes according to the static data, it's possible
+        // that only one of those routes is typical, in which case we don't want to display it as a
+        // grouped direction in the UI. If there are only predicted trips on a single route, this
+        // will split the grouped direction up into as many headsign rows as there are headsigns.
```
KaylaBrady, line 109: praise: this is very descriptive, thank you!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -99,4 +101,143 @@ data class PatternsByStop(
...
+        fun resolveRealtimePatternForDirection(
```
KaylaBrady, line 110: suggestion: this function is pretty long - consider breaking out some of the helper steps into separate functions

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -99,4 +101,143 @@ data class PatternsByStop(
...
+                staticData.routes.firstOrNull {
+                    it.id == headsignsAndPatternsToDisplayByRoute.keys.firstOrNull()
+                }
+            if (headsignsAndPatternsToDisplayByRoute.size == 1 && firstDisplayedRoute != null) {
+                val headsignsToDisplay =
+                    (headsignsAndPatternsToDisplayByRoute[firstDisplayedRoute.id] ?: emptyMap())
+                return headsignsToDisplay.map { (headsign, patternIds) ->
+                    val patternsForHeadsign = patternIds.mapNotNull { patternsById[it] }
+                    RealtimePatterns.ByHeadsign(
+                        NearbyStaticData.StaticPatterns.ByHeadsign(
```
KaylaBrady, line 215: question (non-blocking): would it be possible / simplify things at all to use the `RealtimePatterns.ByHeadsign` constructor that doesn't take a `StaticPatterns.ByHeadsign`?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -99,4 +101,143 @@ data class PatternsByStop(
...
+                    .toMap()
+            val upcomingPatternsByRouteAndHeadsign =
+                tripsByPattern.values
+                    .flatten()
+                    .groupBy({ it.trip.routeId }, { it.trip.headsign to it.trip.routePatternId })
+                    .mapValues { routeEntry ->
+                        routeEntry.value.groupBy({ it.first }, { it.second })
+                    }
+
+            val predictedPatternsByRoute =
```
KaylaBrady, line 166: question (non-blocking): Confirming my own learning - if we skip the step of checking predicted patterns by route, then we'd break down into showing individual headsigns prematurely because there could be some late-at-night atypical headsigns.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -99,4 +101,143 @@ data class PatternsByStop(
...
+                staticData.patterns
+                    .filter {
+                        tripsByPattern[it.id]?.any { trip ->
+                            trip.prediction != null && trip.vehicle?.tripId == trip.trip.id
```
KaylaBrady, line 170: question: why check the vehicle's trip ID here? Wouldn't we still show the prediction even if the vehicle is on another trip?

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/PatternsByStopTest.kt
```diff
@@ -300,4 +300,309 @@ class PatternsByStopTest {
         assertEquals(listOf(alert), patternsByStop.alertsHereFor(0, global))
         assertEquals(listOf(alert), patternsByStop.alertsHereFor(1, global))
     }
+
+    @Test
+    fun `resolveRealtimePatternForDirection splits direction groups into headsigns when needed`() {
```
KaylaBrady, line 305: suggestion: I'm having a bit of a hard time understanding this test since there is a lot of data defined & different cases asserted. Consider breaking this down into smaller tests that potentially share relevant data

KaylaBrady (commented):
Nice work, this is tricky! Mostly comments from trying to wrap my head around this GL edge case better

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
EmmaSimon: I actually set it up like that initially, by adding a new constructor for `RealtimePatterns.ByHeadsign` that took a `StaticPatterns.ByDirection`, then decided that doing it like this was actually the most straightforward approach.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -113,20 +115,32 @@ data class NearbyStaticData(val data: List<TransitWithStops>) {
...
+                    } else if (typicalHere.size > 1) {
+                        Direction(
+                            routes.first().directionNames[directionId] ?: "",
+                            null,
+                            directionId
+                        )
+                    } else {
+                        when (val it = typicalHere.first()) {
+                            is StaticPatterns.ByDirection -> it.direction
+                            is StaticPatterns.ByHeadsign -> it.direction
```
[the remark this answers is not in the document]
EmmaSimon: That's right, yeah. I think that would add additional confusion to all of the other places we use this, and here we would still need to do something in the case where it was a regular `StaticPatterns.ByHeadsign`, even though that should never be possible.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
EmmaSimon: Hmm, maybe it's a bad idea. I wanted to prevent cases where we would be displaying the grouping without actually displaying the atypical trip. So like, if there was a prediction for a C which is currently approaching Medford Tufts to start a new trip, but two E trains in front of it heading west, we would still show the group even though only the E trains are shown in the upcoming predictions. But maybe that's fine and this is too cautious and will hide predictions when we don't want it to.

EmmaSimon: I removed this check, I think it'll be okay to briefly show groups sometimes even when it's not totally clear why.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/PatternsByStopTest.kt
EmmaSimon: Thanks for suggesting this! I quite like the pattern I ended up with with all the shared data in an object that each case references, I'll try to remember to do that sort of thing going forward when multiple tests share similar data.

KaylaBrady (approved):

PR #473, opened 2024-10-15
feat: recently viewed stops in search

BrandonTR:
### Summary

_Ticket:_ [Search UI: display recent stops in empty state](https://app.asana.com/0/1205732265579288/1208355161933391/f)

Show recently visited stops in search empty state. I refactored this a bit to always associate with stops/routes/lines from global data instead of using the data from Algolia. 

### Testing

Added unit tests for new RoutePillSpec changes, fixed SearchResultViewTests for latest refactor changes and added a test for Recently Viewed state.


iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
```diff
@@ -65,40 +72,58 @@ struct SearchResultsContainer: View {
...
-            } catch {}
-        }
+                }
+        } catch {}
+    }
+
+    private func mapStopIdToResult(id: String) -> Result? {
+        guard let globalResponse, let stop = globalResponse.stops[id] else { return nil }
+        let isStation = stop.locationType == .station
+        let routePills: [RoutePillSpec] = globalResponse.trips
```
EmmaSimon, line 104: nitpick: I don't think efficiency is actually that important here, but it might make more sense to filter through either all routes or all route patterns here, rather than trips, since that's a much smaller set of objects. Then we'd only have to look at each of the route pattern's representative trips, rather than every trip.

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
```diff
@@ -120,12 +143,10 @@ struct SearchResultsContainer: View {
...
             if !query.isEmpty {
-                loadResults(query: query)
-            }
-            loadVisitHistory()
-            Task {
-                await searchVM.loadSettings()
+                Task { await loadResults(query: query) }
             }
+            Task { await loadVisitHistory() }
+            Task { await searchVM.loadSettings() }
```
EmmaSimon, line 149: Nice, I appreciate the consistency of these now!

iosApp/iosApp/Pages/Search/Results/StopResultsView.swift
```diff
@@ -40,9 +40,11 @@ struct StopResultsView: View {
                     .foregroundStyle(Color.text)
                     .multilineTextAlignment(.leading)
                     .frame(maxWidth: .infinity, alignment: .leading)
-                HStack(spacing: 4) {
-                    ForEach(stop.routes, id: \.self) { route in
-                        RoutePill(stopResultRoute: route, type: .flex)
+                ScrollView(.horizontal, showsIndicators: false) {
```
EmmaSimon, line 43: I think this will still scroll and capture drags even if the content is to short to actually be off the screen, can we add a `.scrollBounceBehavior(.basedOnSize, axes: [.horizontal])` like in [TripDetailsStopView](https://github.com/mbta/mobile_app/blob/main/iosApp/iosApp/Pages/TripDetails/TripDetailsStopView.swift#L56)?

iosApp/iosApp/Pages/Search/Results/StopResultsView.swift
```diff
@@ -10,13 +10,13 @@ import shared
...
 struct StopResultsView: View {
-    let stops: [StopResult]
+    let stops: [SearchResultsContainer.Result]
     let handleStopTap: (String) -> Void
 
     var body: some View {
         ResultContainer {
             VStack(spacing: .zero) {
-                ForEach(stops, id: \.id) { stop in
+                ForEach(stops) { stop in
```
EmmaSimon, line 19: Why get rid of the ID?

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
```diff
@@ -154,32 +175,31 @@ struct SearchResultView_Previews: PreviewProvider {
     static var previews: some View {
         SearchResultsView(
```
EmmaSimon, line 176: On the initial tap to open up the search results, I'm not seeing any recent visits until after I search for and clear the search field.

Unrelated, but also visible in the screenshot, the search being active no longer dismisses the sheet.

Simulator Screenshot - iPhone 15 Pro - 2024-10-16 at 14 51 30

iosApp/iosApp/Pages/Search/Results/StopResultsView.swift
BrandonTR: It's not needed if the struct conforms to the Identifiable protocol, it'll automatically assign it the id of the element.

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
BrandonTR: I think it might be! Looking into it 😄

BrandonTR: So I tried a few things to resolve this, first by just making sure some of this work was done on another thread which resolves the UI hanging but it was still super slow. So I looked into methods to optimize the route association on the Swift side but didn't really come up with anything. In the end it turns out that this is much more efficient if you just do it in Kotlin and call it from Swift!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/GlobalResponse.kt
```diff
@@ -68,4 +68,21 @@ data class GlobalResponse(
         } else {
             null
         }
+
+    fun getRoutesFor(stopId: String): List<Route> {
+        return routePatterns.values.mapNotNull {
```
boringcactus, line 73: Wouldn't using `patternIdsByStop` be more direct here?

BrandonTR: I tried this and I think I'm either overlooking or missing something because it only works for bus stops. Here's the code I was trying:
```
        return patternIdsByStop[stopId]?.mapNotNull { patternId ->
            val routePattern = routePatterns[patternId] ?: return@mapNotNull null
            val tripId = routePattern.representativeTripId
            val trip = trips[tripId] ?: return@mapNotNull null
            val stop = stops[stopId] ?: return@mapNotNull null
            if (
                trip.stopIds?.contains(stopId) == true ||
                    trip.stopIds?.any(stop.childStopIds::contains) == true &&
                        routePattern.typicality == RoutePattern.Typicality.Typical
            ) {
                return@mapNotNull routes[trip.routeId]
            } else {
                return@mapNotNull null
            }
        } ?: emptyList()
```
Does that seem right?

boringcactus: I'd expect something like
```kotlin
val patternIds =
    if (stop.childStopIds.isEmpty()) {
        patternIdsByStop[stopId].orEmpty()
    } else {
        stop.childStopIds.flatMap { patternIdsByStop[it] }
    }
return patternIds.mapNotNull { routePatterns[it]?.routeId }.distinct().mapNotNull { routes[it] }
```
since it's the child stops that are keys in `patternIdsByStop` and the values there are already known to visit that stop.

BrandonTR: Ah right, I understand now. Thank you!

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
```diff
@@ -39,113 +24,41 @@ struct SearchResultsContainer: View {
...
     var body: some View {
         SearchResultsView(
-            state: resultsState,
+            state: searchVM.resultsState,
             handleStopTap: handleStopTap
         )
+        .task {
+            await searchVM.getGlobalData()
+            await searchVM.loadSettings()
+            await searchVM.loadVisitHistory()
```
EmmaSimon, line 50: nitpick: I'm not certain, but I think the way this is written, each of these are done in sequence, right? Could it speed things up slightly to have a separate task for each one?

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
EmmaSimon: Nice, calling it from kotlin seems like the best approach anyway! The recently viewed loading looks good now 🎉 

I am still seeing the sheet displayed over the search results though. In the initial PR I'd added [this check](https://github.com/mbta/mobile_app/pull/429/files#diff-aa4e2052b22ab2454329f9761c75169e801d78127b9849bd18ed3d1e31c447aaR156-R158) to hide the sheet when the search was focused, that seems to still be in place so I'm not sure what's going wrong here...

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
BrandonTR: I think visitHistory at the very least benefits from being in this sequence but it does make sense to move loadSettings out!

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
EmmaSimon: Oh, lol, we stopped using the sheet items at all. This should do it.
```swift
@ViewBuilder var mapWithSheets: some View {
        let nav = nearbyVM.navigationStack.lastSafe()
        let sheetItemId: String? = nav.sheetItemIdentifiable()?.id
        mapSection
            .sheet(isPresented: .constant(!(searchObserver.isSearching && nav == .nearby)), content: { ... }
```

iosApp/iosApp/Pages/Search/Results/SearchResultsContainer.swift
EmmaSimon: Cool, yeah, I wasn't sure which were dependent on each other. It might make sense to just add a helper function in searchVM to call all of them in sequence or parallel as needed, so you only need to do a single call here.

PR #475, opened 2024-10-17
feat: Style stale predictions banner to match spec

EmmaSimon:
### Summary

_Ticket:_ [[Loading states] Style stale predictions banner](https://app.asana.com/0/1205732265579288/1208468599574778/f)

This includes a refactor to move the error banner state into its own view model, this was necessary so that the ErrorBanner component could return an `EmptyView` when the state was nil, otherwise the `.collect` helper would have needed an actual view in the nil state, which was taking up space and padding in the layout that couldn't match spec.

I couldn't resist sneaking in a few broader layout and spacing updates to get things closer to spec, mainly between route cards in nearby and unfiltered stop details, but also made header spacing more consistent, and added some extra padding under filtered stop details departures.

Simulator Screenshot - iPhone 15 Pro - 2024-10-17 at 12 36 14Simulator Screenshot - iPhone 15 Pro - 2024-10-17 at 12 36 54Simulator Screenshot - iPhone 15 Pro - 2024-10-17 at 12 36 36

### Testing

Updated tests to work with the new ErrorBannerViewModel.


BrandonTR (approved):
Nice!

PR #477, opened 2024-10-18
fix: Add back Package.resolved

EmmaSimon:
### Summary

_Ticket:_ No ticket

This got accidentally included in a commit and is causing pipelines to fail (we should really fix the issue where it deletes itself constantly...)


BrandonTR (approved):
Thanks!

boringcactus:
I think Xcode deletes the Package.resolved when the .xcproj file is changed by Git while Xcode is running. The error message mentions "a resolved file is required when automatic dependency resolution is disabled" so maybe we should enable automatic dependency resolution and ditch the lockfiles entirely. (A simpler solution might be to avoid merging PRs where the Xcode Cloud build is failing.)

PR #478, opened 2024-10-18
feat(ErrorBanner): offline banner

KaylaBrady:
### Summary

_Ticket:_ [Offline error banner](https://app.asana.com/0/1205732265579288/1208306566672329/f)

What is this PR for?

This PR adds an error banner that appears when a user is offline.

I used the following resources as a basis for this implementation:
* [Blog post: Use Kotlin Multiplatform With a Network Listener Project](https://www.andela.com/blog-posts/use-kotlin-multiplatform-with-a-network-listener-project)
* [GH comment about using nw_path_monitor in iosMain](https://github.com/ln-12/multiplatform-connectivity-status/issues/7#issuecomment-1924106630)

I drifted from the Blog post's approach in a few ways:
* It publishes NetworkStatus as a flow. We already have an ErrorBannerState flow, so rather than introducing an intermediate flow, NetworkStatus changes just update the ErrorBannerState flow directly.
* It implements the network monitor on the ios-side in Swift - I incorporated the approach in the linked GH comment so the monitoring could be done entirely in kotlin.

### Testing

What testing have you done?
* Added unit tests
* Ran locally turning airplane mode on & off.

I included the android implementation from the blog post but did not test it.
image

### DIFF
--- a/iosApp/iosApp/ComponentViews/ErrorBanner.swift
+++ b/iosApp/iosApp/ComponentViews/ErrorBanner.swift
@@ -50,8 +50,8 @@ struct ErrorBanner: View {
             ErrorCard { HStack {
                 Image(systemName: "wifi.slash")
                 Text("Unable to connect")
-            }
-            }
+                Spacer()
+            }}
         case nil:
             // for some reason, .collect on an EmptyView doesn't work
             EmptyView()
--- a/iosApp/iosApp/ComponentViews/ErrorBanner.swift
+++ b/iosApp/iosApp/ComponentViews/ErrorBanner.swift
@@ -46,6 +46,12 @@ struct ErrorBanner: View {
                 }
                 .frame(minHeight: minHeight)
             }
+        case .networkError:
+            ErrorCard { HStack {
+                Image(systemName: "wifi.slash")
+                Text("Unable to connect")
+            }
+            }
         case nil:
             // for some reason, .collect on an EmptyView doesn't work
             EmptyView()
--- a/iosApp/iosAppTests/ViewModels/ErrorBannerViewModelTests.swift
+++ b/iosApp/iosAppTests/ViewModels/ErrorBannerViewModelTests.swift
@@ -1,0 +1,26 @@
+//
+//  ErrorBannerViewModelTests.swift
+//  iosAppTests
+//
+//  Created by Kayla Brady on 10/18/24.
+//  Copyright © 2024 MBTA. All rights reserved.
+//
+
+import Foundation
+@testable import iosApp
+import shared
+import XCTest
+
+final class ErrorBannerViewModelTests: XCTestCase {
+    func testActivateSubscribesToNetworkChanges() async {
+        let onSubscribeExp = XCTestExpectation(description: "onSubscribe called")
+        let repo = MockErrorBannerStateRepository(state: nil, onSubscribeToNetworkChanges: { onSubscribeExp.fulfill() })
+        let errorVM = ErrorBannerViewModel(
+            errorRepository: repo,
+            initialLoadingWhenPredictionsStale: false,
+            skipListeningForStateChanges: true
+        )
+        await errorVM.activate()
+        wait(for: [onSubscribeExp], timeout: 1)
+    }
+}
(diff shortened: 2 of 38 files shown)


shared/src/androidMain/kotlin/com/mbta/tid/mbta_app/Helpers.kt
```diff
@@ -14,6 +14,6 @@ internal fun createDataStore(context: Context): DataStore<Preferences> =
 fun initKoin(appVariant: AppVariant, nativeModule: Module, context: Context) {
     startKoin {
         androidContext(context)
-        modules(appModule(appVariant) + platformModule() + nativeModule)
+        modules(platformModule() + appModule(appVariant) + nativeModule)
```
EmmaSimon, line 17: Just curious, but why were these swapped?

iosApp/iosApp/ComponentViews/ErrorBanner.swift
```diff
@@ -46,6 +46,12 @@ struct ErrorBanner: View {
                 }
                 .frame(minHeight: minHeight)
             }
+        case .networkError:
+            ErrorCard { HStack {
+                Image(systemName: "wifi.slash")
+                Text("Unable to connect")
```
EmmaSimon, line 52: Because this has no reload button like the other two banners, it also has no `Spacer()`, so we're getting the awkward squished banner. If you toss a `Spacer()` at the end here, I think that'll get it to take up the full width.

iosApp/iosApp/ComponentViews/ErrorBanner.swift
```diff
@@ -46,6 +46,12 @@ struct ErrorBanner: View {
...
+            }
+            }
```
EmmaSimon, line 54: nitpick: I'd put these braces on the same line 😛

iosApp/iosAppTests/ViewModels/ErrorBannerViewModelTests.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Kayla Brady on 10/18/24.
+//  Copyright © 2024 MBTA. All rights reserved.
+//
+
+import Foundation
+@testable import iosApp
+import shared
+import XCTest
+
+final class ErrorBannerViewModelTests: XCTestCase {
```
EmmaSimon, line 14: Thanks for adding this!

iosApp/iosApp/ComponentViews/ErrorBanner.swift
```diff
@@ -46,6 +46,12 @@ struct ErrorBanner: View {
                 }
                 .frame(minHeight: minHeight)
             }
+        case .networkError:
+            ErrorCard { HStack {
```
EmmaSimon, line 50: Is the default alignment in this HStack not center? It looks kinda like the image is below the centerline of the text.

EmmaSimon (commented):
Nothing major, but I have a couple points on the layout. All the logic changes look great.

shared/src/androidMain/kotlin/com/mbta/tid/mbta_app/Helpers.kt
KaylaBrady: Oh this isn't actually necessary. At first I tried having ErrorBannerStateRepo call `subscribeToNetworkStatusChanges` on init, but that was causing dependency injection issues with `BeanNotFound` errors. I thought maybe the platformModule needed to be called first here since appModule depends on it, but that didn't end up fixing the issue, and I just opted to call `subscribeToNetworkStatusChanges` from the ViewModel instead.

EmmaSimon (approved):

iosApp/iosApp/ComponentViews/ErrorBanner.swift
KaylaBrady: Added some borders to each view - I _think_ looks it is centered when accounting for the full line height of the text.
image

I swapped in `HStack(alignment: .center)` and didn't notice a difference

EmmaSimon: Okay yeah, I think it's just a matter of the icon being slightly bottom heavy. Thanks for checking!

PR #480, opened 2024-10-21
feat: shuffle around partially loaded data

boringcactus:
### Summary

_Ticket:_ [Shimmer loading in sheet](https://app.asana.com/0/1201654106676769/1208304466486188/f)

The [spec](https://www.notion.so/mbta-downtown-crossing/Loading-Error-States-Spec-edbb36cc74344e6aa6364f9f54458710?p=9687a4e21c224a859cf2cd2334ab5d39&pm=s) says that predictions and alerts need to be loaded before predictions, alerts, or schedules can be shown, and the shimmer-loading designs (which I will implement in a follow-up PR) imply that we should not show the nearby static data while predictions or alerts are loading.

### Testing

Updated unit tests to reflect new behavior. Manually validated my local branch with the full changes using [Comcast](https://github.com/tylertreat/comcast) to add latency and limit bandwidth since stop and trip details tend to load too fast to see the loading state.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/DirectionPickerTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/DirectionPickerTest.kt
@@ -39,8 +39,8 @@ class DirectionPickerTest {
             }
         val patterns =
             listOf(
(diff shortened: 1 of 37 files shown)


BrandonTR (approved):
lgtm!

PR #488, opened 2024-10-23
feat(iOS): add option to hide maps

boringcactus:
### Summary

_Ticket:_ [Setting to hide maps](https://app.asana.com/0/1205732265579288/1208572284247927/f)

As I expected, the way we broke up the ContentView makes this not difficult at all to implement.

### Testing

Added a unit test. Manually validated that everything behaves reasonably (including with VoiceOver) and [checked in Slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1729711383158839) that no changes need to be made.

### DIFF
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -102,22 +102,32 @@ struct ContentView: View {
     var nearbyTab: some View {
         VStack {
             locationAuthHeader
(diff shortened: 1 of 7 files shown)


EmmaSimon (approved):
The small amount of repeated code makes me sad, and I can see it causing issues if we ever change anything in the full screen cover logic, but I still think this is the best approach for now.

PR #490, opened 2024-10-24
Location Services Button

JackVCurtis:
### Summary

_Ticket:_ [Location services button](https://app.asana.com/0/1205732265579288/1208304466486202/f)

What is this PR for?

Adds a button to link the user to Settings when Location functionality is disabled.

Screenshot 2024-10-25 at 2 07 06 PM

Screenshot 2024-10-25 at 2 08 18 PM

### Testing

What testing have you done?

UI test for the button and alert popup

### DIFF
--- a/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
+++ b/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
@@ -57,7 +57,7 @@ struct LocationAuthButton: View {
             .buttonStyle(LocationAuthButtonStyle())
             .accessibilityIdentifier("locationServicesButton")
             .alert(
-                "Maps work best with Location Services turned on",
+                "MBTA Go works best with Location Services turned on",
                 isPresented: $showingAlert,
                 actions: {
                     Button("Turn On in Settings") {
@@ -66,7 +66,7 @@ struct LocationAuthButton: View {
                             UIApplication.shared.open(url)
                         }
                     }
-                    Button("Keep Location Services off") {
+                    Button("Keep Location Services Off") {
                         showingAlert = false
                     }
                 },
--- a/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
+++ b/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
@@ -46,6 +46,7 @@ struct LocationAuthButton: View {
                     Image(.faChevronRight)
                         .resizable()
                         .scaleEffect(0.6)
+                        .aspectRatio(contentMode: .fit)
                         .frame(width: 20, height: 20, alignment: .center)
                         .backgroundStyle(.white)
                         .background(in: .circle)
--- a/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
+++ b/iosApp/iosApp/ComponentViews/LocationAuthButton.swift
@@ -1,0 +1,84 @@
+//
+//  LocationAuthButton.swift
+//  iosApp
+//
+//  Created by Jack Curtis on 10/28/24.
+//  Copyright © 2024 MBTA. All rights reserved.
+//
+
+import Foundation
+import SwiftUI
+
+struct LocationAuthButtonStyle: ButtonStyle {
+    func makeBody(configuration: Configuration) -> some View {
+        configuration.label
+            .padding()
+            .background(Color.key)
+            .foregroundStyle(.white)
+            .clipShape(Capsule())
+    }
+}
+
+struct LocationAuthLabelStyle: LabelStyle {
+    func makeBody(configuration: Configuration) -> some View {
+        HStack(alignment: .center, spacing: 8) {
+            configuration.title
+            configuration.icon
+        }
+        .fontWeight(.bold)
+    }
+}
+
+struct LocationAuthButton: View {
+    @EnvironmentObject var locationDataManager: LocationDataManager
+
+    @Binding var showingAlert: Bool
+
+    var body: some View {
+        switch locationDataManager.authorizationStatus {
+        case .notDetermined, .denied, .restricted:
+            Button(action: {
+                showingAlert = true
+            }, label: {
+                Label(title: {
+                    Text("Location Services is off")
+                }, icon: {
+                    Image(.faChevronRight)
+                        .resizable()
+                        .scaleEffect(0.6)
+                        .frame(width: 20, height: 20, alignment: .center)
+                        .backgroundStyle(.white)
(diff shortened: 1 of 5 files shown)


iosApp/iosApp/ContentView.swift
```diff
@@ -88,17 +90,70 @@ struct ContentView: View {
...
+                            UIApplication.shared.open(url)
+                        }
+                    }
+                    Button("Keep Location Services off") {
+                        showingLocationPermissionAlert = false
+                    }
+                },
+                message: {
+                    Text(
+                        "You'll get turn-by-turn directions, estimated travel times, and improved search results when you turn on Location Services for Maps"
```
KaylaBrady, line 151: issue: this text should be "You’ll see nearby transit options and get better search results when you turn on Location Services for MBTA Go.", we don't offer turn-by-turn directions (at least not yet!)

iosApp/iosApp/ContentView.swift
```diff
@@ -118,10 +172,11 @@ struct ContentView: View {
             } else {
                 ZStack(alignment: .top) {
                     mapWithSheets
-                    VStack(alignment: .trailing, spacing: 0) {
+                    VStack(alignment: .center, spacing: 0) {
                         if nearbyVM.navigationStack.lastSafe() == .nearby {
                             SearchOverlay(searchObserver: searchObserver, nearbyVM: nearbyVM, searchVM: searchVM)
                         }
+                        locationAuthButton
```
KaylaBrady, line 179: The button is unexpectedly still showing when search is opened. Similar to the RecenterButton below, I think a `if !searchObserver.isSearching` check to remove this would do the trick

IMG_0166

iosApp/iosAppUITests/HomeMapViewUITests.swift
```diff
@@ -62,6 +62,26 @@ final class HomeMapViewUITests: XCTestCase {
         map.swipeDown()
         XCTAssertFalse(recenterButton.exists)
     }
+
+    func testLocationServicesButtonWithNoLocation() throws {
```
KaylaBrady, line 66: praise: thanks for adding a test for this!

iosApp/iosApp/ContentView.swift
```diff
@@ -88,17 +90,70 @@ struct ContentView: View {
         }
     }
 
+    struct LocationAuthButtonStyle: ButtonStyle {
```
KaylaBrady, line 93: suggestion (non-blocking): Consider moving this out into a separate LocationAuthButton file

iosApp/iosApp/ContentView.swift
KaylaBrady: Also seeing some unexpected alignment in the stop details page - button is no longer centered

IMG_0167

iosApp/iosApp/ContentView.swift
JackVCurtis: 🤦 Sorry, I was looking at the image rather than the text.

iosApp/iosApp/ContentView.swift
JackVCurtis: Oh odd, thought I tested that early on. Maybe I messed something up in the rebase, I'll check

JackVCurtis: Ah, it was my fault for not waiting long enough! The button appears to be hidden by the keyboard at first, but then shifts to be visible on the screen if you wait a second

KaylaBrady:
@JackVCurtis those changes are looking good! Only thing outstanding from my perspective is that the button is unexpectedly shifted to the top left in the stop details & trip details pages.

Also, I realize there weren't UI designs for this task, but I'd suggest sharing an image in slack to get any feedback from design which could be incorporated in this or a follow-up PR.

image

JackVCurtis:
> @JackVCurtis those changes are looking good! Only thing outstanding from my perspective is that the button is unexpectedly shifted to the top left in the stop details & trip details pages.
> 
> Also, I realize there weren't UI designs for this task, but I'd suggest sharing an image in slack to get any feedback from design which could be incorporated in this or a follow-up PR.
> 
> image

Thanks @KaylaBrady, I think I just need to tweak a setting on those two pages. Will fix the alignment and post the screenshot to Slack

JackVCurtis:
@KaylaBrady I caught an additional issue, which is that if you navigate to the Stop view via the search bar the Location Services button doesn't appear at all.

KaylaBrady:
> @KaylaBrady I caught an additional issue, which is that if you navigate to the Stop view via the search bar the Location Services button doesn't appear at all.

@JackVCurtis I'm not sure whether we need the button to appear outside nearby transit, worth checking with product! In any case I don't see that as needing to block this current PR and could be addressed as a follow-up if needed.

KaylaBrady (approved):
Looks good to me, pending any blocking feedback from product!

iosApp/iosApp/ComponentViews/LocationAuthButton.swift
```diff
@@ -0,0 +1,84 @@
...
+                        .backgroundStyle(.white)
+                        .background(in: .circle)
+                        .foregroundColor(Color.key)
+                })
+                .labelStyle(LocationAuthLabelStyle())
+            })
+            .buttonStyle(LocationAuthButtonStyle())
+            .accessibilityIdentifier("locationServicesButton")
+            .alert(
+                "Maps work best with Location Services turned on",
```
EmmaSimon, line 59: Oh, just noticed that this says "Maps" when it should probably say "MBTA Go"

iosApp/iosApp/ComponentViews/LocationAuthButton.swift
```diff
@@ -0,0 +1,84 @@
...
+                isPresented: $showingAlert,
+                actions: {
+                    Button("Turn On in Settings") {
+                        showingAlert = false
+                        if let url = URL(string: UIApplication.openSettingsURLString) {
+                            UIApplication.shared.open(url)
+                        }
+                    }
+                    Button("Keep Location Services off") {
```
EmmaSimon, line 69: Could we also capitalize "Off" for consistency?

JackVCurtis: I believe we use sentence case for this copy, it's just that Location Services is a proper noun so it gets capitalized

JackVCurtis: Oh but it is capitalized in the spec. I'll correct it then.

PR #492, opened 2024-10-24
feat: Start caching rail shapes for the map

EmmaSimon:
### Summary

_Ticket:_ [Cache route shape requests](https://app.asana.com/0/1205732265579288/1208254208398835/f)

Stacked onto #489 

Start caching rail route shapes on disk. This cuts the initial time to display map layers down to ~2 seconds, even on a slow internet connection.

### Testing

Added test for new flow in repo.


JackVCurtis (approved):
Not sure I'm the most qualified to review this one. It looks straightforward enough, though. Understandable if you want more eyes.

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -80,25 +69,54 @@ extension HomeMapView {
...
+                errorBannerRepository,
+                errorKey: "HomeMapView.handleAppear",
+                getData: { try await railRouteShapeRepository.getRailRouteShapes() },
+                onRefreshAfterError: fetchRouteShapes
+            )
+        }
+    }
+
+    func loadRouteShapes() {
+        Task(priority: .high) {
```
JackVCurtis, line 116: Do we set task priority anywhere else in the app?

EmmaSimon: I started using it as part of the global data caching when performing map layer tasks, and that's the only other place we use it currently. There were a few times where it seemed like the tasks were getting delayed a weird amount, so I tried adding the priority explicitly. It's probably not strictly necessary, but we do want these to be executed as quickly as possible so that the map layers appear asap.

PR #495, opened 2024-10-28
fix: remove extra chevron in affected stops dropdown

BrandonTR:
### Summary

_Ticket:_ [Fix: extra chevron appearing in affected stops accordion](https://app.asana.com/0/1205732265579288/1208468599574763/f)

Remove duplicate chevron in affected stops dropdown on alert details page. This change looks bigger than it actually is due to reformatting, the only impactful change is the `.tint(.clear)` modifier.

### Testing

Manual, no automated tests.


JackVCurtis (approved):
👍 Odd that the linter didn't catch that on commit

PR #497, opened 2024-10-28
feat(iOS): onboarding flow UX

boringcactus:
### Summary

_Ticket:_ [Implement onboarding UX](https://app.asana.com/0/1205732265579288/1208468599574770/f)

Adds the structure for the onboarding flow.

I'm seeing issues with the first map load when I uninstall and reinstall the app and go through the onboarding flow, but I'm hoping it's just an App Check debug weirdness, because if it's not that then I don't understand what it would be.

### Testing

Manually verified that the flow works and picks up VoiceOver status on iOS.


iosApp/iosApp/Pages/Map/HomeMapView.swift
```diff
@@ -105,8 +105,13 @@ struct HomeMapView: View {
                  */
                 nearbyVM.selectingLocation = true
             }
-            .onAppear {
-                locationDataManager.locationFetcher.requestWhenInUseAuthorization()
+            .task {
```
KaylaBrady, line 108: question: Do we still need to request permissions from within the map view if it is being requested as part of the onboarding flow?

iosApp/iosApp/Pages/Onboarding/OnboardingPage.swift
```diff
@@ -0,0 +1,33 @@
...
+
+struct OnboardingPage: View {
+    let screens: [OnboardingScreen]
+    @State var selectedIndex: Int = 0
+    let onFinish: () -> Void
+
+    let inspection = Inspection<Self>()
+
+    var body: some View {
+        OnboardingScreenView(screen: screens[selectedIndex], advance: {
```
KaylaBrady, line 20: issue: On a run without voice over turned on, I get an index out of range error when hitting "Get started" from the feedback screen

iosApp/iosApp/Pages/Onboarding/OnboardingScreenView.swift
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }
+                Button("Show maps") {
+                    Task {
+                        try await settingUseCase.set(setting: .hideMaps, value: false)
+                        try? await onboardingRepository.markOnboardingCompleted(screen: screen)
+                        advance()
+                    }
+                }
+            case .location:
```
KaylaBrady, line 73: question: This means existing users who already have decided to share or not share their location permissions would see this prompt - is that right? Could we limit it so that only new users see this? (Or confirm with product that it is OK not to if it hasn't been confirmed already)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainApplication.kt
```diff
@@ -17,6 +18,14 @@ class MainApplication : Application() {
 
     override fun onCreate() {
         super.onCreate()
-        initKoin(appVariant, makeNativeModule(appCheck, socket.wrapped()), this)
+        initKoin(
+            appVariant,
+            makeNativeModule(
+                AccessibilityStatusRepository(applicationContext),
```
KaylaBrady, line 24: praise: I like the way you approached putting the AccessibilityStatus in shared code

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/OnboardingRepository.kt
```diff
@@ -0,0 +1,60 @@
...
+    suspend fun getPendingOnboarding(): List<OnboardingScreen>
+
+    suspend fun markOnboardingCompleted(screen: OnboardingScreen)
+}
+
+class OnboardingRepository : IOnboardingRepository, KoinComponent {
+    private val accessibilityStatus: IAccessibilityStatusRepository by inject()
+    private val dataStore: DataStore<Preferences> by inject()
+
+    private val onboardingCompletedKey = stringSetPreferencesKey("onboardingCompleted")
```
KaylaBrady, line 24: suggestion(non-blocking): "onboardingCompleted" as a string name reads to me as a boolean value, event though I know it isn't. Consider calling it something like `onboardingCompletedScreens` for clarity

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/OnboardingRepository.kt
```diff
@@ -0,0 +1,60 @@
...
+
+interface IOnboardingRepository {
+    suspend fun getPendingOnboarding(): List<OnboardingScreen>
+
+    suspend fun markOnboardingCompleted(screen: OnboardingScreen)
+}
+
+class OnboardingRepository : IOnboardingRepository, KoinComponent {
+    private val accessibilityStatus: IAccessibilityStatusRepository by inject()
+    private val dataStore: DataStore<Preferences> by inject()
```
KaylaBrady, line 22: question(non-blocking): Why consume from the dataStore directly instead of from SettingsRepo? It seems ideal for maintaining clean boundaries between repos to limit the number of places that read from the data store directly.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/OnboardingRepository.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+interface IOnboardingRepository {
+    suspend fun getPendingOnboarding(): List<OnboardingScreen>
+
+    suspend fun markOnboardingCompleted(screen: OnboardingScreen)
+}
+
+class OnboardingRepository : IOnboardingRepository, KoinComponent {
```
KaylaBrady, line 20: thought (non-blocking): we haven't done this consistently, but I think it would be worth landing on a naming convention for repositories that use other repositories to help with maintaining clear layers / hierarchy. We could adopt UseCase more consistently for that.

KaylaBrady:
@boringcactus on the map tile loading issue - thanks for recording that! I observed a map tile issue on a [user's phone](https://www.notion.so/mbta-downtown-crossing/6d3d022079af445e81db6cfaaca38db4?v=366dc52cd41e4b36b4a894d2bdb1009d&p=12df5d8d11ea80faadbfd09511af1e8d&pm=s) over the weekend, so any more breadcrumbs around that would be useful.

iosApp/iosApp/Pages/Map/HomeMapView.swift
EmmaSimon: I just got the allow location popup on the location onboarding screen before I tapped any buttons, is that related to this?

EmmaSimon: But we do need to show the popup every time the app launches if the user only selects "Allow Once"

boringcactus: If you get the system popup before you tap anything, that's probably a result of this sleep not being long enough, yeah. I can try wiring it up to the `ContentViewModel` to actually wait for the onboarding state to load instead of just hoping it loads faster than some hardcoded delay.

iosApp/iosApp/Pages/Onboarding/OnboardingPage.swift
boringcactus: This `advance` logic was working when I still had everything in a `TabView` for the page count view, but I guess I forgot to test it again when I got rid of that, and for some reason the unit test works anyway.

boringcactus:
For the map failing to load, the style is 401ing and the sources are missing when the style layers are added. I've tried moving loading the config around some, but it didn't help.

iosApp/iosApp/Pages/Onboarding/OnboardingScreenView.swift
boringcactus: I guess we could define that the location onboarding screen doesn't apply for users who have already given the app location access; that would be complicated to check in the existing framework, though.

iosApp/iosAppTests/Pages/Onboarding/OnboardingScreenViewTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        let sut = OnboardingScreenView(
+            screen: .location,
+            advance: { advanceExp.fulfill() },
+            createLocationFetcher: { locationFetcher },
+            onboardingRepository: MockOnboardingRepository(pendingOnboarding: [.location], onMarkComplete: {
+                XCTAssertEqual($0, .location)
+                markCompleteExp.fulfill()
+            })
+        )
+        let exp = sut.inspection.inspect(after: 0.2) { view in
```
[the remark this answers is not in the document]
boringcactus: Apparently the wait isn't actually needed at all; I guess I just put it there instinctively since so many things do need it.

iosApp/iosApp/Pages/Onboarding/OnboardingScreenView.swift
```diff
@@ -0,0 +1,124 @@
...
+import CoreLocation
+import shared
+import SwiftUI
+
+struct OnboardingScreenView: View {
+    let screen: OnboardingScreen
+    let advance: () -> Void
+
+    let createLocationFetcher: () -> any LocationFetcher
+    let onboardingRepository: IOnboardingRepository
```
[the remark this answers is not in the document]
boringcactus: Sure.

iosApp/iosApp/Pages/Onboarding/OnboardingScreenView.swift
boringcactus: I have [confirmed that this is OK](https://mbta.slack.com/archives/C05QMG9GS9M/p1730221160659249?thread_ts=1730220856.035489&cid=C05QMG9GS9M).

KaylaBrady: Thanks for checking!

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -80,6 +80,12 @@ extension HomeMapView {
         lastNavEntry = navigationStack.last
     }
 
+    func checkOnboardingLoaded() {
+        if contentVM.onboardingScreensPending == [] {
+            locationDataManager.locationFetcher.requestWhenInUseAuthorization()
```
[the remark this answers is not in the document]
boringcactus: It's not obvious to me where or for how long we'd want to store that, or what the behavior should actually be in the meantime, so I've [asked in Slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1730222818020879).

boringcactus:
I think I want to build the location permission deferral as a separate PR.

KaylaBrady (approved):
@boringcactus latest changes look good to me! Building location permission deferral makes sense as a separate PR, assuming that will be merged at the same time as this + the UI changes

PR #501, opened 2024-10-29
refactor: handle settings as a map instead of a set of pairs

boringcactus:
### Summary

_Ticket:_ [Implement onboarding UX](https://app.asana.com/0/1205732265579288/1208468599574770/f)

Stacked on #497. Since this didn't wind up getting done in the More tab implementation, but it'll be valuable for location access deferral, I'm tossing it in here.

### Testing

Manually verified that settings still get read and written correctly (and that the toggle switches update properly and don't change places within a category).


iosApp/iosApp/ViewModels/SettingsViewModel.swift
```diff
@@ -61,14 +56,13 @@ class SettingsViewModel: ObservableObject {
...
-                    id: .featureFlags,
-                    items: settings.filter { $0.category == .featureFlags }.map { MoreItem.toggle(setting: $0) }
-                ),
+                MoreSection(id: .settings, items: [
+                    .toggle(setting: .hideMaps, value: settings[.hideMaps] ?? false),
+                ]),
+                MoreSection(id: .featureFlags, items: [
+                    .toggle(setting: .map, value: settings[.map] ?? false),
+                    .toggle(setting: .searchRouteResults, value: settings[.searchRouteResults] ?? false),
+                ]),
```
boringcactus, line 65: The settings get bridged as an unordered dictionary, meaning these toggle switches weren't in a deterministic order when I was keeping `.filter().map()`, and I figured that the simplest way to ensure they have a deterministic order is to just hardcode that order.

iosApp/iosApp/ViewModels/SettingsViewModel.swift
```diff
@@ -94,9 +88,10 @@ class SettingsViewModel: ObservableObject {
         }
     }
 
-    private func setSettings(_ settings: [Setting]) {
+    private func setSettings(_ settings: [Settings: Bool]) {
         Task {
-            try await settingsRepository.setSettings(settings: Set(settings))
+            try await settingsRepository.setSettings(settings: settings.mapValues { KotlinBoolean(bool: $0) })
+            await self.getSections()
```
boringcactus, line 94: Refreshing the sections here was necessary to get the toggle switches to update their state correctly - without it, the states were writing, but the new values weren't being read back. I'm not sure why that'd be a consequence of my changes.

EmmaSimon (approved):

PR #508, opened 2024-10-31
fix: ignore and hide hide maps setting for now

boringcactus:
### Summary

_Ticket:_ [Temporarily disable Hide Maps](https://app.asana.com/0/1205732265579288/1208660630503490/f)

Stacked on #506 since that overhauls some settings structure.

### Testing

Manually verified that maps are shown even when the hide maps setting was previously enabled.


iosApp/iosAppTests/Views/ContentViewTests.swift
```diff
@@ -150,12 +150,12 @@ final class ContentViewTests: XCTestCase {
         wait(for: [hasAppeared, loadConfigCallback], timeout: 5)
     }
 
-    @MainActor func testHidesMap() throws {
+    @MainActor func testDoesNotHideMap() throws {
```
boringcactus, line 153: I thought about removing or skipping this test, but inverting it instead will make it easier to remember to switch the test back when we restore the functionality.

boringcactus:
Not needed after #510.

PR #503, opened 2024-10-30
feat: Implement onboarding UIs

EmmaSimon:
### Summary

_Ticket:_ [Implement onboarding UI](https://app.asana.com/0/1205732265579288/1208585337751430/f)

Stacked onto #497

Simulator Screenshot - iPhone 15 - 2024-10-30 at 10 34 07 IMG_0036 Simulator Screenshot - iPhone 15 - 2024-10-30 at 10 34 20

WIP: Pending dark mode assets

PR #509, opened 2024-10-31
i18n: Add missing strings for localization

EmmaSimon:
### Summary

_Ticket:_ [Fill in missing comments for translators](https://app.asana.com/0/1205732265579288/1208643843115469/f)

Stacked onto #503 

Audited all strings in the app, I added a dummy language with all strings replaced with a specific character to make sure that everything was being properly translated, and updated a handful of places that weren't being replaced. I also tested using VoiceOver, in both english and the dummy language, to make sure all of the accessibility text was also being properly localized. Also added comments to any potentially ambiguous strings, and cleaned up some copy that snuck through the cracks.

### Testing

Manual testing with alternate locales and VoiceOver

### DIFF
--- a/iosApp/iosApp/ComponentViews/UpcomingTripAccessibilityFormatters.swift
+++ b/iosApp/iosApp/ComponentViews/UpcomingTripAccessibilityFormatters.swift
@@ -1,0 +1,147 @@
+//
+//  UpcomingTripAccessibilityFormatters.swift
+//  iosApp
+//
+//  Created by esimon on 10/31/24.
+//  Copyright © 2024 MBTA. All rights reserved.
+//
+
+import SwiftUI
+
+class UpcomingTripAccessibilityFormatters {
+    private let timeFormatter: DateFormatter = makeTimeFormatter()
+
+    public func boardingFirst(vehicleText: String) -> Text {
+        Text("\(vehicleText) boarding now",
+             comment: """
+             Describe that a vehicle is boarding now, as read aloud for VoiceOver users.
+             First value is the type of vehicle (bus, train, ferry). For example, 'bus boarding now'
+             """)
+    }
+
+    public func boardingOther() -> Text {
+        Text("and boarding now",
+             comment: """
+             The second or more arrival in a list of upcoming arrivals read aloud for VoiceOver users.
+              For example, '[bus arriving in 1 minute], and boarding now'
+             """)
+    }
+
+    public func arrivingFirst(vehicleText: String) -> Text {
+        Text("\(vehicleText) arriving now",
+             comment: """
+             Describe that a vehicle is arriving now, as read aloud for VoiceOver users.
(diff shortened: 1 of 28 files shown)


iosApp/iosApp/ComponentViews/UpcomingTripAccessibilityFormatters.swift
```diff
@@ -0,0 +1,147 @@
...
+//  UpcomingTripAccessibilityFormatters.swift
+//  iosApp
+//
+//  Created by esimon on 10/31/24.
+//  Copyright © 2024 MBTA. All rights reserved.
+//
+
+import SwiftUI
+
+class UpcomingTripAccessibilityFormatters {
```
KaylaBrady, line 11: 👍 thanks for separating this out!

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -1,49 +1,31 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
-    "**%@** affected stops" : {
+    "**%ld** affected stops" : {
```
KaylaBrady, line 4: question(non-blocking): can we preserve the translations we already have for the plural case here?

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -89,6 +71,28 @@
         }
       }
     },
+    "%@ %@ %@" : {
+      "comment" : "VoiceOver text for the vehicle status on the trip details page,\nex 'train approaching Alewife' or 'bus now at Harvard'",
```
KaylaBrady, line 75: suggestion (non-blocking): These seem tricky - maybe adding brackets around each piece in the comment so it is clear where one unit begins and another ends?

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -4041,45 +3797,8 @@
     "Terms of Use" : {
       "comment" : "Label for a More page link to the MBTA.com terms of use"
     },
-    "This vehicle is completing another trip." : {
```
KaylaBrady, line 4044: question(non-blocking): can we keep these translations?

KaylaBrady (approved):
Wow there was more missing than I had thought - thanks for auditing so thoroughly!! 

A couple minor questions about whether existing translations can be preserved.

Also one non-blocking suggestion - what do you think about adding something to the PR template about checking that any strings added have localizations / using `Text` views where possible? That can be a follow-up though and definitely doesn't need to block sending out this batch!

iosApp/iosApp/Localizable.xcstrings
EmmaSimon: I thought about it but since the original wasn't pluralized, I don't know which translations apply to the two/few/many/other forms for languages that have those, and filling in a single plural translation seems to fill in the remaining ones in english, which could potentially cause confusion/missed translations when we send them to the vendor.

PR #510, opened 2024-11-01
fix: Fix hide maps location changes

EmmaSimon:
### Summary

_Ticket:_ [Location doesn't update when Hide Maps is turned on](https://app.asana.com/0/1205732265579288/1208660630503484/f)

Fix an issue where the current location was never updating the nearby transit location when "Hide Maps" was toggled on.

### Testing

Added a test for new ViewportProvider helper function

### DIFF
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -167,6 +167,12 @@ struct ContentView: View {
                         nearbyVM.goBack()
                     }
                 }, content: coverContents)
+                .onAppear {
+                    viewportProvider.updateCameraState(locationDataManager.currentLocation)
+                }
+                .onChange(of: locationDataManager.currentLocation) { location in
+                    viewportProvider.updateCameraState(location)
+                }
         } else {
             mapSection
                 .sheet(
--- a/iosApp/iosAppTests/Fetchers/ViewportProviderTest.swift
+++ b/iosApp/iosAppTests/Fetchers/ViewportProviderTest.swift
@@ -86,6 +86,22 @@ final class ViewportProviderTest: XCTestCase {
         XCTAssertEqual(provider.viewport.camera?.zoom, newZoom)
     }
 
+    func testUpdateCameraLocationFromCoordinate() throws {
+        let updatedExp = expectation(description: "updated camera")
+        let startCenter: CLLocationCoordinate2D = .init(latitude: 0, longitude: 0)
+        let provider = ViewportProvider(viewport: .camera(center: startCenter, zoom: 16.0))
+
+        let newCenter: CLLocationCoordinate2D = .init(latitude: 1, longitude: 1)
+        provider.updateCameraState(CLLocation(latitude: 1, longitude: 1))
+        let cancelSink = provider.cameraStatePublisher.sink { cameraUpdate in
+            XCTAssertEqual(cameraUpdate.center, newCenter)
+            updatedExp.fulfill()
+        }
+
+        wait(for: [updatedExp], timeout: 1)
+        cancelSink.cancel()
+    }
+
     func testSavePanned() throws {
         let provider = ViewportProvider(viewport: .idle)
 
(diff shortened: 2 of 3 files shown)


boringcactus (approved):

iosApp/iosApp/ContentView.swift
```diff
@@ -167,6 +167,12 @@ struct ContentView: View {
                         nearbyVM.goBack()
                     }
                 }, content: coverContents)
+                .onAppear {
+                    viewportProvider.updateCameraState(locationDataManager.currentLocation)
+                }
+                .onChange(of: locationDataManager.currentLocation) { location in
+                    viewportProvider.updateCameraState(location)
+                }
```
KaylaBrady, line 175: suggestion (non-blocking): Might be worth a comment here about why this is necessary.

KaylaBrady, line 175: question: I'm a little nervous about this not not directly testing this onChange behavior since if it is accidentally removed at some point, we might not spot that right away in our day-to-day use of the app.

Would it be feasible to add a test that prevents any future regressions here?

iosApp/iosAppTests/Fetchers/ViewportProviderTest.swift
```diff
@@ -86,6 +86,22 @@ final class ViewportProviderTest: XCTestCase {
         XCTAssertEqual(provider.viewport.camera?.zoom, newZoom)
     }
 
+    func testUpdateCameraLocationFromCoordinate() throws {
```
KaylaBrady, line 89: 🙌

iosApp/iosApp/ContentView.swift
```diff
@@ -167,6 +167,12 @@ struct ContentView: View {
                         nearbyVM.goBack()
                     }
                 }, content: coverContents)
+                .onAppear {
+                    viewportProvider.updateCameraState(locationDataManager.currentLocation)
```
KaylaBrady, line 171: issue: it looks like the 2nd AC isn't addressed here - if I pan the map & then turn on hide maps, I am stuck at that panned location

iosApp/iosApp/ContentView.swift
```diff
@@ -167,6 +167,12 @@ struct ContentView: View {
...
+                }
+                .onChange(of: locationDataManager.currentLocation) { location in
+                    viewportProvider.updateCameraState(location)
```
KaylaBrady, line 174: potential issue: It wasn't part of the original AC, but I think it would be worth adding the Location Services button below the search bar in hidden map mode. Otherwise, a user who has hidden the map but didn't grand location permissions the first time would be stuck on Park St

KaylaBrady (commented):
Love the simple approach you took! A few small concerns about the behavior with map panning / Location Services off

iosApp/iosApp/ContentView.swift
EmmaSimon: Oh hm, I thought I tried to reproduce that and couldn't after the fix 🤔

iosApp/iosApp/ContentView.swift
KaylaBrady: [this HomeMapView test](https://github.com/mbta/mobile_app/blob/296f9a4fcaea015106b354b507351bff91885520/iosApp/iosAppTests/Views/HomeMapViewTests.swift#L95) involves changing the mocked locationDataManager's current location. That is definitely a simpler test set up than ContentView, so OK if it isn't really possible to cover this with the test with the limitations of ViewInspector - I would just more strongly suggest including the comment suggested above

iosApp/iosApp/ContentView.swift
EmmaSimon: Oh, I think this might only happen when location services is turned off. But if it is, then it's unclear what we would want to do instead, since we have no way to get a different location.

KaylaBrady: Oh looks like it only happens when location services is turned off. That may be fine behavior, in coordination with [showing the location services button](https://github.com/mbta/mobile_app/pull/510/files#r1826047141).

PR #516, opened 2024-11-01
i18n: Add possible CR status override strings

EmmaSimon:
### Summary

_Ticket:_ No ticket

@boringcactus flagged a few more possible prediction status text overrides we can get for CR. We're not handling them yet, but we want to include them in the localization file we send to the translators so that they're ready when we do.

### DIFF
--- a/iosApp/iosApp/Localizable.xcstrings
+++ b/iosApp/iosApp/Localizable.xcstrings
@@ -1016,6 +1016,10 @@
         }
       }
     },
+    "Arriving" : {
+      "comment" : "Status for a commuter rail train",
+      "extractionState" : "manual"
+    },
     "Autos Impeding Service" : {
       "comment" : "Possible alert cause",
       "localizations" : {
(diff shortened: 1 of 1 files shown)


iosApp/iosApp/Localizable.xcstrings
```diff
@@ -1016,6 +1016,10 @@
         }
       }
     },
+    "Arriving" : {
+      "comment" : "Status for a commuter rail train",
+      "extractionState" : "manual"
+    },
```
boringcactus, line 1022: We could probably plan to rewrite overridden-to-"Arriving" as the existing`ARR` status - and we could probably turn "Now boarding" into `BRD`, too.

EmmaSimon: Yeah, I was thinking that would probably be the way to go, but might as well request them in case.

boringcactus (approved):

PR #517, opened 2024-11-01
fix: always join alerts on app startup

KaylaBrady:
### Summary

Hotfix, no ticket

What is this PR for?

This PR fixes an issue where after the onboarding flow, users would get stuck on a shimmer loading screen. 
The cause was that by the time the nearby transit sheet popped after onboarding, there was no scene phase to trigger joining. Moved to the top of the view hierarchy so that it is always called on app startup.

### Testing

What testing have you done?

Tested locally & updated existing unit tests. I moved the map config handling logic to the same top level because fiddling with the list of `implicitAnyViews` was giving me trouble.

### DIFF
--- a/iosApp/iosAppTests/Views/ContentViewTests.swift
+++ b/iosApp/iosAppTests/Views/ContentViewTests.swift
@@ -37,8 +37,9 @@ final class ContentViewTests: XCTestCase {
 
         ViewHosting.host(view: sut)
 
-        try sut.inspect().implicitAnyView().view(ContentView.self).implicitAnyView().implicitAnyView().implicitAnyView()
-            .vStack().callOnChange(newValue: ScenePhase.background)
+        try sut.inspect().implicitAnyView().view(ContentView.self).implicitAnyView()
+            .vStack()
+            .callOnChange(newValue: ScenePhase.background)
         wait(for: [disconnectedExpectation], timeout: 5)
     }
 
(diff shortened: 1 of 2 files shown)


EmmaSimon (approved):

iosApp/iosAppTests/Views/ContentViewTests.swift
```diff
@@ -56,8 +57,9 @@ final class ContentViewTests: XCTestCase {
...
         ViewHosting.host(view: sut)
 
-        try sut.inspect().implicitAnyView().view(ContentView.self).implicitAnyView().implicitAnyView().implicitAnyView()
-            .vStack().callOnChange(newValue: ScenePhase.background)
+        try sut.inspect().implicitAnyView().view(ContentView.self).implicitAnyView()
+            .vStack()
+            .callOnChange(newValue: ScenePhase.background)
         wait(for: [disconnectedExpectation], timeout: 1)
         try sut.inspect().implicitAnyView().view(ContentView.self).implicitAnyView().implicitAnyView().implicitAnyView()
             .vStack().callOnChange(newValue: ScenePhase.active)
```
boringcactus, line 65: The second `callOnChange` in `testReconnectsSocketAfterBackgroundingAndReactivating` needs to also be trimmed down to one `implicitAnyView` for this test to pass in Xcode 16.

KaylaBrady: 👍 if you have that change locally, could you put up a separate PR for it and I'll ✅? That way we can get this merged & start the CI for a new build ASAP

KaylaBrady: (or I can put it up if you don't have it locally)

PR #518, opened 2024-11-01
test: fix an Xcode 16 issue

boringcactus:
### Summary

_Ticket:_ none

#517 introduced an Xcode 16 test issue, but that's lower-priority than getting those changes out the door.

### Testing

Verified that the test now passes in Xcode 16.


KaylaBrady (approved):
Thank you for catching!! I need to upgrade.

boringcactus:
Honestly, I wouldn't recommend it - some of the accessibility label tests that got added while I was stuck on https://github.com/mbta/mobile_app/pull/457 have been failing the whole time, and when I tried to add the right set of `.implicitAnyView()` calls I was getting _the wrong text_ back, so I'm kinda just hoping ViewInspector eventually figures out a way to not require `.implicitAnyView()` all over the place.

PR #524, opened 2024-11-05
feat: include dependency licenses

boringcactus:
### Summary

_Ticket:_ [Include licenses of dependencies in app](https://app.asana.com/0/1205732265579288/1208585337751423/f)

This was not a three point task.

- [X] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

Validated locally that iOS dependencies are displayed correctly. After a long battle with Xcode Cloud, verified that this works in CI.


BrandonTR (approved):
Awesome work!

PR #527, opened 2024-11-07
fix: Display error state when trip doesn't exist

EmmaSimon:
### Summary

_Ticket:_ [Fix trip details crash when trip not in service](https://app.asana.com/0/1205732265579288/1207687497283289/f)

We can get added trip IDs that don't exist in the API in trip details, which before these changes was resulting in shimmer loading forever. Worse, it would continue to shimmer load even after switching to a different vehicle on the map.

Now, the trip header will be displayed, and scheduled stops will be displayed if they exist. Also fixed a related issue where `zipPredictedSchedules` would throw an exception if the sequence of predictions didn't exactly match the sequence of scheduled stops, sometimes in the actual data this can happen though, so now it just skips the scheduled stop and keeps checking to see if the next ones match.

### Testing

Updated test behavior to check for new state. Tested locally with https://github.com/mbta/mobile_app_backend/pull/235


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -316,8 +320,9 @@ data class TripDetailsStopList(val stops: List<Entry>) {
                 while (scheduleIndex in stopIds.indices && predictionIndex in predictions.indices) {
                     val stopId = stopIds[scheduleIndex]
                     val prediction = predictions[predictionIndex]
-                    check(Stop.equalOrFamily(stopId, prediction.stopId, globalData.stops)) {
-                        "predictions=$predictions stopIds=$stopIds predictionIndex=$predictionIndex scheduleIndex=$scheduleIndex"
+                    if (Stop.equalOrFamily(stopId, prediction.stopId, globalData.stops)) {
+                        scheduleIndex--
+                        continue
                     }
```
boringcactus, line 326: Since `check(true)` succeeds and `check(false)` throws, shouldn't this be `if (!Stop.equalOrFamily(...`?

EmmaSimon: Oh, yes definitely. Weird that it seemed to be working okay with that 🤔

iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
```diff
@@ -71,9 +72,10 @@ struct TripDetailsPage: View {
     var body: some View {
         VStack(spacing: 16) {
             header
-            if let trip, let globalResponse, let vehicle = vehicleResponse?.vehicle,
+            if tripPredictionsLoaded, let globalResponse, let vehicle = vehicleResponse?.vehicle,
                let stops = TripDetailsStopList.companion.fromPieces(
```
boringcactus, line 76: There are tests that assert that `fromPieces` will return something reasonable if the predictions haven't loaded but the schedules have. Are we certain we want to hide schedules if predictions have not loaded?

iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
```diff
@@ -24,6 +24,7 @@ struct TripDetailsPage: View {
     @State var globalResponse: GlobalResponse?
     @State var tripPredictionsRepository: ITripPredictionsRepository
     @State var tripPredictions: PredictionsStreamDataResponse?
+    @State var tripPredictionsLoaded: Bool = false
```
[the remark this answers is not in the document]
EmmaSimon: Yes, when trip predictions [error](https://github.com/mbta/mobile_app/blob/main/iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift#L222) it remains null, that specific case is why this was added. We could instead use an ApiResult or something, but when the trip doesn't exist, we need to know that predictions have been attempted to have been fetched. This is related to the other thread, initially I was trying to not hide the schedules before predictions are loaded, but it results in the "x stops away" being inaccurate and flickering in with the wrong number before predictions can load.

https://github.com/user-attachments/assets/d69a627a-d13f-4b81-a90d-bbf8d8ce40d8

iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
EmmaSimon: Were you waiting on a reply to this? I tried to address it in the other thread.

boringcactus: Oh. I might have been, or I might have just lost track of this PR.

boringcactus (approved):

PR #533, opened 2024-11-08
fix: ensure gradle knows pods are installed

boringcactus:
### Summary

_Ticket:_ none

Builds on `main` are failing right now, and from the error message I think this is why.

### Testing

When I get the same error message locally, this is how I fix it, but I have no idea if it'll fix the non-test Xcode Cloud targets until we merge this (unless we want to temporarily add archive or analyze to the PR workflow).

### DIFF
--- a/iosApp/ci_scripts/ci_post_clone.sh
+++ b/iosApp/ci_scripts/ci_post_clone.sh
@@ -63,6 +63,7 @@ retry ./gradlew :shared:generateDummyFramework
 cd "${CI_PRIMARY_REPOSITORY_PATH}/iosApp"
 retry bundle exec pod install
 cd ..
+retry ./gradlew :shared:podInstallSyntheticIos
 
 # Install Node.js and GitHub CLI for codegen
 retry brew install node gh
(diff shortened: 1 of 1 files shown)


iosApp/ci_scripts/ci_post_clone.sh
```diff
@@ -63,6 +63,7 @@ retry ./gradlew :shared:generateDummyFramework
 cd "${CI_PRIMARY_REPOSITORY_PATH}/iosApp"
 retry bundle exec pod install
 cd ..
+retry ./gradlew :shared:podInstallSyntheticIos
```
EmmaSimon, line 66: Hmm, this seems to be failing because it can't find the pod install
```
Execution failed for task ':shared:podInstallSyntheticIos'.

>         ERROR: CocoaPods executable not found in your PATH.

          Please make sure CocoaPods is installed on your system.
```
Which is strange, since you can see the pod install complete directly above it. Could XCC be messing with the path somehow? Or could the dependency license changes have messed up gradle's path?

boringcactus: When we install CocoaPods via Bundler, it's not actually on the `PATH`, and `bundle exec` rewrites the `PATH` to include all the gems it manages. It might be sufficient to put the Gradle task inside `bundle exec`, though.

EmmaSimon (approved):

PR #536, opened 2024-11-12
fix: potential fix for blank map tiles

BrandonTR:
### Summary

_Ticket:_ [Fix blank / black map background](https://app.asana.com/0/1205732265579288/1208662152598557/f)

Potential fixes for blank map tiles, upgraded AppCheck/Firebase to the version with fixes for errors we're seeing. I also opted to make it so the map doesn't attempt to render until we retrieve a config.

### Testing

Modified an existing test and added a new test for the empty map grid state.

### DIFF
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -48,6 +48,7 @@ struct ContentView: View {
         VStack {
             contents
         }
+        .task { await contentVM.loadConfig() }
         .onReceive(inspection.notice) { inspection.visit(self, $0) }
         .onAppear {
             Task { await contentVM.loadOnboardingScreens() }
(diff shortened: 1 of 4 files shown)


iosApp/iosApp/ContentView.swift
```diff
@@ -170,13 +171,26 @@ struct ContentView: View {
...
-        HomeMapView(
-            contentVM: contentVM,
-            mapVM: mapVM,
-            nearbyVM: nearbyVM,
-            viewportProvider: viewportProvider,
-            sheetHeight: $sheetHeight
-        )
+        // If we don't have an access token, don't attempt to show the map
+        switch onEnum(of: contentVM.configResponse) {
+        case .ok:
```
KaylaBrady, line 176: I think we *would* want to show the map if the user doesn't have the access token, because there could be good reason they don't have an access token: they might be in a subway tunnel and have no access. With the current implementation where the map is always shown, a user can still see the stops, route shapes, and cached map tiles while offline.

KaylaBrady (approved):

PR #538, opened 2024-11-13
feat(DebugView): trip, stop, vehicle ids

KaylaBrady:
### Summary

_Ticket:_ [Debug setting to show (trip, stop, vehicle, etc) IDs](https://app.asana.com/0/1205732265579288/1208629979901034/f)

What is this PR for?

This PR adds the low hanging fruit ids - trip, stop, and vehicle. I opted to add each element in the detail view (stop in stop details, trip & vehicle in trip details). 

I didn't add anything around route patterns since our route pattern grouping logic is in flux - if we find it will be valuable, we can add that separately. I did take an initial pass at it and it would involve quite a bit of prop drilling unless we switch to using an EnvironmentObject to represent debug mode.

### Testing

What testing have you done?

Added some unit tests & ran locally. Also checked formatting to address the DebugView expanding unexpectedly as raised [in slack](https://mbta.slack.com/archives/C062QNAJZ2M/p1731098298241369)

### DIFF
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -51,6 +51,7 @@ struct ContentView: View {
         .onReceive(inspection.notice) { inspection.visit(self, $0) }
         .onAppear {
             Task { await contentVM.loadOnboardingScreens() }
+            Task { await nearbyVM.loadDebugSetting() }
         }
         .task {
             // We can't set stale caches in ResponseCache on init because of our Koin setup,
--- a/iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
+++ b/iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
@@ -75,7 +75,7 @@ struct TripDetailsPage: View {
             if nearbyVM.showDebugMessages {
                 DebugView {
                     VStack {
-                        Text(verbatim: "trip id \(tripId)")
+                        Text(verbatim: "trip id: \(tripId)")
                         Text(verbatim: "vehicle id: \(vehicleId)")
                     }
                 }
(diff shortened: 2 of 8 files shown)


iosApp/iosApp/ContentView.swift
```diff
@@ -51,6 +51,7 @@ struct ContentView: View {
         .onReceive(inspection.notice) { inspection.visit(self, $0) }
         .onAppear {
             Task { await contentVM.loadOnboardingScreens() }
+            Task { await nearbyVM.loadDebugSetting() }
```
EmmaSimon, line 54: I'm not seeing the debug info if I just toggle the setting on without restarting the app (and vice versa). I think the nearby VM needs to reload when the setting changes.

iosApp/iosApp/Pages/TripDetails/TripDetailsPage.swift
```diff
@@ -72,6 +72,14 @@ struct TripDetailsPage: View {
     var body: some View {
         VStack(spacing: 16) {
             header
+            if nearbyVM.showDebugMessages {
+                DebugView {
+                    VStack {
+                        Text(verbatim: "trip id \(tripId)")
```
EmmaSimon, line 78: nitpick: Add a `:` after trip id 😛

iosApp/iosApp/ContentView.swift
EmmaSimon: If you haven't looked at this yet, I based the combined stop and trip details changes off of this branch (partly to use the debug view, partly to use the existing debug loading in nearby VM), and I just fixed it as part of that, so don't worry about getting it working in this PR, it's fine as an interim state on a debug feature flag only we'll see.

KaylaBrady: Okay great, thank you!!

EmmaSimon (approved):

PR #540, opened 2024-11-14
feat: rearchitect everything to use trip headsigns

boringcactus:
### Summary

_Ticket:_ [Use trip headsigns in Nearby Transit](https://app.asana.com/0/1205732265579288/1208759166078934/f)

It probably makes sense to keep this behind a feature flag for a bit.

- [x] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

Manually verified that the new logic works correctly. Made existing unit tests parametric to check both old and new code paths. Added a new unit test to ensure that the new code path delivers on its stated goals.

I have not written unit tests that directly check the NearbyHierarchy itself; I could do that, but this PR is already too large.

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/StopDetailsView.swift
+++ b/iosApp/iosApp/Pages/StopDetails/StopDetailsView.swift
@@ -151,7 +151,8 @@ struct StopDetailsView: View {
         else { return }
         analytics.tappedRouteFilter(routeId: patterns.routeIdentifier, stopId: stop.id)
         let defaultDirectionId = patterns.patterns.flatMap { headsign in
-            headsign.patterns.map { pattern in pattern.directionId }
+            // RealtimePatterns.patterns is a List<RoutePattern?> but that gets bridged as [Any] for some reason
+            headsign.patterns.compactMap { pattern in (pattern as? RoutePattern)?.directionId }
         }.min() ?? 0
         setFilter(.init(routeId: filterId, directionId: defaultDirectionId))
     }
--- a/iosApp/iosApp/ViewModels/SettingsViewModel.swift
+++ b/iosApp/iosApp/ViewModels/SettingsViewModel.swift
@@ -95,10 +95,8 @@ class SettingsViewModel: ObservableObject {
                         value: settings[.searchRouteResults] ?? false
                     ),
                     .toggle(
-                        label: NSLocalizedString(
-                            "Trip Headsigns",
-                            comment: "A setting on the More page to use the destination/headsign on each individual trip in Nearby Transit (only visible for developers)"
-                        ),
+                        // not localized since it's a feature flag
+                        label: "Trip Headsigns",
                         setting: .tripHeadsigns,
                         value: settings[.tripHeadsigns] ?? false
                     ),
--- a/iosApp/iosApp/ViewModels/SettingsViewModel.swift
+++ b/iosApp/iosApp/ViewModels/SettingsViewModel.swift
@@ -94,6 +94,14 @@ class SettingsViewModel: ObservableObject {
                         setting: .searchRouteResults,
                         value: settings[.searchRouteResults] ?? false
                     ),
+                    .toggle(
+                        label: NSLocalizedString(
+                            "Trip Headsigns",
+                            comment: "A setting on the More page to use the destination/headsign on each individual trip in Nearby Transit (only visible for developers)"
+                        ),
+                        setting: .tripHeadsigns,
+                        value: settings[.tripHeadsigns] ?? false
+                    ),
                 ]),
                 MoreSection(id: .other, items: [
                     .link(
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
@@ -1,0 +1,457 @@
+package com.mbta.tid.mbta_app.model
+
+import com.mbta.tid.mbta_app.model.NearbyHierarchy.DirectionOrHeadsign
+import com.mbta.tid.mbta_app.model.NearbyHierarchy.LineOrRoute
+import com.mbta.tid.mbta_app.model.NearbyHierarchy.NearbyLeaf
+import com.mbta.tid.mbta_app.model.NearbyHierarchy.RouteAtStop
+import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.model.response.NearbyResponse
+import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
+import com.mbta.tid.mbta_app.model.response.ScheduleResponse
+import io.github.dellisd.spatialk.geojson.Position
+import io.github.dellisd.spatialk.turf.ExperimentalTurfApi
+import io.github.dellisd.spatialk.turf.distance
+import kotlinx.datetime.Instant
+
+private typealias InnerLayer = Map<DirectionOrHeadsign, NearbyLeaf>
+
+private typealias MiddleLayerData = RouteAtStop<List<Direction>, InnerLayer>
+
+private typealias MiddleLayer = Map<Stop, MiddleLayerData>
+
+private typealias OuterLayer = Map<LineOrRoute, MiddleLayer>
+
+data class NearbyHierarchy(private val data: OuterLayer) {
+    sealed interface LineOrRoute {
+        data class Line(
+            val line: com.mbta.tid.mbta_app.model.Line,
+            val routes: Set<com.mbta.tid.mbta_app.model.Route>
+        ) : LineOrRoute
+
+        data class Route(val route: com.mbta.tid.mbta_app.model.Route) : LineOrRoute
+    }
+
+    data class RouteAtStop<StopData, T>(val stopData: StopData, val data: T) {
+        fun <U> map(f: (T) -> U): RouteAtStop<StopData, U> = RouteAtStop(stopData, f(data))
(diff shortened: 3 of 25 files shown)


iosApp/iosApp/Pages/StopDetails/StopDetailsView.swift
```diff
@@ -151,7 +151,8 @@ struct StopDetailsView: View {
         else { return }
         analytics.tappedRouteFilter(routeId: patterns.routeIdentifier, stopId: stop.id)
         let defaultDirectionId = patterns.patterns.flatMap { headsign in
-            headsign.patterns.map { pattern in pattern.directionId }
+            // RealtimePatterns.patterns is a List<RoutePattern?> but that gets bridged as [Any] for some reason
```
JackVCurtis, line 154: Huh. Something to keep an eye on maybe, that seems real weird.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
```diff
@@ -63,7 +63,7 @@ sealed class RealtimePatterns {
         val route: Route,
         val headsign: String,
         val line: Line?,
-        override val patterns: List<RoutePattern>,
+        override val patterns: List<RoutePattern?>,
```
KaylaBrady, line 66: question: When would this be null?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -627,27 +627,157 @@ fun NearbyStaticData.withRealtimeInfo(
...
-        schedules,
-        predictions,
-        alerts,
-        filterAtTime,
-        showAllPatternsWhileLoading = false,
-        hideNonTypicalPatternsBeyondNext = 90.minutes,
-        filterCancellations = true,
-        pinnedRoutes
-    )
+    if (useTripHeadsigns) {
```
KaylaBrady, line 754: 🙌 appreciate this feature flagging!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
```diff
@@ -0,0 +1,433 @@
...
+        ): PartialHierarchy<SD2, T> =
+            data
+                .mapValues { (lineOrRoute, routeData) ->
+                    routeData.mapValues { (stop, routeAtStop) ->
+                        RouteAtStop(
+                            f(lineOrRoute, stop, routeAtStop.stopData, routeAtStop.data),
+                            routeAtStop.data
+                        )
+                    }
+                }
```
KaylaBrady, line 92: suggestion: documentation for these various map functions would be helpful. The difference between these two is particularly confusing to me on first read, as is the significance of `SD2` as a generic param

iosApp/iosApp/ViewModels/SettingsViewModel.swift
```diff
@@ -94,6 +94,14 @@ class SettingsViewModel: ObservableObject {
                         setting: .searchRouteResults,
                         value: settings[.searchRouteResults] ?? false
                     ),
+                    .toggle(
+                        label: NSLocalizedString(
+                            "Trip Headsigns",
+                            comment: "A setting on the More page to use the destination/headsign on each individual trip in Nearby Transit (only visible for developers)"
```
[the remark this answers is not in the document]
boringcactus: Strictly speaking, `verbatim:` is a SwiftUI `Text` argument and not something `MoreItem.toggle` can do, but since `MoreItem.toggle` takes just `label: String` it's absolutely possible to just pass a literal string directly. (If that weren't the case, it'd also be possible to define this string as don't translate in the `Localizable.xcstrings` editor; I thought about it but decided not to start setting that for this feature flag since we already translated other feature flags.)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
boringcactus: If there's an added trip with no route pattern that gets sorted into this `RealtimePatterns`, then `patterns` will contain `null`.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
```diff
@@ -0,0 +1,433 @@
...
+
+private typealias InnerLayer = Map<DirectionOrHeadsign, NearbyLeaf>
+
+private typealias MiddleLayerData = RouteAtStop<List<Direction>, InnerLayer>
+
+private typealias MiddleLayer = Map<Stop, MiddleLayerData>
+
+private typealias OuterLayer = Map<LineOrRoute, MiddleLayer>
+
+data class NearbyHierarchy(private val data: OuterLayer) {
```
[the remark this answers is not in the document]
boringcactus: Added a doc comment to every nontrivial public declaration in this file.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
```diff
@@ -0,0 +1,433 @@
...
+import io.github.dellisd.spatialk.turf.distance
+import kotlinx.datetime.Instant
+
+private typealias InnerLayer = Map<DirectionOrHeadsign, NearbyLeaf>
+
+private typealias MiddleLayerData = RouteAtStop<List<Direction>, InnerLayer>
+
+private typealias MiddleLayer = Map<Stop, MiddleLayerData>
+
+private typealias OuterLayer = Map<LineOrRoute, MiddleLayer>
```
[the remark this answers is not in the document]
boringcactus: Renamed.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
```diff
@@ -0,0 +1,433 @@
...
+    sealed interface DirectionOrHeadsign {
+        data class Direction(val direction: com.mbta.tid.mbta_app.model.Direction) :
+            DirectionOrHeadsign
+
+        data class Headsign(val headsign: String, val route: Route, val line: Line?) :
+            DirectionOrHeadsign
+    }
+
+    data class PartialHierarchy<StopData, T>(
+        val data: Map<LineOrRoute, Map<Stop, RouteAtStop<StopData, T>>>
```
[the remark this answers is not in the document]
boringcactus: The route isn't part of the structure there, `RouteAtStop` is just the only way to attach the `List<Direction>` to the entire `ByDirectionOrHeadsign`. If you think it'd be clearer to just use a `Pair`, I can, and if you can think of a better name, I'd be happy to hear it.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
boringcactus: `U` is a natural successor to `T`, so the type argument for `mapInContext` was easy, but there's no obvious natural successor to `StopData`, so my inelegant compromise is `SD2`. If you think `StopData2` would be clearer, or some different name for `StopData` would be helpful, I can change those around.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
KaylaBrady: Ah, right! May be worth a comment about that edge case

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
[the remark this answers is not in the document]
KaylaBrady: Got it. I think it could be clearer to have separate types for the final state from the intermediate state, along the lines of `ByStopData(directions: List<Direction>, data: ByDirectionOrHeadsign)` defined as a class and PartialByStopData(directions: T, data: U)`. That way, when working with the permanent structure, the field names are clearer. 

Taking that thought further - if there is only one source for the directions and it isn't really being combined with anything, could that be part of the fixed hiearchy structure and removed from the partial zipping entirely?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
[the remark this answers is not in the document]
KaylaBrady: That helped my understanding of this change a lot, thank you!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RealtimePatterns.kt
boringcactus: Commented.

KaylaBrady (approved):
Nice work!!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyHierarchy.kt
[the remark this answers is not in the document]
boringcactus: If we put the directions into the map key, then merging the stop and its directions from the scheduled data with the stop without directions from the realtime data gets a lot more complicated.

PR #541, opened 2024-11-15
fix: remove AppCheck

BrandonTR:
### Summary

_Ticket:_ [remove appcheck](https://app.asana.com/0/1205732265579288/1208768145496164/f)

Removing AppCheck


iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,62 +6,62 @@
...
       "location" : "https://github.com/google/abseil-cpp-binary.git",
       "state" : {
-        "revision" : "748c7837511d0e6a507737353af268484e1745e2",
-        "version" : "1.2024011601.1"
+        "revision" : "194a6706acbd25e4ef639bcaddea16e8758a3e27",
+        "version" : "1.2024011602.0"
       }
     },
     {
       "identity" : "app-check",
```
KaylaBrady, line 14: question (non-blocking): Can this dependency be removed now, or does it get pulled in by default with firebase?

iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,62 +6,62 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/google/abseil-cpp-binary.git",
       "state" : {
-        "revision" : "748c7837511d0e6a507737353af268484e1745e2",
```
KaylaBrady, line 9: question: are these package resolution changes expected?

KaylaBrady (approved):
Small q about the package.resolve file, otherwise LGTM! 🎉

iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
boringcactus: I think this is the project lockfile instead of the workspace lockfile, and we usually change the workspace lockfile but not the project lockfile. (Why there are two of them that are maintained independently is anyone's guess.)

iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
BrandonTR: I did remove AppCheck as a dependency from the app so it seems like this is just getting pulled in by default now

iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
BrandonTR: I am also not sure why there is two, these changes look like they're correct though just based on the version updates from when I bumped AppCheck

PR #550, opened 2024-11-19
use predictions v2 and fix bug in decodeRawMessage

JackVCurtis:
### Summary

_Ticket:_ [Use predictions v2 channel in Android app](https://app.asana.com/0/1205732265579288/1208501775350660/f)

What is this PR for?

Use the same, more scalable endpoint as iOS for the predictions real time data. This PR also fixes a parsing issue with the `decodeRawMessage` function.

- [x] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

What testing have you done?

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/SocketMessage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/SocketMessage.kt
@@ -1,42 +1,25 @@
 package com.mbta.tid.mbta_app.android.util
 
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonPrimitive
 import org.phoenixframework.Message
 
 fun decodeMessage(rawMessage: String): Message {
-    // https://github.com/dsrees/JavaPhoenixClient/blob/1.3.1/src/main/kotlin/org/phoenixframework/Defaults.kt#L68
-    val parseValue: (String) -> String? = { value ->
-        when (value) {
-            "null" -> null
-            else -> value.replace("\"", "")
-        }
-    }
+    val message = Json.parseToJsonElement(rawMessage).jsonArray
+    val joinRef = message[0].jsonPrimitive.content
 
-    var message = rawMessage
-    message = message.removeRange(0, 1) // remove '['
+    val ref = message[1].jsonPrimitive.content
+    val topic = message[2].jsonPrimitive.content
+    val event = message[3].jsonPrimitive.content
 
-    val joinRef = message.takeWhile { it != ',' } // take "join ref", "null" or "\"5\""
-    message = message.removeRange(0, joinRef.length) // remove join ref
-    message = message.removeRange(0, 1) // remove ','
-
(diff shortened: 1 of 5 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/SocketMessage.kt
```diff
@@ -1,42 +1,25 @@
...
-    val topic = message.takeWhile { it != ',' } // take topic, "\"topic\""
-    message = message.removeRange(0, topic.length)
-    message = message.removeRange(0, 1) // remove ','
-
-    val event = message.takeWhile { it != ',' } // take event, "\"phx_reply\""
-    message = message.removeRange(0, event.length)
-    message = message.removeRange(0, 1) // remove ','
-
-    val payloadJson = message.removeRange(message.length - 1, message.length) // remove ']'
+    val payloadJson = message[4].toString()
```
boringcactus, line 16: I'm nervous about the performance impact of turning this back into a JSON string just so that we can parse it again as JSON a few moments later, but I'm not really sure how straightforward it'd be to avoid that, so it shouldn't have to block this PR.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/SocketMessage.kt
```diff
@@ -1,42 +1,25 @@
...
 fun decodeMessage(rawMessage: String): Message {
-    // https://github.com/dsrees/JavaPhoenixClient/blob/1.3.1/src/main/kotlin/org/phoenixframework/Defaults.kt#L68
-    val parseValue: (String) -> String? = { value ->
-        when (value) {
-            "null" -> null
-            else -> value.replace("\"", "")
-        }
-    }
+    val message = Json.parseToJsonElement(rawMessage).jsonArray
+    val joinRef = message[0].jsonPrimitive.content
```
boringcactus, line 10: Per [the `JsonPrimitive.content` docs](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-primitive/content.html), `content` will be `"null"` if the value is `null`, and [`contentOrNull`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/content-or-null.html) is the correct way to get `null` from `null`. (The Kotlin serialization library stores JSON internally in a weird half-hydrated state that is optimized specifically for serialization and deserialization more than for in-memory processing, which I learned way too much about as part of [FeatureProperties](https://github.com/mbta/mobile_app/blob/78e046e2de357c0d7406701054ed51d0163e758a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/style/FeatureProperties.kt#L5-L11).)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/SocketMessage.kt
JackVCurtis: Sure, I don't expect it'll be an issue but I'm digging into Android performance again so I'll keep this in mind as a potential problem. Can always revisit.

boringcactus (approved):

PR #552, opened 2024-11-21
feat(Android): create separate staging app

boringcactus:
### Summary

_Ticket:_ [Set up separate staging and prod Android builds](https://app.asana.com/0/1205732265579288/1208788669093059/f)

I'm surprised that the hardest part of this was finding the Android Studio option to switch between flavors (which I've added to the README).

I've dusted off my bad placeholder staging app icon and set it up with the same background/foreground setup as when I imported the placeholder prod app icon back in the day. I've gone ahead and updated the names on both to match the iOS app names.

We can't ship updates to the Google Play staging app since it doesn't exist, so the CI is still pointing at prod. I've submitted the infra request to create the staging app in Google Play.

None of these assets actually need to be rendered as multi-DPI PNGs; I may have been too eager to automate that setup in the beginning. I have created a [separate task](https://app.asana.com/0/1208621177839510/1208811125895014/f) to resolve this mess.

### Testing

Manually verified that the apps can be installed in parallel and talk to the right backend.


EmmaSimon (approved):
Oh wow, yeah this is surprisingly straightforward!

PR #553, opened 2024-11-22
fix(Android): only render map-related iOS icons as PNGs

boringcactus:
### Summary

_Ticket:_ [[Android] Sort out assets mess](https://app.asana.com/0/1208621177839510/1208811125895014/f)

We have reached a point where there are enough non-map-related icons in the iOS app that automatically rendering all of them to PNGs for the Android app is silly.

### Testing

Added a new `DrawableByNameTest` ensuring that every function that had previously been calling `drawableByName` can still pass in all the values it possibly could.


JackVCurtis (approved):
👍

PR #556, opened 2024-11-25
refactor: Move some stop details logic into a VM

EmmaSimon:
### Summary

_Ticket:_ [Combined stop + trip details: departure list UI](https://app.asana.com/0/1205732265579288/1208725410378678/f)

Discussed in https://github.com/mbta/mobile_app/pull/551#discussion_r1854636220, this moves a chunk of the logic in the stop details extension into a new stop details VM. I didn't include any logic that touched other VMs (nearby and errordetails), so about half of the logic still lives in the extension.

This should probably also include any required trip details calls, but we can do that as part of [Combined stop + trip details: Trip details UI](https://app.asana.com/0/1205732265579288/1208725410378682/f).

~- [ ] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?~

### Testing

Updated tests to use the VM instead of passing repos directly into the view

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
+++ b/iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
@@ -99,7 +99,7 @@ struct StopDetailsPage: View {
                     joinPredictions()
                 },
                 onInactive: stopDetailsVM.leavePredictions,
-                onBackground: { @MainActor in
+                onBackground: {
                     stopDetailsVM.leavePredictions()
                     errorBannerVM.loadingWhenPredictionsStale = true
                 }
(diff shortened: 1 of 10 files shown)


iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -118,19 +92,15 @@ struct StopDetailsPage: View {
...
-                    }
-                    joinPredictions(stopId)
+                    stopDetailsVM.returnFromBackground()
+                    joinPredictions()
                 },
-                onInactive: leavePredictions,
-                onBackground: {
-                    leavePredictions()
+                onInactive: stopDetailsVM.leavePredictions,
+                onBackground: { @MainActor in
```
boringcactus, line 102: I'm curious if it'd work to make these callbacks `@MainActor` in `withScenePhaseHandlers` instead of specifying that this specific one is `@MainActor` - I'd guess that several of these need to be on the main thread and none of them need to be off the main thread - but if that doesn't work then this seems like it's fine.

boringcactus (approved):

iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
EmmaSimon: Yeah good call, updated.

PR #560, opened 2024-11-27
feat: Android more tab

KaylaBrady:
### Summary

_Ticket:_ [[Android] set up more page](https://app.asana.com/0/1205732265579288/1208833570216723/f)

What is this PR for?
Adds basic more tab.

This maintains the structure of NearbyTransitPage and MorePage having separate BottomNavBars. 
We could potentially pull the bar up a level as described in the [Navigate with Compose ](https://developer.android.com/develop/ui/compose/navigation)tutorial, but we don't always want the bar to show (NearbyTransitPage manages the logic of hiding the bar for StopDetails, and leaving that within NearbyTransitPage feels appropriate). 

- [ ] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

What testing have you done?
* Ran locally and switched back and forth between more and nearby transit successfully. 
image


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -153,6 +149,10 @@ fun NearbyTransitPage(
                             }
 
                             LaunchedEffect(navRoute) {
+                                if (navBarVisible) {
```
KaylaBrady, line 152: I noticed that when navigating between Nearby Transit & Stop Details, the bottom nav bar would flicker in and out. This seems to resolve the problem

JackVCurtis (approved):
Looks great!

PR #566, opened 2024-12-04
fix: updated filter logic

BrandonTR:
### Summary

_Ticket:_ [Always show headsign that is canonically not arrival-only in Nearby Transit](https://app.asana.com/0/1205732265579288/1208690341598924/f)

This aims to resolve the issues we were seeing with headsigns not showing up at alert boundaries.
Before:
Simulator Screenshot - iPhone 16 Pro - 2024-12-06 at 15 12 28

After:
Simulator Screenshot - iPhone 16 Pro - 2024-12-06 at 15 09 38

### Testing

Tweaked any affected tests and included Kayla's Oak Grove/North Station test captured during the disruption.


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/NearbyResponseTest.kt
```diff
@@ -2919,4 +2847,281 @@ class NearbyResponseTest {
...
+                childStopIds = listOf("70001")
+            }
+
+        val forestHillsPlatform =
+            objects.stop {
+                id = "70001"
+                locationType = LocationType.STOP
+                parentStationId = "place-forhl"
+            }
+
```
KaylaBrady, line 2974: I think the test passes as-is now because isTypical = true, isUpcoming = false, isLastStopOnRoutePattern = true, isArrivalOnly = false (because there are no predictions).

For this test we want to include a prediction order to test the isArrivalOnly piece. 
```

        val northboundPrediction =
            objects.prediction {
                trip = orangeNorthboundTypicalTrip
                stopId = northStationNorthboundPlatform.id
                stopSequence = 130
                arrivalTime = time + 8.minutes
                departureTime = null
            }
```
When that is added, the test fails.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt
```diff
@@ -568,14 +569,20 @@ fun NearbyStaticData.withRealtimeInfoWithoutTripHeadsigns(
...
                 null -> this.isUpcoming()
                 else -> this.isUpcomingWithin(filterAtTime, cutoffTime)
             }
-        return (isTypical() || isUpcoming) && !isArrivalOnly()
+        val isLastStopOnRoutePattern =
+            stopPatterns.patterns.any {
+                val lastStopIdInPattern = it.stopIds.last()
+                val stop = stopPatterns.stop
+                lastStopIdInPattern == stop.id || stop.childStopIds.contains(lastStopIdInPattern)
+            }
```
KaylaBrady, line 584: Sorry I missed this before!! I think part of the problem is that `stopPatterns.patterns` contains the patterns in *both* directions, whereas we only want to consider the patterns in the current direction being considered (In this example, looking only at the Northbound Oak Grove route pattern, not Southbound to Forest Hills).

`this.patterns` would be the better fit for that purpose. We can use the global data to resolve the ordered list of stops for each pattern in `this.patterns` `representativeTripId`

KaylaBrady (approved):
Nice work! Could you remove the old OL screenshots and add a new GL one for posterity?

PR #568, opened 2024-12-04
refactor: port iOS location architecture to Android

boringcactus:
### Summary

_Ticket:_ [🤖 | Map | Map => Nearby Transit location sharing architecture](https://app.asana.com/0/1201654106676769/1208730131933864/f)

The log gets spammed with "sending message to a Handler on a dead thread" via some Google Play location stuff, but only if location permission was already granted when the app first launched. I tried to investigate this, but I don't think anything is actually running on a different thread than it's supposed to, and it doesn't appear to be interfering with the actual functionality.

android

### Testing

Manually verified that both follow puck and manually select states properly track the current location.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/LocationDataManager.kt
```diff
@@ -0,0 +1,148 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+open class LocationDataManager {
+    private val _currentLocation = MutableStateFlow<Location?>(null)
+
+    var hasPermission by mutableStateOf(false)
+
+    /**
+     * Attach the event handlers for this [LocationDataManager] in the context of the current
```
JackVCurtis, line 41: Thanks for the comment, much appreciated!

BrandonTR (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -282,17 +292,108 @@ fun HomeMapView(
 
             val context = LocalContext.current
 
+            val locationProvider = remember {
+                object : LocationProvider {
```
JackVCurtis, line 296: Could we move this to it's own file? `HomeMapView` is pretty big already, I think it would help readability to pull this out. Nonblocking!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -282,17 +292,108 @@ fun HomeMapView(
...
+                onStopOrDispose {}
+            }
+
+            LifecycleStartEffect(Unit) {
+                onStopOrDispose { viewportProvider.saveCurrentViewport() }
+            }
+
+            DisposableMapEffect { map ->
+                val listener =
+                    object : OnMoveListener, OnScaleListener, OnShoveListener {
```
JackVCurtis, line 350: I also think this would benefit from being moved out of the file

JackVCurtis (approved):
Looks great with a couple suggestions. I agree about the log message, it's low priority even if it's not some weird simulator thing.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/LocationDataManager.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.google.android.gms.location.LocationServices
+import com.google.android.gms.location.LocationSettingsRequest
+import com.google.android.gms.location.Priority
+import com.mbta.tid.mbta_app.android.util.LocalActivity
+import com.mbta.tid.mbta_app.android.util.LocalLocationClient
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+open class LocationDataManager {
```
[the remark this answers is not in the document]
boringcactus: This has to be `open` so that the `MockLocationDataManager` can inherit from it. I could've made `LocationDataManager` an interface instead, but I don't like the pattern of making everything a redundant interface; that said, we already do that for all our dependency injection repositories, so if you'd rather I also do that here for consistency I could make that change.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
boringcactus: Moved.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
boringcactus: Moved.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/LocationDataManager.kt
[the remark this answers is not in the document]
JackVCurtis: Nah, I didn't realize it was for testing. That's fine IMO

PR #571, opened 2024-12-06
fix(DownstreamAlerts): Exclude the target stop's alerts from downstream calc

KaylaBrady:
### Summary
Fix for [Display downstream disruptions](https://app.asana.com/0/1205732265579288/1208643635656242/f)

What is this PR for?

This PR fixes an issue where within suspensions, users would see the alert for their stop in addition to a downstream "Suspension ahead" alert. 

When calculating downstream alerts, we should exclude alerts at the current stop from consideration.

iOS
android

### Testing

What testing have you done?
* Added unit test
* Ran locally, confirmed downstream alerts no longer showing up within the bounds of the current GL suspension alert
IMG_0213


JackVCurtis (approved):
👍

PR #573, opened 2024-12-06
feat(Android): draw crosshair at selected location

boringcactus:
### Summary

_Ticket:_ [🤖 | Map | Crosshairs when map dragged](https://app.asana.com/0/1201654106676769/1208850994085736/f)

Adds the crosshair on the map.

iOS
android

### Testing

Validated that the crosshair is being shown while panning, after panning, and during the momentum of a fling. The crosshair briefly flickers when it crosses over from outside the map to inside the map, but that's also the case on iOS.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -336,12 +343,20 @@ fun HomeMapView(
...
 
-            if (!viewportProvider.isFollowingPuck && lastNearbyTransitLocation != null) {
-                CircleAnnotation(
-                    point = lastNearbyTransitLocation.toPoint(),
-                    circleColorString = "#ba75c7",
-                    circleRadius = 10.0
-                )
+            if (
+                isNearbyNotFollowing &&
+                    lastNearbyTransitLocation != null &&
```
JackVCurtis, line 348: 🎱 (stylistic) `hasStableNearbyTransitLocation` would be help me here, I blame Ruby for my dislike of long booleans.

JackVCurtis (approved):
Looks good! It would be good at some point to address the question of testing things that show up on the map, but way out of scope here.

BrandonTR (approved):
nice!

PR #575, opened 2024-12-09
feat(android): Dynamic mapbox token

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Map | Dynamic mapbox token on Android](https://app.asana.com/0/1205732265579288/1207910202706568/f)

What is this PR for?
Dynamically configures the mapbox token on android, same as iOS
**Note**
I put all of this functionality into a new `MapViewModel`, whereas on iOS it is spread across `MapViewModel` and `ContentViewModel`. Since there isn't an analogous of `ContentViewModel` for android and this functionality is exclusively map related, this felt like a small and appropriate break in consistency. 

iOS
android

### Testing

What testing have you done?
* Ran locally pointing to[ this branch](https://github.com/mbta/mobile_app_backend/pull/255) that rotates tokens more quickly, verified I could load map tiles on android & continue loading after panning the map
* Added unit tests. There is one test I'm not sure about how to write yet, flagging with comment


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -112,6 +116,10 @@ fun NearbyTransitPage(
         vehiclesRepository.disconnect()
     }
 
+    LaunchedEffect(mapViewModel.lastMapboxErrorTimestamp.collectAsState(initial = null).value) {
+        mapViewModel.loadConfig()
+    }
```
KaylaBrady, line 121: I'm not sure how to write a test that covers this - looking more closely at existing tests to see if there is an analogous case already

JackVCurtis (approved):
👍 One idea around the testing issue

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
KaylaBrady: Thanks for the suggestion! It ended up being a little hairy to mock at that deeper level, but it got me thinking about mocking at the view model level which worked!

PR #578, opened 2024-12-10
docs: explain troubleshooting for bad PATH

boringcactus:
### Summary

_Ticket:_ [Document Xcode build troubleshooting](https://app.asana.com/0/1205425564113216/1208924185049483/f)

This covers all the issues of this sort that I run into on a regular basis.

iOS

android

### Testing

Checked that `fix-cocoapods.sh` works regardless of the directory it's run from.


bin/fix-cocoapods.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env zsh
+set -e
+
+# Installs the CocoaPods dependencies and updates the Gradle cache to recognize that the
+# dependencies are installed.
```
boringcactus, line 5: My personal copy of this script also has `git restore iosApp/iosApp.xcworkspace/xcshareddata/swiftpm/Package.resolved`, but I'm not sure how often that's actually needed when running this script, and I remember needing it when just switching Git branches and not messing with CocoaPods at all, although now that I'm thinking more about it I'm not sure that I've run into that issue since updating to Xcode 15 (although I run into other issues often enough that I still wouldn't recommend it).

iosApp/Podfile.lock
```diff
@@ -22,4 +22,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 32413a7fdb16d40c9f57cf8da0c74d4507ff5f90
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.16.2
```
boringcactus, line 25: I'm sneaking this in here since we may as well sneak it in somewhere - as discussed at some point, it's not just me who's been manually keeping this change out of my Git commits for a while.

EmmaSimon (approved):
This looks great, thanks for putting this together!

PR #579, opened 2024-12-10
feat(android.NearbyTransitPage): leave / rejoin vehicles channel after backgrounding

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Stability | Leave realtime channels when backgrounding](https://app.asana.com/0/1205732265579288/1208725410378690/f)

What is this PR for?

This PR is one slice of the above ticket, focused only on the vehicles channel for early feedback. I plan to extend this approach to the existing `subscribeToPredictions` and `subscribeToAlerts` channels.

iOS
android

### Testing

What testing have you done?
* Ran locally with logs in place, confirmed that vehicles channel properly opens & closes when changing routes and when backgrounding / restoring.
* Added tests for pause / resume events

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -82,6 +82,7 @@ dependencies {
     implementation(libs.mapbox.turf)
     implementation(libs.okhttp)
     implementation(libs.playServices.location)
+    implementation(libs.androidx.lifecycle.runtime.testing)
     debugImplementation(platform(libs.compose.bom))
     debugImplementation(libs.compose.ui.test.manifest)
     debugImplementation(libs.compose.ui.tooling)
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/ContentView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/ContentView.kt
@@ -6,6 +6,7 @@ import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.rememberBottomSheetScaffoldState
 import androidx.compose.material3.rememberStandardBottomSheetState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -24,22 +25,26 @@ import com.mbta.tid.mbta_app.android.location.rememberLocationDataManager
 import com.mbta.tid.mbta_app.android.pages.MorePage
 import com.mbta.tid.mbta_app.android.pages.NearbyTransit
 import com.mbta.tid.mbta_app.android.pages.NearbyTransitPage
+import com.mbta.tid.mbta_app.android.pages.OnboardingPage
 import com.mbta.tid.mbta_app.android.phoenix.PhoenixSocketWrapper
 import com.mbta.tid.mbta_app.android.state.getGlobalData
 import com.mbta.tid.mbta_app.android.state.subscribeToAlerts
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
 import com.mbta.tid.mbta_app.network.PhoenixSocket
 import io.github.dellisd.spatialk.geojson.Position
+import org.koin.androidx.compose.koinViewModel
 import org.koin.compose.koinInject
 
 @OptIn(MapboxExperimental::class, ExperimentalMaterial3Api::class)
 @Composable
 fun ContentView(
     socket: PhoenixSocket = koinInject(),
+    viewModel: ContentViewModel = koinViewModel(),
 ) {
     val navController = rememberNavController()
-    var alertData: AlertsStreamDataResponse? = subscribeToAlerts()
+    val alertData: AlertsStreamDataResponse? = subscribeToAlerts()
     val globalResponse = getGlobalData()
+    val pendingOnboarding = viewModel.pendingOnboarding.collectAsState().value
     val locationDataManager = rememberLocationDataManager()
     val mapViewportState = rememberMapViewportState {
         setCameraOptions {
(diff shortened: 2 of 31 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -92,8 +91,8 @@ fun HomeMapView(
     var selectedStop by remember { mutableStateOf<Stop?>(null) }
 
     val railRouteShapes = getRailRouteShapes()
-    var railRouteLineData: List<RouteLineData>? by rememberSaveable { mutableStateOf(null) }
-    var stopSourceData: FeatureCollection? by rememberSaveable { mutableStateOf(null) }
+    var railRouteLineData: List<RouteLineData>? by remember { mutableStateOf(null) }
+    var stopSourceData: FeatureCollection? by remember { mutableStateOf(null) }
```
KaylaBrady, line 95: Just to prevent crashes while debugging

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -70,46 +69,43 @@ fun NearbyTransitPage(
...
     bottomBar: @Composable () -> Unit
 ) {
     val navController = rememberNavController()
     val currentNavEntry: NavBackStackEntry? by
         navController.currentBackStackEntryFlow.collectAsStateWithLifecycle(initialValue = null)
-    var stopDetailsFilter by rememberSaveable { mutableStateOf<StopDetailsFilter?>(null) }
-    var stopDetailsDepartures by rememberSaveable { mutableStateOf<StopDetailsDepartures?>(null) }
-    var vehiclesData: List<Vehicle> by remember { mutableStateOf(emptyList()) }
+
+    var stopDetailsDepartures by remember { mutableStateOf<StopDetailsDepartures?>(null) }
```
KaylaBrady, line 78: changed to remember for debugging. Should probably move into the new VM.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribedToVehicles.kt
```diff
@@ -0,0 +1,96 @@
...
+}
+
+@Composable
+fun subscribeToVehicles(
+    topic: VehiclesTopic?,
+    vehiclesRepository: IVehiclesRepository = koinInject()
+): List<Vehicle> {
+    Log.i("KB", "new topic ${topic}")
+    val viewModel: VehiclesViewModel =
+        viewModel(factory = VehiclesViewModel.Factory(vehiclesRepository))
```
KaylaBrady, line 81: I initially tried following the same approach as [subscribeToPredictions](https://github.com/mbta/mobile_app/blob/main/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt), but `onCleared` was never invoked (and from what I can tell, also isn't invoked with `subscribeToPredictions`. 

with that, having a single viewModel that persists across renders and dynamically connecting / disconnecting to the correct channel based on changing topic or lifecycle event seemed more reliable.

If this approach is sound, I'll follow-up with making subscribeToPredictions & subscribeToAlerts consistent.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/ContentView.kt
```diff
@@ -56,11 +56,12 @@ fun ContentView(
         rememberBottomSheetScaffoldState(bottomSheetState = rememberStandardBottomSheetState())
     var navBarVisible by remember { mutableStateOf(true) }
 
-    DisposableEffect(null) {
+    LifecycleResumeEffect(null) {
```
JackVCurtis, line 59: 🎉

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribedToVehicles.kt
JackVCurtis: I'm definitely up to adopt the ViewModel-centric approach and refactor the two `subscribe` functions. The only thing I want to suggest is that the ViewModel should abstract away the concept of Channels from the View, e.g. is I'd prefer to avoid lower-level functions like `subscribeToVehicleTopic(topic: String)` in favor of functions like `updateFollowedVehicle(vehicle: Vehicle)`.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToVehicles.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.lifecycle.viewmodel.compose.viewModel
+import com.mbta.tid.mbta_app.model.RouteDirection
+import com.mbta.tid.mbta_app.model.Vehicle
+import com.mbta.tid.mbta_app.model.response.ApiResult
+import com.mbta.tid.mbta_app.model.response.VehiclesStreamDataResponse
+import com.mbta.tid.mbta_app.repositories.IVehiclesRepository
+import org.koin.compose.koinInject
+
+class VehiclesViewModel(
+    private val vehiclesRepository: IVehiclesRepository,
```
KaylaBrady, line 21: One thing worth mentioning here is that I didn't use the TImerViewModel that was used in subscribeToPredictions & subscribeToAlerts to synchronize updates. My perception was it was taking vehicles a long time to appear on the screen - perhaps because waiting for the 1s synchronized update. @JackVCurtis would be curious to learn more about your findings with using the TimerViewModel though - perhaps it should just have a more frequent rate.

EmmaSimon (approved):

androidApp/build.gradle.kts
```diff
@@ -82,6 +82,7 @@ dependencies {
     implementation(libs.mapbox.turf)
     implementation(libs.okhttp)
     implementation(libs.playServices.location)
+    implementation(libs.androidx.lifecycle.runtime.testing)
```
boringcactus, line 85: Should this be an `androidTestImplementation` dependency so it's not available in the `main` source root (and so it's not included in the dependency list)?

PR #584, opened 2024-12-12
Search in Android Nearby Transit

JackVCurtis:
### Summary

_Ticket:_ [🤖 | Search | Search bar in nearby transit](https://app.asana.com/0/1205732265579288/1208833569791046/f)

What is this PR for?

This adds the search bar overlay to Android and allows search by stop with navigation to stop details. UI is left for the next ticket. I've skipped testing this PR in order to unblock other work, but will be adding tests before marking the ticket complete.

iOS
android

### Testing

What testing have you done?


androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/SearchBarOverlay.kt
```diff
@@ -0,0 +1,146 @@
...
+) {
+    var visible =
+        remember(currentNavEntry) {
+            currentNavEntry?.arguments?.getString("stopId")?.isBlank() ?: true
+        }
+    var expanded by rememberSaveable { mutableStateOf(false) }
+    var searchInputState by rememberSaveable { mutableStateOf("") }
+    val searchResults = getSearchResults(searchInputState)
+
+    fun handleSearch(stopId: String) {
```
KaylaBrady, line 52: suggestion(non-blocking): `handleSearch` seems like a misnomer here; this seems to be something along the lines of `handleTapResult`?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
+
+@Composable
+fun getSearchResults(
+    query: String,
+    searchResultRepository: ISearchResultRepository = koinInject()
+): SearchResults? {
+    val viewModel = remember(query) { SearchResultsViewModel(searchResultRepository, query) }
```
KaylaBrady, line 55: question: Why recreate the VM on each query, vs. persist the VM & update the data it uses via a launchedEffect? I found the latter approach helpful in https://github.com/mbta/mobile_app/pull/588, and I think it will help with a missing piece of search functionality:

When I search for a stop, tap that stop, then click the X button to go back to nearby transit, I should see the search bar with my previous query & search results like I do on iOS.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/SearchBarOverlay.kt
```diff
@@ -0,0 +1,146 @@
...
+                            placeholder = { Text(stringResource(R.string.search_by_stop)) },
+                            expanded = expanded,
+                            onQueryChange = { searchInputState = it },
+                            onExpandedChange = { expanded = it },
+                            onSearch = {
+                                expanded = false
+                                if (
+                                    searchResults?.stops != null && searchResults.stops.isNotEmpty()
+                                ) {
+                                    handleSearch(searchResults.stops[0].id)
```
KaylaBrady, line 106: question: selecting the first result by default is unexpected to me - is that an android-specific pattern?

KaylaBrady (approved):
This is a good start! A few comments about structure / functionality. 

The main piece from this PR that is blocking other work is the existence of the `SearchBarOverlay` view. I'm OK with merging this as-is since there is a plan to add tests, and so long as these comments are addressed.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/SearchBarOverlay.kt
KaylaBrady: on the iOS-side, hitting the "done" button on search does nothing (or maybe re-runs the search in the background? but the effect is the same - you still see a list of search results to choose from, rather than automatically being taken to the first result)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
JackVCurtis: The difference in iOS behavior is unrelated, that was me manually clearing the state. I've switched to a persistent VM for consistency with your approach elsewhere, though.

JackVCurtis:
Untouched iOS tests are failing due to network issues on XCode Cloud - merging anyway.

PR #586, opened 2024-12-13
feat: Update stop details navigation to match spec

EmmaSimon:
### Summary

_Ticket:_ [Combined Stop + Trip Details: nav changes](https://app.asana.com/0/1205732265579288/1208725410378683/f)

This updates the combined stop/trip navigation to match the desired behavior. Back buttons are removed from the filtered and unflitered stop pages, and the X button on filtered stop details will bring you back to the place you came from, either the unfiltered stop or nearby transit. Because this doesn't involve changing the actual entry type, but just the stop filter on the entry, there's some slightly convoluted logic to ensure that filter entries are appended rather than replacing the unfiltered entry, and that when the filtered entries are removed, that they don't add duplicate unfiltered entries. This logic also unfortunately needs to exist in two places, since a new entry can be pushed in its entirety to the nearby VM, or the filter value can be modified independently through `lastStopDetailsFilter`.

iOS
android

### Testing

Added tests for new behavior

### DIFF
--- a/iosApp/iosApp/ViewModels/NearbyViewModel.swift
+++ b/iosApp/iosApp/ViewModels/NearbyViewModel.swift
@@ -116,19 +116,40 @@ class NearbyViewModel: ObservableObject {
             )
             pushNavEntry(.stopDetails(stopId: stopId, stopFilter: stopFilter, tripFilter: tripFilter))
         } else if case let .legacyStopDetails(targetStop, _) = entry,
-                  case let .legacyStopDetails(currentStop, _) = navigationStack.last,
-                  targetStop == currentStop {
+                  case let .legacyStopDetails(lastStop, _) = navigationStack.last,
+                  targetStop == lastStop {
             _ = navigationStack.popLast()
             navigationStack.append(entry)
-        } else if case let .stopDetails(stopId: targetStop, stopFilter: _, tripFilter: _) = entry,
-                  case let .stopDetails(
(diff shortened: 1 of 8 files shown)


iosApp/iosApp/ViewModels/NearbyViewModel.swift
```diff
@@ -116,14 +116,40 @@ class NearbyViewModel: ObservableObject {
...
+            ) = navigationStack.last,
+            lastFilter != nil,
+            targetStop == lastStop {
+            // When the stop filter changes, we want a new entry to be added (i.e. no pop) only when
+            // you're on the unfiltered (lastFilter == nil) page, but if there is already a filter,
+            // the entry with the old filter should be popped and replaced with the new value.
             _ = navigationStack.popLast()
+
+            // If the new filter is nil and there is already a nil filter in the stack for the same stop ID,
+            // we don't want a duplicate unfiltered entry, so skip appending a new one
```
boringcactus, line 142: Should this condition really be inside the `lastFilter != nil`? It isn't in the other copy of this logic. (It would probably be worth refactoring this logic out somewhere else even if the result is a bit ugly at each call site.)

EmmaSimon: Good call, how's that? I didn't include the pop check in the function because I didn't want to give the extension function side effects.

boringcactus (approved):

PR #587, opened 2024-12-13
fix(Android): internationalize all hardcoded text

boringcactus:
### Summary

_Ticket:_ [🤖 | Localization catchup](https://app.asana.com/0/1205732265579288/1208936978928073/f)

Since we are using the iOS side as the source of truth after all, this task is easy.

android

### Testing

Verified that the new labels appear properly.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/RouteIcon.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/RouteIcon.kt
@@ -11,8 +11,8 @@ import com.mbta.tid.mbta_app.model.RouteType
 @Composable
 fun routeIcon(routeType: RouteType) =
     when (routeType) {
-        RouteType.BUS -> Pair(painterResource(id = R.drawable.mode_bus), "Bus")
-        RouteType.COMMUTER_RAIL -> Pair(painterResource(id = R.drawable.mode_cr), "Commuter Rail")
-        RouteType.FERRY -> Pair(painterResource(id = R.drawable.mode_ferry), "Ferry")
-        else -> Pair(painterResource(id = R.drawable.mode_subway), "Subway")
+        RouteType.BUS -> Pair(painterResource(id = R.drawable.mode_bus), null)
+        RouteType.COMMUTER_RAIL -> Pair(painterResource(id = R.drawable.mode_cr), null)
+        RouteType.FERRY -> Pair(painterResource(id = R.drawable.mode_ferry), null)
+        else -> Pair(painterResource(id = R.drawable.mode_subway), null)
     }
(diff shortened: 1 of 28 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/RouteIcon.kt
```diff
@@ -11,8 +11,8 @@ import com.mbta.tid.mbta_app.model.RouteType
...
 fun routeIcon(routeType: RouteType) =
     when (routeType) {
-        RouteType.BUS -> Pair(painterResource(id = R.drawable.mode_bus), "Bus")
-        RouteType.COMMUTER_RAIL -> Pair(painterResource(id = R.drawable.mode_cr), "Commuter Rail")
-        RouteType.FERRY -> Pair(painterResource(id = R.drawable.mode_ferry), "Ferry")
-        else -> Pair(painterResource(id = R.drawable.mode_subway), "Subway")
+        RouteType.BUS -> Pair(painterResource(id = R.drawable.mode_bus), null)
+        RouteType.COMMUTER_RAIL -> Pair(painterResource(id = R.drawable.mode_cr), null)
+        RouteType.FERRY -> Pair(painterResource(id = R.drawable.mode_ferry), null)
+        else -> Pair(painterResource(id = R.drawable.mode_subway), null)
```
boringcactus, line 17: These have no screen reader text on iOS, and consequently there is no existing translation for these (although lowercase "bus" and "ferry" are translated as vehicle types for screen reader text), so I'm just removing these here for now. If I read things right, the usages of this are route pills (which should probably specify their own alt text when using these icons for shuttle or ferry routes) and vehicle pucks (which are inherently visual presentations of that information), so this is probably close enough to acceptable that we can fix it at QA time if we need to.

androidApp/src/main/res/values/strings.xml
```diff
@@ -18,21 +27,31 @@
...
     <string name="no_stops_nearby">You’re outside the MBTA service area.</string>
     <string name="no_stops_nearby_title">No nearby stops</string>
+    <string name="northbound">Northbound</string>
     <string name="now">Now</string>
     <string name="other_link_privacy_policy">Privacy Policy</string>
     <string name="other_link_source_code">View source on GitHub</string>
     <string name="other_link_tos">Terms of Use</string>
+    <string name="outbound">Outbound</string>
+    <string name="pin_route">Star route</string>
+    <string name="recenter" tools:ignore="MissingTranslation">Recenter</string>
```
boringcactus, line 44: The map recenter button has no alt text on iOS, and I don't think it's correct to add it in this PR, so I'm punting on these translations for now here too.

BrandonTR (approved):

PR #588, opened 2024-12-16
feat(android): leave / rejoin predictions & alerts after backgrounding

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Stability | Leave realtime channels when backgrounding](https://app.asana.com/0/1205732265579288/1208725410378690/f)

What is this PR for?

Follow-up to https://github.com/mbta/mobile_app/pull/579, applies the same pattern to predictions & alerts. 

iOS
android

### Testing

What testing have you done?
* Ran locally, confirmed I can background & revive the app without crashing
* Updated unit tests

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
@@ -1,87 +1,89 @@
 package com.mbta.tid.mbta_app.android.state
 
+import android.util.Log
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.asFlow
+import androidx.lifecycle.compose.LifecycleResumeEffect
 import androidx.lifecycle.viewmodel.compose.viewModel
-import com.mbta.tid.mbta_app.android.util.TimerViewModel
 import com.mbta.tid.mbta_app.model.response.ApiResult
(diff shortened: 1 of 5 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -98,7 +98,7 @@ fun HomeMapView(
     var railRouteLineData: List<RouteLineData>? by remember { mutableStateOf(null) }
     var stopSourceData: FeatureCollection? by remember { mutableStateOf(null) }
 
-    val now = timer(updateInterval = 10.seconds)
+    val now = timer(updateInterval = 300.seconds)
```
KaylaBrady, line 101: I noticed this was causing lag / jankiness and saw the[ iOS value is 300 seconds](https://github.com/mbta/mobile_app/blob/98c275c218017a428aa41951a3420eed706f25cb/iosApp/iosApp/Pages/Map/HomeMapView.swift#L46), so updated to match.

This can be further improved by moving `getAlertsByStop` to a background thread for calculation since the calculation alone takes ~1 second. That wasn't straight forward to do within the `remember` func though, and would probably need to be moved into the ViewModel. I can make a separate ticket for that

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
```diff
@@ -1,87 +1,89 @@
...
     val predictions: LiveData<PredictionsByStopJoinResponse> = _predictions
     val predictionsFlow = predictions.asFlow().map { it.toPredictionsStreamDataResponse() }
 
-    init {
-        CoroutineScope(Dispatchers.IO).launch {
-            if (stopIds.size > 0) {
-                connectToPredictions()
+    override fun onCleared() {
+        super.onCleared()
+        predictionsRepository.disconnect()
```
EmmaSimon, line 32: nitpick: Should this call the vm's `disconnect()` instead?

EmmaSimon (approved):

KaylaBrady:
android-only change, ran iOS app locally and confirmed unaffected. Merging before iOS workflows complete

PR #592, opened 2024-12-16
feat(Android): add basic onboarding UX

boringcactus:
### Summary

_Ticket:_ [🤖 | Onboarding Flow UX](https://app.asana.com/0/1205732265579288/1208850994085732/f)

Adds the basic onboarding UX to Android.

iOS
android

### Testing

Verified that the fixed onboarding screens appear on first launch and the hide maps onboarding screen appears if TalkBack is turned on. Added unit tests for each individual screen and the overall flow, similar to what exists on iOS.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/ContentView.kt
```diff
@@ -62,6 +66,16 @@ fun ContentView(
         (socket as? PhoenixSocketWrapper)?.attachLogging()
         onDispose { socket.detach() }
     }
+
+    if (!pendingOnboarding.isNullOrEmpty() && !onboardingJustCompleted) {
+        OnboardingPage(
+            pendingOnboarding,
+            onFinish = { onboardingJustCompleted = true },
```
boringcactus, line 73: On iOS, this handler sets the UI state directly
https://github.com/mbta/mobile_app/blob/673e762e7dc4819b7f24a0a9855503713a20de79/iosApp/iosApp/ContentView.swift#L91-L93
but our Android ViewModel-driven state pattern does not really lend itself well to either overwriting the cached state or manually refreshing the data on some external event. This approach is not elegant, but it is direct and fits within the existing architecture.

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -8884,6 +8884,47 @@
         }
       }
     },
+    "When using TalkBack, we can skip reading out maps to keep you focused on transit information." : {
```
boringcactus, line 8887: It does feel a little bit weird to have to add this string to iOS when it only actually exists in Android, but that's a natural consequence of having the source of truth live on one side specifically, and if we didn't add it here there'd be no way to mark it as needs review. (My initial plan had been to just swap in "TalkBack" for "VoiceOver" in the translated strings, but not all the translated strings even mention VoiceOver by name, so I've just gone with the usual approach instead.)

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenViewTest.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.mbta.tid.mbta_app.repositories.Settings
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import org.junit.Rule
+import org.junit.Test
+
+class OnboardingScreenViewTest {
+    @get:Rule val composeTestRule = createComposeRule()
+
+    // We can't easily mock the permission request, so we grant the permission eagerly.
```
KaylaBrady, line 21: 👍 nice

KaylaBrady (approved):
Awesome!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
```diff
@@ -0,0 +1,107 @@
...
+@Composable
+fun OnboardingScreenView(
+    screen: OnboardingScreen,
+    advance: () -> Unit,
+    locationDataManager: LocationDataManager,
+    skipLocationDialogue: Boolean = false,
+    settingsRepository: ISettingsRepository = koinInject()
+) {
+    val coroutineScope = rememberCoroutineScope()
+    var sharingLocation by remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
boringcactus: It's not! If the orientation changes, the permission response is ignored, because this will have been forgotten. Good catch.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -258,6 +258,11 @@ fun HomeMapView(
 
             val locationProvider = remember { PassthroughLocationProvider() }
 
+            MapEffect(true) { map ->
+                map.mapboxMap.addOnMapClickListener { point -> handleStopClick(map, point) }
+                map.location.setLocationProvider(locationProvider)
+            }
+
```
boringcactus, line 265: After completing the onboarding flow, the `LocationDataManager.currentLocation` already has the current location when the `HomeMapView` is being initialized, so if we don't `setLocationProvider` until after we've already `collect`ed, the first location will be completely lost, the puck won't show, and recentering will not work. (This was not fun to track down.)

PR #593, opened 2024-12-17
fix: parcel crash, move data/functions to VM

BrandonTR:
### Summary

_Ticket:_ [Data parsing crash on backgrounding](https://app.asana.com/0/1205732265579288/1208885747081174/f)

What is this PR for?

Resolves the crashes that occur when backgrounding the app.

iOS
android

### Testing

What testing have you done?

No logic changes, just shuffling data around into a view model. Not sure if there's a way to automate lifecycle testing.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
@@ -131,7 +131,7 @@ fun HomeMapView(
     fun updateDisplayedRoutesBasedOnStop() {
         val globalResponse = globalResponse ?: return
         val railRouteShapes = railRouteShapes ?: return
-        if (stopMapData == null) return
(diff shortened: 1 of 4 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -149,7 +129,9 @@ fun HomeMapView(
     }
 
     fun updateDisplayedRoutesBasedOnStop() {
-        if (railRouteShapes == null || globalResponse == null || stopMapData == null) return
+        val globalResponse = globalResponse ?: return
+        val railRouteShapes = railRouteShapes ?: return
+        if (stopMapData == null) return
```
EmmaSimon, line 134: Why not also use a `?: return` for this one?

EmmaSimon (approved):

PR #557, opened 2024-11-26
Android performance/state management refactor

JackVCurtis:
### Summary

_Ticket:_ [Address jankiness / lag in UI](https://app.asana.com/0/1205732265579288/1208725410378688/f)

What is this PR for?

This PR addresses consistent "jank" (lag, stuttering) in the app due to an overloaded UI thread.

- [x] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

What testing have you done?

Manual testing for the expected improvements in performance, unit testing for the code changes related to state management.

PR #596, opened 2024-12-18
fix(subscribeToPredictions): ensure old data present when merging in new data

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Nearby | Predictions unavailable more than expected](https://app.asana.com/0/1205732265579288/1208990382069666/f)

What is this PR for?
Nearby Transit was getting stuck showing "Predictions Unavailable" For routes that had predictions, such as RL. Predictions reliably loaded in Stop Details, but not in Nearby Transit.

After adding debug logs, it seems the issue is that where we [merge new predictions messages with the current state of predictions](https://github.com/mbta/mobile_app/blob/ba896d39ccc1c62013b6544a3b2fab7c6370d513/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt#L60), the current state of predictions is out of date & often null, so we lose predictions from earlier messages. The [docs for LiveData.getValue()](https://developer.android.com/reference/android/arch/lifecycle/LiveData) mention

> Note that calling this method on a background thread does not guarantee that the latest value set will be received.

Whereas [StateFlow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/):

> always has a value which can be safely read at any time.

iOS
android

### Testing

What testing have you done?
* Ran locally, confirmed predictions reliably load in nearby transit


androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
```diff
@@ -3,29 +3,31 @@ package com.mbta.tid.mbta_app.android.state
...
 ) : ViewModel() {
-    private val _predictions: MutableLiveData<PredictionsByStopJoinResponse> = MutableLiveData()
-    val predictions: LiveData<PredictionsByStopJoinResponse> = _predictions
-    val predictionsFlow = predictions.asFlow().map { it.toPredictionsStreamDataResponse() }
+    private val _predictions = MutableStateFlow<PredictionsByStopJoinResponse?>(null)
+    private val _counter = MutableStateFlow<Int>(0)
+
+    val predictions: StateFlow<PredictionsByStopJoinResponse?> = _predictions
+    val predictionsFlow =
+        predictions.map { it?.toPredictionsStreamDataResponse() }.debounce(0.1.seconds)
```
JackVCurtis, line 30: Is the ordering of `map` vs. `debounce` relevant?

JackVCurtis (approved):
👍

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
KaylaBrady: Oh good point, I'll swap to the two to avoid excessive maps

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToAlerts.kt
```diff
@@ -61,7 +63,7 @@ fun subscribeToAlerts(
     val viewModel: AlertsViewModel = viewModel(factory = AlertsViewModel.Factory(alertsRepository))
 
     LifecycleResumeEffect(key1 = null) {
-        viewModel.connect()
+        CoroutineScope(Dispatchers.IO).launch { viewModel.connect() }
```
KaylaBrady, line 66: @JackVCurtis added the use of Dipsatchers.IO in https://github.com/mbta/mobile_app/pull/557, and I mistakenly removed in https://github.com/mbta/mobile_app/pull/588. 

This addresses some of the jankiness / lag while panning the map or scrolling the list of departures.

JackVCurtis (approved):
Looks good!

PR #597, opened 2024-12-18
feat(Android): implement no stops nearby state

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby | Handle outside service area](https://app.asana.com/0/1205732265579288/1208852051039931/f)

The initial predictions loading bug is causing nearby transit to get stuck in loading if the app's first state is outside the service area, but if you load data by moving or panning into the service area, this state will correctly display if you move or pan back out.

iOS
android

### Testing

Added unit tests for several pieces. Manually verified that both buttons work and that the state displays correctly aside from the separate data loading bug.


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewTest.kt
```diff
@@ -258,4 +262,50 @@ class NearbyTransitViewTest : KoinTest {
         composeTestRule.onNodeWithText("Green Line Head Sign").assertIsDisplayed()
         composeTestRule.onNodeWithText("5 min").assertIsDisplayed()
     }
+
+    @OptIn(ExperimentalTestApi::class)
+    @Test
+    fun testNearbyTransitViewNoNearbyStops() {
+        val emptyNearbyKoinApplication = koinApplication {
```
KaylaBrady, line 269: thought(non-blocking): Making a ticket to explore an approach of [reducing the boilerplate of configuring Koin in tests](https://insert-koin.io/docs/reference/koin-android/instrumented-testing/)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -87,12 +89,25 @@ fun NearbyTransitPage(
     val stopDetailsFilter by viewModel.stopDetailsFilter.collectAsState()
     var vehiclesData: List<Vehicle> = subscribeToVehicles(routeDirection = stopDetailsFilter)
 
+    val searchFocusRequester = remember { FocusRequester() }
```
KaylaBrady, line 92: question (non-blocking): Why not rememberSaveable? My understanding is that we'd always want to default to using the saveable version

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
```diff
@@ -37,8 +37,7 @@ class PredictionsViewModel(
     }
 
     fun connect(stopIds: List<String>?) {
-
-        if (!stopIds.isNullOrEmpty()) {
+        if (stopIds != null) {
```
KaylaBrady, line 40: 👍 thanks for catching

KaylaBrady (approved):
Nice!!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
boringcactus: I do not think the `FocusRequester` supports `Parcel`, and even if it did, it doesn't contain any data that we need to preserve while the app is in the background, the only important thing is that we use the same instance in `openSearch` and `SearchBarOverlay`, and if it's a different instance after the app comes back from background then that's entirely fine.

PR #598, opened 2024-12-18
fix(Android): parse join responses correctly

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby Transit | Initial load jank](https://app.asana.com/0/1205732265579288/1209007469708327/f)

I appear to have broken this all the way back in #263.

iOS
android

### Testing

Manually verified that alerts load when the alerts join response is sent now (and that predictions load when the predictions join response is sent now).


JackVCurtis (approved):
👍

PR #600, opened 2024-12-18
fix(Android): minimize issues when updating map global data

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby Transit | Initial load jank](https://app.asana.com/0/1205732265579288/1209007469708327/f)

The acceptance criteria mention this too and I entirely forgot about it when implementing #598 so here it is as a separate PR. More details in the Asana comments.

iOS
android

### Testing

Verified that the alerts are correctly drawn as they arrive and that the UI does not hang while the global data is recalculating.


BrandonTR (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapViewModel.kt
```diff
@@ -94,12 +102,20 @@ open class MapViewModel(
...
-                GlobalMapData(it, GlobalMapData.getAlertsByStop(it, alertsData, now))
-            }
+    override suspend fun globalMapData(now: Instant): GlobalMapData? =
+        globalResponse.first()?.let {
+            GlobalMapData(it, GlobalMapData.getAlertsByStop(it, alertsData, now))
+        }
+
+    @Composable
+    override fun rememberGlobalMapData(now: Instant): GlobalMapData? {
+        var result by remember { mutableStateOf<GlobalMapData?>(null) }
```
KaylaBrady, line 112: suggestion(non-blocking): For consistency with the `refresh` functions below, this could update a mutable state property of the VM rather than remembering the result itself.

boringcactus: I don't think it works the same way as those `refresh` functions - it's a `@Composable` called from the UI thread and not a `suspend fun` called from a `LaunchedEffect`, and even if I made `globalMapData()` update some mutable state it'd still need to return the value to avoid uglier code or less up-to-date output of the `refresh` functions.

PR #601, opened 2024-12-19
feat(iOS): Combined stop details - display scheduled trips

EmmaSimon:
### Summary

_Ticket:_ [Combined Stop + Trip Details: Show trip details when there's no predicted vehicle](https://app.asana.com/0/1205425564113216/1208790341716800/f)

Stacked onto #599 

Display trip schedules if no vehicle or predictions exist. For any non-ferry trips, also allow tapping on the trip card to open an explainer cover that shows some text explaining the state.

Simulator Screenshot - iPhone 15 - 2024-12-18 at 20 49 22 Simulator Screenshot - iPhone 15 - 2024-12-18 at 20 49 44 Simulator Screenshot - iPhone 15 - 2024-12-18 at 20 51 44

iOS

android

### Testing

Updated existing tests, added new tests for new behavior.


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -149,7 +149,7 @@ struct StopDetailsFilteredDepartureDetails: View {
         ForEach(Array(statuses.enumerated()), id: \.offset) { index, row in
             VStack(spacing: 0) {
                 OptionalNavigationLink(
-                    value: row.navigationTarget,
+                    value: nil,
```
boringcactus, line 152: An `OptionalNavigationLink(value: nil)` is just its label, so if this never needs to be a link anymore then it should probably not be an `OptionalNavigationLink`.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LoadingPlaceholders.kt
```diff
@@ -123,7 +123,7 @@ object LoadingPlaceholders {
                     val stop = objects.stop { name = "Loading" }
                     val prediction =
                         objects.prediction {
-                            this.trip = trip
+                            this.trip = objects.trip(routePattern)
```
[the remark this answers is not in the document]
EmmaSimon: Oops, yep, this was a misguided attempt at getting rid of duplicate trip IDs in the loading data that I forgot to revert, good catch.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
EmmaSimon: The UI to replace the status rows isn't implemented yet, this is just a placeholder so I wasn't paying much attention to it, just made it a little prettier and got rid of the nav link.

boringcactus (approved):

PR #605, opened 2024-12-20
feat(Android): animate new page sliding up

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby | Sheet / navigation](https://app.asana.com/0/1205732265579288/1208852051039919/f)

Apparently the Android stock navigation subsystem is just this flexible.

I haven't been able to see this animation play back smoothly at its default speed because we drop frames pretty badly on the page transition, but if I slow it way down then I do briefly see smooth motion. I don't think that frame loss is in-scope for this PR, but I may look into it separately.

iOS
android

### Testing

Manually validated that the animation should theoretically play.


JackVCurtis (approved):
Cool!

PR #612, opened 2024-12-24
feat(Android): show live data indicator on predictions

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby | Realtime indicator](https://app.asana.com/0/1205732265579288/1208990382069675/f)

I tried making this a custom modifier like it is on iOS, but Android modifiers can't (as far as I was able to tell) just wrap the entire component with another Composable.

iOS
android

### Testing

Manually verified (since testing the presence of semantically-invisible UI elements in Android is a bit difficult) that the icons show correctly and look the same way they do on iOS.


JackVCurtis (approved):
👍

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
```diff
@@ -60,57 +60,65 @@ fun formatTime(time: Instant): String =
     format.format(time.toLocalDateTime(TimeZone.currentSystemDefault()).toJavaLocalDateTime())
 
 @Composable
-fun UpcomingTripView(state: UpcomingTripViewState) {
+fun UpcomingTripView(state: UpcomingTripViewState, hideRealtimeIndicators: Boolean = false) {
```
KaylaBrady, line 63: suggestion: maybe worth a test that the indicator image is included when expected

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/WithRealtimeIndicator.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    Row(
+        modifier,
+        horizontalArrangement = Arrangement.spacedBy(subjectSpacing),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        if (!hideIndicator) {
+            Image(
+                painterResource(R.drawable.live_data),
+                contentDescription = null,
```
KaylaBrady, line 34: question(non-blocking): should there be a content description here, like "live"? This seems like an important symbol to convey to Talk Back users, though acknowledge that this is matching the existing iOS behavior

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
boringcactus: Since the icon has no content description, I'm not aware of a way for us to test that it exists without using test tags, and I don't want to either add a test tag to the image unconditionally or add an extra flag to add the test tag that's supposed to be only ever true in testing.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/WithRealtimeIndicator.kt
boringcactus: There shouldn't, because we override the accessibility label for the entire nearby transit row at once, and on iOS we [annotate](https://github.com/mbta/mobile_app/blob/fea79c200cead979b3494ee9062313d48ad2a57d/iosApp/iosApp/ComponentViews/UpcomingTripAccessibilityFormatters.swift#L64) scheduled times with "scheduled" instead of annotating live times with "live".

KaylaBrady: I think those annotations predate the decision to add the "live" indicator rather than the scheduled indicator. We should probably go back and update those, I can make a ticket for it.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
KaylaBrady: I don't mind the use of test tags as a last resort when there are no other good ways to test.

KaylaBrady (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/WithRealtimeIndicator.kt
KaylaBrady: [Ticket added](https://app.asana.com/0/1208621177839510/1209072793746972/f)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
boringcactus: Added some checks for the realtime indicators to the `UpcomingTripViewTest`.

PR #616, opened 2024-12-31
feat(accessibility): Android Nearby Transit Talk Back optimizations

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Nearby | Screen reader text](https://app.asana.com/0/1205732265579288/1208629983782645/f)

What is this PR for?

This PR adds a few key improvements to TalkBack behavor
* announce headings
* update pin / unpin button hints
* nicer descriptions of upcoming arrivals

A few pieces are out of scope and will be addressed in subsequent PRs
* Incorporating vehicle type into the upcoming trip announcement
    * this was done on iOS at some point but is no longer working. May break out into a separate ticket entirely to resolve both at once
* Announcing GL arrivals more nicely ("B train arriving in 5 minutes" instead of "arriving in 5 minutes B train"). Also not done on iOS, will make separate ticket.
iOS
android

### Testing

What testing have you done?


iosApp/iosApp/Localizable.xcstrings
```diff
@@ -595,96 +595,96 @@
         }
       }
     },
-    "%@ arriving in %@ min" : {
+    "%@ arriving in %d min" : {
```
KaylaBrady, line 598: using the strings as-is on android gave a warning about passing an int when the formatter expected a string.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/TransitHeader.kt
```diff
@@ -44,6 +46,7 @@ fun TransitHeader(
             Text(
                 text = name,
                 maxLines = 1,
+                modifier = Modifier.semantics { heading() },
```
JackVCurtis, line 49: Is there an easy rule for when we'll need to add custom semantics?

JackVCurtis (approved):
Looks great, happy to see the test coverage!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/TransitHeader.kt
KaylaBrady: Not quite an easy rule, but when the default behavior doesn't match the designs or impedes the user's ability to find and understand the information. Any visual cues that a sighted user gets from using the app should be represented via semantics (such as heading level here). When we are using built-in material components, we shouldn't have to worry about custom semantics unless we want to override the defaults to be more descriptive for that particular element in its context.

For the most part this PR is directly duplicating the customizations done on iOS, though there are some differences (iOS supports heading levels, whereas heading is binary on android).

PR #617, opened 2024-12-31
feat(accessibility): Talk back descriptions support pluralized vehicle types

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Nearby | Screen reader text](https://app.asana.com/0/1205732265579288/1208629983782645/f)

What is this PR for?
Adds pluralized vehicle type in route pill & upcoming trip content descriptions. 
**Note:** The UpcomingTrip is not currently passed the vehicle type on android or iOS - this is a regression from https://github.com/mbta/mobile_app/pull/204 that should be fixed in this new[ bug ticket](https://app.asana.com/0/1208621177839510/1209064576129923/f).

While adding tests I noticed that the android test was missing a case for cancelled trips, and in fact the iOS cancelled trip behavior was broken! Includes the small fix for that as well.

iOS

android

### Testing

What testing have you done?

Added unit tests & ran locally with talk back on.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/ObjectCollectionBuilder.kt
```diff
@@ -163,13 +163,13 @@ class ObjectCollectionBuilder {
     class RouteBuilder : ObjectBuilder<Route> {
         var id = uuid()
         var type = RouteType.LIGHT_RAIL
-        var color = ""
+        var color = "FFFFFF"
```
KaylaBrady, line 166: Got errors when writing RoutePill tests that the color code needed to be at least 6 characters - setting that by default on all test objects.

EmmaSimon (approved):

PR #620, opened 2025-01-03
feat(iOS): Combined stop details - Add no prediction states

EmmaSimon:
### Summary

_Ticket:_ [Combined Stop + Trip Details: Handle predictions unavailable / service ended / no service today](https://app.asana.com/0/1205732265579288/1208725410378680/f)

Adds the UI for states where there are no predictions for the selected route. This includes the basic "predictions unavailable" state, along with "service ended" and "no service today".

Simulator Screenshot - iPhone 15 - 2025-01-03 at 17 30 35 Simulator Screenshot - iPhone 15 - 2025-01-03 at 17 31 43 Simulator Screenshot - iPhone 15 - 2025-01-03 at 17 33 59

This also includes a fix for an `onChange` race condition where a selected vehicle could stay selected after the stop filter was changed, this was resolved by moving the stop ID and filters into a single struct and only running the `onChange` with that.

iOS

### Testing

Added kotlin tests for new shared code, SwiftUI tests for updated views, and manual testing to verify behavior.


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -60,31 +70,54 @@ struct StopDetailsFilteredDepartureDetails: View {
                         .padding(.bottom, 6)
                         .dynamicTypeSize(...DynamicTypeSize.accessibility1)
 
-                        departureTiles(patternsByStop, view)
-                            .dynamicTypeSize(...DynamicTypeSize.accessibility3)
-                            .onAppear { if let id = tripFilter?.tripId { view.scrollTo(id) } }
+                        if !tiles.isEmpty {
```
KaylaBrady, line 73: question(non-blocking):`tiles` and `noPredictionsStatus` mutually exclusive, right? If so, It might be worth making a single enum field to convey that, like `tilesOrStatus`

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
```diff
@@ -181,6 +181,39 @@ data class StopDetailsDepartures(val routes: List<PatternsByStop>) {
...
 
+        fun getNoPredictionsStatus(
+            realtimePatterns: List<RealtimePatterns>,
+            now: Instant
+        ): RealtimePatterns.Format? {
+            val patternStatuses =
+                realtimePatterns.mapNotNull { pattern ->
+                    when (pattern) {
+                        is ByHeadsign -> getStatusFormat(pattern, now)
+                        else -> null
```
KaylaBrady, line 192: question(non-blocking): Why is ByDirection always null?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
```diff
@@ -181,6 +181,39 @@ data class StopDetailsDepartures(val routes: List<PatternsByStop>) {
             return routes?.let { StopDetailsDepartures(it) }
         }
 
+        fun getNoPredictionsStatus(
+            realtimePatterns: List<RealtimePatterns>,
+            now: Instant
+        ): RealtimePatterns.Format? {
```
KaylaBrady, line 187: suggestion(non-blocking): The return type for this seems a bit broad, since `RealtimePatterns.Format.Some` would never be returned. It might be nice to have a new sealed class that groups together the no predictions status types, but that might be a bigger refactor / better left for a separate change.

iosApp/iosApp/Pages/StopDetails/StopDetailsNoTripCard.swift
```diff
@@ -0,0 +1,83 @@
...
+                Explanation under the 'Predictions unavailable' header in stop details.
+                The interpolated value can be "buses" or "trains".
+                """
+            ), routeType.typeText(isOnly: false))
+        default: nil
+        }
+    }
+
+    @ViewBuilder
+    var headerText: some View {
```
KaylaBrady, line 50: 👍 I don't mind this minor duplication!

iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -10,10 +10,14 @@ import shared
 import SwiftPhoenixClient
 import SwiftUI
 
-struct StopDetailsPage: View {
+struct StopDetailsPageFilters: Equatable {
```
KaylaBrady, line 13: praise: I like the way you broke out this struct

iosApp/iosAppTests/Pages/StopDetails/StopDetailsFilteredDepartureDetailsTests.swift
```diff
@@ -222,17 +225,48 @@ final class StopDetailsFilteredDepartureDetailsTests: XCTestCase {
...
+            errorBannerVM: .init(),
+            nearbyVM: .init(),
+            mapVM: .init(),
+            stopDetailsVM: .init(),
+            viewportProvider: .init()
+        ).environmentObject(ViewportProvider())
+
+        XCTAssertThrowsError(try sut.inspect().find(TripDetailsView.self))
+        XCTAssertThrowsError(try sut.inspect().find(DepartureTile.self))
+        XCTAssertNotNil(try sut.inspect().find(StopDetailsNoTripCard.self))
```
KaylaBrady, line 270: 👍 nice testing!

KaylaBrady (approved):
Nice work! A few minor comments

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
EmmaSimon: Yeah, I started with an approach more along those lines, but it seemed unnecessarily redundant since we already had the Format class with the same values we needed. It won't be used in this code, but we'll also need to handle the cancelled prediction state in a similar way in the NoTripCard, so it seems less awkward to just have everything use the `Format` directly, rather than translate between that and a redundant class, but idk, I could be convinced either way!

KaylaBrady: Oh interesting, I was seeing NoTripCard a little differently. To me the cancelled prediction state is something different from the NoTripCard (which is more like a NoTrip**s**Card, related to [my comment about](https://github.com/mbta/mobile_app/pull/620#discussion_r1904217655) status & tiles as mutually exclusive). 

The way I was thinking of the hierarchy, you either have a status or tiles, and if you click a tile, you'd either have a CancelledTripCard or a TripDetailsView.

Along those lines, I was thinking there could be a new sealed class like:
```
    sealed class NoTripsFormat {
        abstract val secondaryAlert: SecondaryAlert?

        data class NoSchedulesToday(override val secondaryAlert: SecondaryAlert?): NoTripsFormat()
        data class ServiceEndedToday(override val secondaryAlert: SecondaryAlert?): NoTripsFormat()
        data class None(override val secondaryAlert: SecondaryAlert?) : NoTripsFormat()
    }
```

And the existing `Format` class could be updated to use that sealed class 
```
sealed class Format {
        ...
        data class NoTripsFormat(val noTripsFormat: RealtimePatterns.NoTripsFormat) : Format() {
            override val secondaryAlert: SecondaryAlert? = noTripsFormat.secondaryAlert
        }
        ...
       
}
```

I think that refactor would make the types clearer without having to introduce a new / redundant enum, but it would require some refactoring that is probably out of scope for this PR. 

But interested in your thoughts on distinguishing between a `NoTripsCard` and `CancelledTripCard`!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
EmmaSimon: I was copy pasting the existing logic so I didn't think about it very hard and assumed that it was excluded because it would always be `ByHeadsign` (at some point in the process we assign all predictions into both the relevant `ByHeadsign` and `ByDirection` in a map, I figured that here had only filtered to headsign), but that's not actually true and it seems to be broken at places where it can be grouped by direction. Lechmere and the next couple stops westbound are probably the only place this is likely to happen under normal service, but I did see it get into a forever loading state there because of this, thanks for calling it out!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
EmmaSimon: I don't see a great reason to use a totally separate component for the cancelled trip card since the styling and placement is exactly the same, though it would need to be a separate case in the if statement that we're currently using to determine whether or not to display the no trip card or the trip details. We'll also need the first trip of the day state eventually, which would just be another `NoTripsFormat` if we did this refactor. I think it might be worth doing and does clarify things a bit, though it will require nested switch statements in a few places, which might be mildly awkward (though obviously the `NoTripsFormat` switch would need to be split out into a separate function at least).

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
EmmaSimon: This should be fixed now but it's hard to get a real world case to see it working. I'll try to keep Lechmere up on my screen while I work on other stuff in case it happens.

KaylaBrady: Awesome, appreciate you adding a test for it!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
KaylaBrady: Yeah fair point on not needing to be a totally separate component - maybe there is a more basic shared component with the styling & placement they could share, even if the text / images are passed in separately. 

In any case, that can be considered further as part of the ticket for handling cancelled trips!

KaylaBrady (approved):

PR #622, opened 2025-01-06
Jvc android search recent stops

JackVCurtis:
### Summary

_Ticket:_ [🤖 | Search | Search recently visited](https://app.asana.com/0/1205732265579288/1208839363050256/f)

What is this PR for?

When there is no query entered in the Search view, shows a list of recently visited stops.

Screenshot_20250106_094147

iOS

android

### Testing

What testing have you done?
Updated unit tests for affected views.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
@@ -36,8 +36,8 @@ class SearchResultsViewModel(
         job?.cancel()
         job =
             CoroutineScope(Dispatchers.IO).launch {
-                delay(500)
                 if (query.isNotEmpty()) {
+                    delay(500)
                     when (val data = searchResultRepository.getSearchResults(query)) {
                         is ApiResult.Ok -> _searchResults.emit(data.data)
                         is ApiResult.Error -> _searchResults.emit(null)
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
@@ -4,12 +4,19 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.lifecycle.ViewModel
+import com.mbta.tid.mbta_app.history.Visit
+import com.mbta.tid.mbta_app.model.LocationType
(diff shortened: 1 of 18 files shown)


androidApp/src/main/res/values/strings.xml
```diff
@@ -64,6 +64,7 @@
     <string name="other_link_tos">Terms of Use</string>
     <string name="outbound">Outbound</string>
     <string name="recenter" tools:ignore="MissingTranslation">Recenter</string>
+    <string name="recently_visited" tools:ignore="MissingTranslation">Recently Visited</string>
```
boringcactus, line 67: This is translated in iOS, so those translations should be automatically converting to Android. Is that not working?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
```diff
@@ -4,56 +4,89 @@ import androidx.compose.runtime.Composable
...
-                    }
-                }
-        } else if (lastClickTime == null) {
-            job =
-                CoroutineScope(Dispatchers.IO).launch {
+        job?.cancel()
+        job =
+            CoroutineScope(Dispatchers.IO).launch {
+                delay(500)
+                if (query.isNotEmpty()) {
```
boringcactus, line 40: I'm a little bit nervous about this 500ms delay on initial load - we want to show the history as soon as it's requested - or when the query becomes empty - if the idea is to avoid spamming the backend with queries that are about to be invalidated, we don't need to worry about that if nothing will be sent to the backend anyway. A brute-force solution would be to move the delay inside the `if` so that there's no delay when the query is empty.

It looks like these issues also exist on iOS to some degree, because of the way Publisher.debounce works (which matches the way Flow.debounce works in Kotlin), so we could keep this potential minor UI papercut in the name of platform consistency if we so desired.

androidApp/src/main/res/values/strings.xml
JackVCurtis: It was not within my IDE, but it's possible I need to run something.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getSearchResults.kt
JackVCurtis: I went with this as the simplest debounce implementation I could find. I didn't notice an issue in practice since the job will be fired on render and it takes at least half a second to expand the search, but I like moving the delay inside the `if` statement for logical clarity if nothing else.

boringcactus (approved):

PR #627, opened 2025-01-07
feat(iOS): Stop details cancelled trip card

EmmaSimon:
### Summary

_Ticket:_ [Combined Stop + Trip Details: Handle cancelled trip](https://app.asana.com/0/1205732265579288/1209010745257743/f)

Makes it possible to select cancelled trips on the stop page, and when you do, shows an explainer card saying that the trip is cancelled. Also includes a change to the auto trip filter logic to skip any cancelled trips and automatically select the first predicted or scheduled trip when you open stop details, so that cancelled trips must be manually tapped to be selected (unless there are only cancelled trips, in which case the first is selected).

Simulator Screenshot - iPhone 15 - 2025-01-07 at 16 26 24

iOS

android

### Testing

Manually tested with cancelled trips, added unit tests for auto trip filter changes and for UI changes.

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
+++ b/iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
@@ -49,6 +49,18 @@ struct StopDetailsFilteredDepartureDetails: View {
         return if routeType == .bus, !isSL { Color.text } else { routeColor }
     }
 
+    var selectedTripIsCancelled: Bool {
+        if let tripFilter {
+            patternsByStop.patterns.contains { pattern in
+                pattern.upcomingTrips.contains { trip in
+                    trip.trip.id == tripFilter.tripId && trip.isCancelled
+                }
+            }
+        } else {
+            false
+        }
+    }
+
     var body: some View {
         ZStack(alignment: .top) {
             routeColor.ignoresSafeArea(.all)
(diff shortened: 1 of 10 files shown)


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -43,10 +43,17 @@ struct StopDetailsFilteredDepartureDetails: View {
 
     var routeColor: Color { Color(hex: patternsByStop.representativeRoute.color) }
     var routeType: RouteType { patternsByStop.representativeRoute.type }
-    var headerColor: Color {
-        // Regular bus color needs to be overridden for contrast, but SL should not be
```
boringcactus, line 47: This logic looks like it's been removed; is that no longer an issue with the new component?

boringcactus (approved):

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
EmmaSimon: I brought up that there were also contrast issues in dark mode for other route colors (mainly blue and CR), so Adam updated [the styling](https://www.figma.com/design/eFiLM8xjo9hKprg93BBCtZ/App-UI-Design?node-id=10794-608870&t=7BXsFnmPwXnzzY4L-4) to always use the route color for the icon and divider line, regardless of contrast, and just use the standard text color for the text.

PR #630, opened 2025-01-08
feat(Android): error banner

BrandonTR:
### Summary

_Ticket:_ [🤖 | Nearby | Display error banners](https://app.asana.com/0/1205732265579288/1208968297244757/f)

What is this PR for?

Adding the error banner on Android for nearby transit.

Screenshot_20250108_081216

iOS

android

### Testing

Added previews and automated tests for various error banner states.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
@@ -53,11 +53,7 @@ fun ErrorBanner(vm: ErrorBannerViewModel) {
             ErrorCard(
                 details = {
                     Row(verticalAlignment = Alignment.CenterVertically) {
-                        Icon(
-                            Icons.Rounded.Clear,
-                            contentDescription =
-                                "Displayed when the phone is not connected to the network"
-                        )
+                        Icon(Icons.Rounded.Clear, contentDescription = "")
                         Text(
                             stringResource(R.string.unable_to_connect),
                             modifier = Modifier.padding(start = 8.dp)
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
@@ -1,0 +1,168 @@
+package com.mbta.tid.mbta_app.android.component
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.rounded.Clear
+import androidx.compose.material.icons.rounded.Refresh
+import androidx.compose.material3.Button
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.mbta.tid.mbta_app.android.R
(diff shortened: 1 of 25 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
```diff
@@ -0,0 +1,168 @@
...
+            )
+        }
+        is ErrorBannerState.NetworkError -> {
+            ErrorCard(
+                details = {
+                    Row(verticalAlignment = Alignment.CenterVertically) {
+                        Icon(
+                            Icons.Rounded.Clear,
+                            contentDescription =
+                                "Displayed when the phone is not connected to the network"
```
KaylaBrady, line 59: suggestion: Add the contentDescription to strings.xml so that it is localized. It looks like this is pulling from the comment field of the iOS implementation, but that is a commenter for translators to have context, whereas contentDescription is what should be read aloud to users using TalkBack. 

It looks like there is no description set for this image on the iOS side. I think it would be OK to omit here since it is a decorative element redundant to the "Unable to connect" text, but you can also check with design.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/fetchApi.kt
```diff
@@ -0,0 +1,33 @@
...
+    errorKey: String,
+    getData: suspend () -> ApiResult<T>,
+    onSuccess: suspend (T) -> Unit = {},
+    onRefreshAfterError: () -> Unit
+) {
+    val result: ApiResult<T> =
+        try {
+            getData()
+        } catch (e: Exception) {
+            ApiResult.Error(code = null, message = e.message ?: "No error message")
```
KaylaBrady, line 22: suggestion: I _think_ "No error message" would only be exposed in dev debug mode, but can leave it empty string to prevent unlocalized string from surfacing in UI

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getNearby.kt
```diff
@@ -40,10 +45,15 @@ class NearbyViewModel(
...
 
-    suspend fun getNearby(globalResponse: GlobalResponse, location: Position) {
-        when (val data = nearbyRepository.getNearby(globalResponse, location)) {
-            is ApiResult.Ok -> _nearbyResponse.emit(data.data)
-            is ApiResult.Error -> TODO("handle errors")
+    fun getNearby(globalResponse: GlobalResponse, location: Position) {
+        CoroutineScope(Dispatchers.IO).launch {
+            fetchApi(
+                errorBannerRepo = errorBannerRepository,
+                errorKey = "NearbyViewModel.getNearby",
```
KaylaBrady, line 52: issue: This looks great for getNearby, but I would expect to see the other static data fetches for the nearby page to be now fetched with `fetchApi` as well, like `getGlobalData` and `getSchedules`. Will those be added in a follow-up PR?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
```diff
@@ -0,0 +1,168 @@
...
+import androidx.compose.ui.unit.dp
+import com.mbta.tid.mbta_app.android.R
+import com.mbta.tid.mbta_app.model.ErrorBannerState
+import com.mbta.tid.mbta_app.repositories.MockErrorBannerStateRepository
+import com.mbta.tid.mbta_app.repositories.MockSettingsRepository
+import kotlin.time.Duration.Companion.minutes
+import kotlinx.datetime.Clock
+
+@Composable
+fun ErrorBanner(vm: ErrorBannerViewModel) {
```
KaylaBrady, line 38: suggestion (non-blocking): This looks good! Maybe something to consider for future PRs as a way to break down into smaller PRs - perhaps this could have been 2 PRs, one with the logic changes & a ErrorBanner w/ just a generic placeholder, then a second that added the different UX for each error banner state.

KaylaBrady (approved):
Nice! Left a few comments that I'd like to see addressed as a follow-up PR, but not blocking for merging this PR since they are minimal impact & merging will unblock other tasks

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/getNearby.kt
BrandonTR: Since I needed to resolve some conflicts anyway I went ahead and addressed this and other comments, thanks for flagging!

androidApp/src/main/res/values/strings.xml
```diff
@@ -98,5 +98,10 @@
     <string name="westbound">Westbound</string>
     <string name="stop_filter_pills_filter_hint">Apply a filter so that only arrivals from this route are displayed</string>
     <string name="stop_filter_pills_remove_hint">Remove the filter and display all routes</string>
-
+    <string name="refresh">Refresh</string>
+    <string name="refresh_predictions">Refresh predictions</string>
+    <string name="updated_mins_ago" tools:ignore="MissingTranslation">Updated %1$d minutes ago</string>
```
KaylaBrady, line 103: question: Why is this missing translation? I think there should be an analogous string in use on iOS

PR #631, opened 2025-01-08
feat(android.NearbyTansitView): Shimmer loading

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Nearby | Loading states](https://app.asana.com/0/1205732265579288/1208629983782633/f)

What is this PR for?
Adds shimmer loading to the nearby transit view only. 

If this approach looks good, will take the same approach for the unfiltered stop details view. Also TODO: make sure accessibility of loading component is correct on iOS side.

Notes:
* Unlike in SwiftUI, there is no built-in redacted functionality, and applying a modifier to a container doesn't apply it to all children. This requires getting the state of whether or not to display the placeholder down into each view that needs a placeholder. `CompositionLocal` seemed like the cleanest option for that. 

Alternatives considered:
* Using unicode shaded box character, which produced gaps between each box in certain views
* Using the redacted font, which didn't work cleanly where font style tokens were used, required a different solution for images, and didn't work automatically on localized text if set to Chinese
* Recomposing views to have simple layout-only components and passed all views (placeholder or regular) as props. Significant refactor with little benefit

iOS

android
- [X] All user-facing strings added to strings resource
- [X] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible

### Testing

What testing have you done?
* Updated unit tests
* Ran locally

https://github.com/user-attachments/assets/668d4532-81e1-48c2-af82-65cb937e9355


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/LoadingRouteCard.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.datetime.Clock
+
+@Composable
+fun LoadingRouteCard() {
+    val placeholderRouteData = LoadingPlaceholders.nearbyRoute()
+
+    val contentDesc = stringResource(R.string.loading)
+
+    Column(
+        modifier = Modifier.shimmer().clearAndSetSemantics { contentDescription = contentDesc }
```
EmmaSimon, line 22: non-blocking: Can be done in a follow up PR, but can we create our own custom modifier for this which includes shimmer, clearing semantics, and [turning off interactions](https://stackoverflow.com/questions/77960798/how-to-disable-interactions-of-all-components-at-once-in-jetpack-compose)?

androidApp/build.gradle.kts
```diff
@@ -71,6 +71,9 @@ dependencies {
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.navigation.compose)
     implementation(libs.compose.material3)
+    implementation(libs.compose.placeholder.material3)
+    // TODO toml
+    implementation("com.valentinilk.shimmer:compose-shimmer:1.3.1")
```
EmmaSimon, line 76: Was this only not added because we weren't sure if we would end up using the library?

EmmaSimon (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/LoadingRouteCard.kt
KaylaBrady: Totally, good idea! My understanding is that [clearAndSetSemantics](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.Modifier).clearAndSetSemantics(kotlin.Function1)) does remove all tap interactions within the contained child views - tapping pieces of the loading card does nothing as-is. Is there something I'm missing that we'd get with the box approach?

androidApp/build.gradle.kts
KaylaBrady: Yeah I should have just taken the extra 2 seconds to add it the right way in the first place 😅 
[648a301](https://github.com/mbta/mobile_app/pull/631/commits/648a3011c0ce25e7a2d968da552b555fecbbaa3c)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/LoadingRouteCard.kt
EmmaSimon: Ah, okay, I didn't realize `clearAndSetSemantics` already did that, I wasn't worried about the interactivity of this component, but future ones that we'll want to apply a shared modifier to.

PR #635, opened 2025-01-09
fix(Android): reset nearby transit state on pan/recenter

boringcactus:
### Summary

_Ticket:_ [🤖 | Nearby | Reset state when panning or recentering](https://app.asana.com/0/1205732265579288/1208915825996354/f)

I thought this would be easy, and if I knew one more thing about ViewModels it would've been easy, but I did not, so this was really difficult.

iOS

android

### Testing

Manually verified that the ViewModel gets reset and the stale data does not flash in for a frame or two.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
@@ -46,6 +46,7 @@ import com.mbta.tid.mbta_app.android.map.IMapViewModel
 import com.mbta.tid.mbta_app.android.map.MapViewModel
 import com.mbta.tid.mbta_app.android.nearbyTransit.NearbyTransitTabViewModel
 import com.mbta.tid.mbta_app.android.nearbyTransit.NearbyTransitView
+import com.mbta.tid.mbta_app.android.nearbyTransit.NearbyTransitViewModel
 import com.mbta.tid.mbta_app.android.nearbyTransit.NoNearbyStopsView
 import com.mbta.tid.mbta_app.android.search.SearchBarOverlay
 import com.mbta.tid.mbta_app.android.state.subscribeToVehicles
@@ -64,6 +65,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.launch
+import org.koin.androidx.compose.koinViewModel
 import org.koin.compose.koinInject
 
 @OptIn(ExperimentalMaterial3Api::class)
@@ -216,6 +218,8 @@ fun NearbyTransitPage(
                 }
             }
             composable<SheetRoutes.NearbyTransit> {
+                // for ViewModel reasons, must be within the `composable` to be the same instance
+                val nearbyViewModel: NearbyTransitViewModel = koinViewModel()
                 LaunchedEffect(true) {
                     if (!navBarVisible) {
                         showNavBar()
@@ -234,13 +238,13 @@ fun NearbyTransitPage(
                 }
                 LaunchedEffect(nearbyTransit.viewportProvider.isManuallyCentering) {
                     if (nearbyTransit.viewportProvider.isManuallyCentering) {
-                        // TODO reset view model
+                        nearbyViewModel.reset()
                         targetLocation = null
                     }
                 }
                 LaunchedEffect(nearbyTransit.viewportProvider.isFollowingPuck) {
                     if (nearbyTransit.viewportProvider.isFollowingPuck) {
-                        // TODO reset view model
+                        nearbyViewModel.reset()
                         targetLocation = null
                     }
                 }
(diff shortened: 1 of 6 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -216,6 +218,8 @@ fun NearbyTransitPage(
                 }
             }
             composable<SheetRoutes.NearbyTransit> {
+                // for ViewModel reasons, must be within the `composable` to be the same instance
+                val nearbyViewModel: NearbyTransitViewModel = koinViewModel()
```
boringcactus, line 222: Originally, I had `nearbyViewModel: NearbyTransitViewModel = koinViewModel()` as an argument to `NearbyTransitPage`, but since navigation states [can own ViewModels](https://developer.android.com/topic/libraries/architecture/viewmodel#lifecycle), calling `koinViewModel()` outside this `composable` creates a different instance than the one that's created as an argument in `NearbyTransitView`, but calling it here gives the same instance.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/phoenix/PhoenixSocketWrapper.kt
```diff
@@ -20,7 +20,9 @@ value class PhoenixSocketWrapper(private val socket: Socket) : PhoenixSocket {
 
     fun attachLogging() {
         socket.onMessage { message -> Log.i("Socket", message.toString()) }
-        socket.onError { throwable, response -> Log.e("Socket", response.toString(), throwable) }
+        socket.onError { throwable, response ->
+            Log.e("Socket", response?.toString() ?: throwable.toString(), throwable)
+        }
```
boringcactus, line 25: This is a minor improvement to the way we log errors that aren't sent by the server - for a bit I was getting domain resolution failures, and those were just logging as `null` since `response` was null, and that's not useful.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
```diff
@@ -88,6 +88,7 @@ class PredictionsViewModel(
 
     fun disconnect() {
         predictionsRepository.disconnect()
+        _predictions.value = null
```
boringcactus, line 91: Resetting the predictions on disconnect prevents showing stale data while reconnecting - I'm surprised we didn't have this in place before.

KaylaBrady: Was this presenting issues as is for your changes? Looking at the iOS implementation, [leavePredictions](https://github.com/mbta/mobile_app/blob/5108bc6e9520e10055c4ee1f8a19bc20837a6b31/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift#L270) only disconnects from the channel. 

I thought we wanted to keep predictions around for the case where a user backgrounds the app / disconnects from the channel briefly, then returns and should still see the latest predictions.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
KaylaBrady: That is tricky. To clarify my own understanding, that would be an issue even if we weren't using koin and were instantiating the view models directly, is that right? I haven't been using koinViewModel regularly, and wondering about differences / benefits we get with it aside from test defaults.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
boringcactus: Yes; the existing behavior was causing "Predictions unavailable" to display while the predictions channel reconnects, because as is we don't [reset the predictions](https://github.com/mbta/mobile_app/blob/5108bc6e9520e10055c4ee1f8a19bc20837a6b31/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift#L213) when the location gets reset. I can scope this a little more narrowly to avoid that issue without breaking the background behavior, though.

KaylaBrady: >    // reset predictions if nearby stop list has been reset

I'm struggling a bit with this logic. Couldn't the stop list be reset to a new list rather than null?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
boringcactus: I believe that this is inherent to the Android ViewModel subsystem and not specific to Koin's integration with it. My understanding is that using `koinViewModel` lets us pass repositories via Koin into the ViewModel without having to define a custom `ViewModelProvider.Factory` or deal with `CreationExtras`.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/state/subscribeToPredictions.kt
boringcactus: `NearbyTransitViewModel.reset` sets `nearby = null`, which means that the `NearbyTransitView`'s `stopIds` will be `null`. Do you find 27c34a54 clearer?

KaylaBrady: Yes, thank you!

KaylaBrady (approved):

PR #647, opened 2025-01-13
fix(Android): sheet drag handle padding

BrandonTR:
### Summary

_Ticket:_ [Padding around handle and/or nearby transit header is too big](https://app.asana.com/0/1205425564113216/1209092938498574/f)

Fixes the drag handle padding in the sheet. I opted to follow this [spec](https://www.figma.com/design/eFiLM8xjo9hKprg93BBCtZ/App-UI-Design?node-id=3719-635167&t=SazsnJp5On2eXxhB-4).

iOS

android
~~- [ ] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible~~

### Testing

Manually verified drag handle matches spec.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/DragHandle.kt
```diff
@@ -1,23 +1,41 @@
...
+        modifier =
+            Modifier.height(36.dp).fillMaxWidth().background(colorResource(id = R.color.fill1)),
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.Center
     ) {
-        BottomSheetDefaults.DragHandle()
+        val description = stringResource(R.string.drag_handle)
+        Surface(
+            modifier = Modifier.semantics { contentDescription = description },
+            color = colorResource(R.color.neutral75),
```
EmmaSimon, line 35: The color spec seems to use neutral 45 for dark mode in the [colors section](https://www.figma.com/design/eFiLM8xjo9hKprg93BBCtZ/App-UI-Design?node-id=1232-87717&t=6QKXObnVI5JTApq1-4), but there's no role specified for those colors. I don't think we're setting this at all in iOS and are just using the default color, so we can't really compare what it's doing.

BrandonTR: Hmmm, maybe it's worth making a one off color resource called `drag_handle` with the respective light/dark colors and if this ends up being needed for scenarios outside of the drag handle we can consider giving it a role? Might be worth a quick standup discussion too.

EmmaSimon: Yeah that seems fine to me! It's probably not really worth bringing up at standup. 🤷

EmmaSimon (approved):

PR #652, opened 2025-01-14
feat(analytics): other iOS GA updates

boringcactus:
### Summary

_Ticket:_ [GA Updates](https://app.asana.com/0/1205732265579288/1209010745257736/f)

Handles the Search, Combined Stop + Trip Details, and User settings elements. Nearby Transit was covered in #650 since it had some architectural consequences that these do not.

It's not clear that there's actually a way to track things per session, so I'm [tracking things per user](https://firebase.google.com/docs/analytics/user-properties?platform=ios) instead. I think I've set up the custom definitions correctly in GA, but I'm not sure how to be certain.

iOS

android

### Testing

Manually verified that everything is called when it should be.


iosApp/iosApp/LocationDataManager.swift
```diff
@@ -17,23 +17,30 @@ public class LocationDataManager: NSObject, LocationFetcherDelegate, ObservableO
...
         super.init()
         self.locationFetcher.locationFetcherDelegate = self
     }
 
     public func locationFetcherDidChangeAuthorization(_ fetcher: LocationFetcher) {
         authorizationStatus = fetcher.authorizationStatus
+        analytics.recordSession(
+            locationAccess: fetcher.authorizationStatus,
+            locationAccuracy: fetcher.accuracyAuthorization
+        )
```
boringcactus, line 43: For some reason, this gets called far more often than the authorization actually changes; this shouldn't actually be a problem, but it does flood the log with `11.5.0 - [FirebaseAnalytics][I-ACS023081] Analytics is disabled. User property not set` repeatedly when running the staging app locally; I could try to check that things have actually changed before sending the variable to Firebase if we thought it was worth the effort.

JackVCurtis (approved):
👍

iosApp/iosApp/LocationDataManager.swift
JackVCurtis: That seems like a good quick fix, and close enough to the ticket scope. If it turns out to not be quick for whatever eldritch reason we can always just make it a ticket.

boringcactus: Remembering the previous `fetcher.authorizationStatus` and `fetcher.authorizationAccuracy` didn't actually work - these were still being sent to Firebase way too often - but some judicious breakpoints led me to https://github.com/mbta/mobile_app/blob/8b6bc35ffd375848fd36b1cfafb5bdca86ea1478/iosApp/iosApp/Pages/Map/HomeMapView.swift#L65 as the culprit - we're creating a new `LocationDataManager` every time the `ContentView` renders.

PR #654, opened 2025-01-15
feat: set up logic for new feature promotion

boringcactus:
### Summary

_Ticket:_ [Set up logic for new feature promo](https://app.asana.com/0/1205732265579288/1209120993502865/f)

I am happy with this structure, which means I may have overengineered it.

iOS

android

### Testing

Added some unit tests. Manually verified that the versions are being stored on both Android and iOS and therefore if we ship this in 1.1.5 we should be able to use it in 1.2.0.

### DIFF
--- a/iosApp/iosApp/ViewModels/ContentViewModel.swift
+++ b/iosApp/iosApp/ViewModels/ContentViewModel.swift
@@ -13,20 +13,26 @@ import shared
 class ContentViewModel: ObservableObject {
     @Published var configResponse: ApiResult<ConfigResponse>?
     @Published var hideMaps: Bool
+    @Published var featurePromosPending: [FeaturePromo]?
     @Published var onboardingScreensPending: [OnboardingScreen]?
 
     var configUseCase: ConfigUseCase
+    var featurePromoUseCase: FeaturePromoUseCase
     var onboardingRepository: IOnboardingRepository
     var settingsRepository: ISettingsRepository
 
     init(configUseCase: ConfigUseCase = UsecaseDI().configUsecase,
          configResponse: ApiResult<ConfigResponse>? = nil,
+         featurePromoUseCase: FeaturePromoUseCase = UsecaseDI().featurePromoUsecase,
+         featurePromosPending: [FeaturePromo]? = nil,
          onboardingRepository: IOnboardingRepository = RepositoryDI().onboarding,
          onboardingScreensPending: [OnboardingScreen]? = nil,
          settingsRepository: ISettingsRepository = RepositoryDI().settings,
          hideMaps: Bool = false) {
         self.configUseCase = configUseCase
         self.configResponse = configResponse
+        self.featurePromoUseCase = featurePromoUseCase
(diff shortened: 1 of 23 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/FeaturePromo.kt
```diff
@@ -0,0 +1,17 @@
+package com.mbta.tid.mbta_app.model
+
+enum class FeaturePromo(val addedInVersion: AppVersion) {
+    // ktfmt does not like empty enums as of the version we get with spotless
+    Nothing(AppVersion(0u, 0u, 0u));
```
KaylaBrady, line 5: suggestion(non-blocking): To confirm my understanding, having a single AppVersion works for a feature promo b/c we have previously discussed that iOS and android should only ever differ in the patch number, correct? With that, small suggestion to document that in the deploy section of the readme.

iosApp/iosApp/ViewModels/ContentViewModel.swift
```diff
@@ -45,6 +51,10 @@ class ContentViewModel: ObservableObject {
         }
     }
 
+    @MainActor func loadFeaturePromos() async {
+        featurePromosPending = await (try? featurePromoUseCase.getFeaturePromos()) ?? []
```
KaylaBrady, line 55: suggestion: Maybe worth a test that these are loaded as expected

KaylaBrady (approved):
I like the approach you took here & don't think it was over-engineered! I was originally thinking of feature promotion as a version of an onboarding screen, but I think it is very reasonable to make them entirely separate.

iosApp/iosApp/ViewModels/ContentViewModel.swift
boringcactus: I mean, I added [a test](https://github.com/mbta/mobile_app/pull/654/files#diff-797b98f92f1d0e257b46b509e4852a85f686c4dabaf371fa21420d77e3559724) that these are loaded as empty; there are no features to promote (and there is no version before `Nothing` was added, since the version numbers are unsigned) so the list can only ever be empty right now.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/FeaturePromo.kt
boringcactus: That combined with the principle that any feature worth promoting is also worth bumping the minor version number for, yeah. That said, though, promoting combined stop + trip details on Android would be incorrect, since it never had anything else, and representing that accurately under the assumption that version numbers are only ever synchronized would be somewhat difficult, so it may be simpler to just allow separate version numbers to be used.

iosApp/iosApp/ViewModels/ContentViewModel.swift
KaylaBrady: Oh sorry, I totally overlooked that test somehow - that looks great to me!

PR #655, opened 2025-01-15
fix(Android): Layer search bar under sheet

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Small UI fixes before Nearby Transit QA - Search bar should be underneath the expanded sheet](https://app.asana.com/0/1205732265579288/1209144205590529/f)

The search bar was displaying above the sheet header when the sheet was fully expanded. This moves it underneath the sheet, but it's impossible to get the search bar to sit on a lower z index than the sheet without also including the search overlay beneath the sheet. This sidesteps that issue by adding a `Hidden` state to the sheet, which puts the sheet entirely off of the screen, and allows the search result overlay to take up the full screen size.

iOS

android


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/sheet/BottomSheetScaffold.kt
```diff
@@ -216,12 +216,22 @@ private fun StandardBottomSheet(
...
-                    val newTarget = state.anchoredDraggableState.targetValue
+                    var newTarget = state.anchoredDraggableState.targetValue
+
+                    val newAnchors: DraggableAnchors<SheetValue> =
+                        if (state.currentValue == SheetValue.Hidden) {
+                            newTarget = SheetValue.Hidden
+                            DraggableAnchors {
+                                SheetValue.Hidden at layoutHeight
+                                SheetValue.Medium at layoutHeight / 2
+                            }
```
boringcactus, line 227: If we decide later that we want to return the sheet to its prior height when returning from search, instead of always using `Medium`, the lack of `Small` and `Large` anchors here might cause problems that would be unpleasant to debug. It might work to do something like
```kotlin
DraggableAnchors {
    if (state.currentValue == SheetValue.Hidden) {
        SheetValue.Hidden at layoutHeight
    }
    SheetValue.Small at ...
}
```
since the `at` function itself is what creates the anchor.

boringcactus (commented):
If I'd remembered that we needed this I might have kept `Hidden` around from the stock Android sheet in #604. I'm not sure I would've been thinking far enough ahead to allow automatically hiding the sheet but prevent manually hiding the sheet, though.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/sheet/BottomSheetScaffold.kt
EmmaSimon: Yeah good call, I condensed it a little bit.

boringcactus (approved):

PR #658, opened 2025-01-16
refactor: move analytics abstractions to shared

boringcactus:
### Summary

_Ticket:_ [GA for Android](https://app.asana.com/0/1205425564113216/1208968297244781/f)

This will let us be confident we're using consistent parameter names between iOS and Android.

I've jumped through a few hoops to preserve the `class_name` parameter for existing analytics events, but coming up with a value to use there for newly created events may not be straightforward, and we should probably pull these out into constants if they are in fact load-bearing; I'll ask at standup if these are important.

iOS

android

### Testing

Checked that all the code still compiles and logs that it's not sending events due to running in debug mode at the times I'd expect it to try to send events. Updated the one existing unit test that touches analytics to pass with the new setup.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/AnalyticsColorScheme.kt
```diff
@@ -0,0 +1,12 @@
+package com.mbta.tid.mbta_app.analytics
+
+enum class AnalyticsColorScheme(val recordedValue: String) {
+    Light("light"),
+    Dark("dark"),
+    /**
+     * Swift requires forwards compatibility for the SwiftUI ColorScheme enum; presumably Apple is
+     * planning to implement an [Aing-Tii](https://starwars.fandom.com/wiki/Aing-Tii/Legends) color
+     * scheme in iOS 50.
```
BrandonTR, line 9: 😆

BrandonTR (approved):
Nice!

PR #659, opened 2025-01-16
feat(Android): set up Firebase

boringcactus:
### Summary

_Ticket:_ [GA for Android](https://app.asana.com/0/1205425564113216/1208968297244781/f)

I'm keeping the same values in-tree vs secret for Android as we already are for iOS.

We'll need to set up the relevant environment variable secrets in GitHub Actions; I don't have permission to do that.

iOS

android

### Testing

Checked that the Firebase Analytics API successfully loads when referenced.


iosApp/ci_scripts/ci_post_clone.sh
```diff
@@ -89,4 +89,4 @@ touch .envrc
 echo "export SENTRY_DSN_IOS=${SENTRY_DSN}" >> .envrc
 echo "export SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}" >> .envrc
 echo "export FIREBASE_KEY=${FIREBASE_KEY}" >> .envrc
-echo "export GOOGLE_APP_ID=${GOOGLE_APP_ID}" >> .envrc
+echo "export GOOGLE_APP_ID_IOS=${GOOGLE_APP_ID_IOS}" >> .envrc
```
boringcactus, line 92: We could preserve backwards compatibility with the Xcode Cloud workflow with `export GOOGLE_APP_ID_IOS=${GOOGLE_APP_ID}` but I think that that could get confusing.

androidApp/build.gradle.kts
```diff
@@ -146,11 +148,12 @@ task("envVars") {
         println(".envrc file not configured, reading from system env instead")
     }
 
+    fun getPropsOrEnv(key: String): String? = props.getProperty(key) ?: System.getenv(key)
```
KaylaBrady, line 151: Nice!

KaylaBrady (approved):

iosApp/ci_scripts/ci_post_clone.sh
boringcactus: Actually, though, that might be a good idea so that we can merge this PR before changing the Xcode Cloud workflow instead of breaking the Xcode Cloud workflow for other PRs while this one is still in flight.

PR #653, opened 2025-01-15
refactor(android.NearbyTransitPage): navigation refactor to support trip filter

KaylaBrady:
### Summary

_Ticket:_ [🤖 | stop + trip details | barebones UX](https://app.asana.com/0/1205732265579288/1209118361132266/f)

What is this PR for?

This is a first step in getting the barebones UX trip details.
1. Rename filter => stopFilter and add tripFilter 
2. Rework navigation to pass around `SheetRoute` instead of `NavBackStackEntry`. This made it easier to read the filters that are set on the stop details pages.
3. Copy `setStopFilter` logic from iOS.
iOS
- [ ] If you added any user-facing strings on iOS, are they included in Localizable.xcstrings?
  - [ ] Add temporary machine translations, marked "Needs Review"

android
- [ ] All user-facing strings added to strings resource
- [ ] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible

### Testing

What testing have you done?
* Added unit tests
* Ran locally and confirmed navigation between nearby transit & stop details still works as expected
    * Note: the back button now works as a close button since on iOS there is no longer a back button on stop details.

PR #661, opened 2025-01-17
refactor(StopDetailsPage): Pull data into VM

KaylaBrady:
### Summary

_Ticket:_ [🤖 | stop + trip details | barebones UX](https://app.asana.com/0/1205732265579288/1209118361132266/f)

What is this PR for?

This PR moves the data management of the `StopDetailsPage` out of the component and into a new `StopDetailsViewModel`. This will be helpful for organization going forward and is key to resolving an issue with https://github.com/mbta/mobile_app/pull/653, where changing a filter within the stop details page (like tapping a route pill) caused the whole page to reload.

Heavily referenced [StopDetailsViewModel.swift](https://github.com/mbta/mobile_app/blob/main/iosApp/iosApp/ViewModels/StopDetailsViewModel.swift), pulling in the minimum logic necessary to resolve this one particular issue and unblock other work. We can pull more into the view model as we go.

iOS

android

### Testing

What testing have you done?
* Ran locally and confirmed predictions, schedules all loaded on nearby transit & stop details when navigating back and forth.
* Added unit tests

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitTabViewModel.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitTabViewModel.kt
@@ -8,6 +8,7 @@ import com.mbta.tid.mbta_app.model.StopDetailsDepartures
 import com.mbta.tid.mbta_app.model.StopDetailsFilter
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
 
 class NearbyTransitTabViewModel : ViewModel() {
 
@@ -16,17 +17,23 @@ class NearbyTransitTabViewModel : ViewModel() {
 
     private val _currentNavEntry = MutableStateFlow<SheetRoutes?>(null)
     val currentNavEntry: StateFlow<SheetRoutes?> = _currentNavEntry
+    private val _previousNavEntry = MutableStateFlow<SheetRoutes?>(null)
+    val previousNavEntry: StateFlow<SheetRoutes?> = _previousNavEntry
 
     /**
-     * Record the current sheetRoute. Calling this function does *not* affect the navigation stack.
-     * It as a helper function to more conveniently access the SheetRoute directly rather than
-     * reading off `NavBackStackEntry`, where it may not always be known what SheetRoute type to
-     * resolve to.
+     * Record the current `navEntry` and archive the `previousNavEntry` Calling this function does
+     * *not* affect the navigation stack. It as a helper function to more conveniently access the
+     * SheetRoute directly rather than reading off `NavBackStackEntry`, where it may not always be
+     * known what SheetRoute type to resolve to.
      */
 
     // https://stackoverflow.com/a/78911122
     fun recordCurrentNavEntry(sheetRoute: SheetRoutes?) {
-        _currentNavEntry.value = sheetRoute
+
+        _currentNavEntry.update {
+            _previousNavEntry.value = it
+            sheetRoute
+        }
     }
 
     fun setStopDetailsDepartures(departures: StopDetailsDepartures?) {
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitTabViewModel.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitTabViewModel.kt
@@ -40,8 +40,6 @@ class NearbyTransitTabViewModel : ViewModel() {
         _stopDetailsDepartures.value = departures
     }
 
-    // TODO: Move these to separate Navigation utils helper
-
     fun setStopFilter(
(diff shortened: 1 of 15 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
```diff
@@ -70,11 +70,6 @@ fun NearbyTransitView(
     val predictionsVM = subscribeToPredictions(stopIds, errorBannerViewModel = errorBannerViewModel)
     val predictions by predictionsVM.predictionsFlow.collectAsState(initial = null)
 
-    LaunchedEffect(targetLocation == null) {
```
KaylaBrady, line 73: TODO: Bring this back

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
```diff
@@ -1,5 +1,217 @@
...
+    private val predictionsRepository: IPredictionsRepository,
+    private val errorBannerRepository: IErrorBannerStateRepository,
+) : ViewModel() {
+    private val stopScheduleFetcher = ScheduleFetcher(schedulesRepository, errorBannerRepository)
+
+    private val _stopData = MutableStateFlow<StopData?>(null)
+    val stopData: StateFlow<StopData?> = _stopData
+
+    val _stopDepartures = MutableStateFlow<StopDetailsDepartures?>(null)
+    val stopDepartures: StateFlow<StopDetailsDepartures?> = _stopDepartures
```
KaylaBrady, line 44: Departures aren't stored in the iOS view model, but I think it makes things a bit easier to work with on the android side. The departures need to get passed from NearbyTransit => StopDetails so that they aren't recalculated & temporarily null when a stop filter changes, so may as well pass it on the VM.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitTabViewModel.kt
```diff
@@ -16,25 +17,29 @@ class NearbyTransitTabViewModel : ViewModel() {
...
+     * *not* affect the navigation stack. It as a helper function to more conveniently access the
+     * SheetRoute directly rather than reading off `NavBackStackEntry`, where it may not always be
+     * known what SheetRoute type to resolve to.
      */
 
     // https://stackoverflow.com/a/78911122
     fun recordCurrentNavEntry(sheetRoute: SheetRoutes?) {
-        _currentNavEntry.value = sheetRoute
+
+        _currentNavEntry.update {
```
KaylaBrady, line 33: TIL about [update](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/update.html) for atomically changing the value of state. This was more relevant for updating `stopData`, but also was convenient for updating the current nav entry & previous nav entry together.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/stopDetailsPage/StopData.kt
```diff
@@ -0,0 +1,13 @@
+package com.mbta.tid.mbta_app.model.stopDetailsPage
+
+import com.mbta.tid.mbta_app.model.response.PredictionsByStopJoinResponse
+import com.mbta.tid.mbta_app.model.response.ScheduleResponse
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class StopData(
```
KaylaBrady, line 8: Added in shared code in case we want to unite it with the swift representation. Schedules are a required field in the swift side, but it was easier to make it optional on the android side & let the predictions & schedules load independently with the way we are using coroutines.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
```diff
@@ -1,5 +1,196 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.flow.updateAndGet
+import kotlinx.coroutines.withContext
+import org.koin.androidx.compose.koinViewModel
 
-class StopDetailsViewModel() : ViewModel() {}
+class StopDetailsViewModel(
+    schedulesRepository: ISchedulesRepository,
```
EmmaSimon, line 31: Is there a reason this isn't also a `private val`?

EmmaSimon (approved):
This looks like a really good foundation!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
KaylaBrady: It isn't used as a property anywhere, so I got a suggestion not to mark it as a property.

PR #664, opened 2025-01-17
feat: analytics on Android

boringcactus:
### Summary

_Ticket:_ [GA for Android](https://app.asana.com/0/1205732265579288/1208968297244781/f)

I think this means I did a good job of designing the shared elements in #658.

iOS

android

### Testing

Verified with Firebase Analytics Debug View that all the events are being sent correctly.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainApplication.kt
```diff
@@ -34,6 +36,7 @@ class MainApplication : Application() {
 
     companion object {
         val koinViewModelModule = module {
+            single<Analytics> { AnalyticsProvider.shared }
             viewModelOf(::ContentViewModel)
             viewModelOf(::NearbyTransitViewModel)
```
boringcactus, line 41: I'm injecting the `Analytics` via Koin so that I don't have to pass it as an explicit argument to the `NearbyTransitViewModel` constructor; this may be more trouble than it's worth.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
```diff
@@ -162,13 +162,6 @@ abstract class Analytics {
         )
     }
 
-    fun toggledPinnedRouteAtStop(pinned: Boolean, routeId: String) {
```
boringcactus, line 165: This turned out to be redundant with the `togglePinnedRoute` that precedes it; we needed different names so the extensions didn't collide, but since we aren't defining each page's analytics as a separate extension on the `AnalyticsProvider` anymore, we do not need two equivalent versions of methods like this. I didn't need to make this part of this PR, but it felt like it wasn't worth spinning out into a separate PR.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainApplication.kt
boringcactus: Injecting the `Analytics` via Koin turns out to actually be a really good idea because it's a lot harder to mock out the analytics for testing without doing that.

BrandonTR (approved):

PR #669, opened 2025-01-21
Onboarding ui for android

JackVCurtis:
### Summary

_Ticket:_ [🤖 | Onboarding Flow UI](https://app.asana.com/0/1205732265579288/1208936978928069/f)

What is this PR for?

Updates the Onboarding Flow to look like the designs.
onboarding3
onboarding2
onboarding1

(Highlighting is due to the Android voice accessibility interface)

iOS

android

### Testing

What testing have you done?
Manual testing for display.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
```diff
@@ -60,46 +92,250 @@ fun OnboardingScreenView(
...
+                        )
                     }
+                    Spacer(modifier = Modifier.height(56.dp))
                 }
             }
             OnboardingScreen.Location -> {
-                Column(
-                    verticalArrangement = Arrangement.spacedBy(16.dp),
-                    horizontalAlignment = Alignment.Start
+                Box(
```
BrandonTR, line 269: Would it be a significant amount of work to split the screens into their own file? If we have to pass around too many params I'm fine with punting on this.

BrandonTR (commented):
Screenshot_20250121_160432
I think we need to add the dark mode backgrounds as well.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
JackVCurtis: Yeah, I felt the same way but opted to follow the example in the existing iOS code. I'm down to change things though.

BrandonTR (approved):
Screenshot_20250122_141253
One more small thing, the button text color seems to be opposite of the Figma spec. Other than that LGTM!

PR #671, opened 2025-01-21
fix(Android): Make entire more page rows tappable

EmmaSimon:
### Summary

_Ticket:_ [[fix: more tab links]](https://app.asana.com/0/1205425564113216/1209204151721902/f)

Previously, only the text on the more page links was tappable, tapping on the empty space or icon in the row didn't do anything. This switches to a different method for opening links and phone numbers. It also includes the padding around the toggle rows in the tappable/animated area.

android

### Testing

Verified that links still behave as expected, and are tappable along the entire row.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/more/MoreLink.kt
```diff
@@ -14,45 +17,43 @@ import androidx.compose.runtime.Composable
...
-                    if (isKey) {
-                        colorResource(R.color.key)
-                    } else {
-                        colorResource(R.color.fill3)
+            Modifier.clickable {
+                    val webpage: Uri = Uri.parse(url)
+                    val intent = Intent(Intent.ACTION_VIEW, webpage)
+                    if (intent.resolveActivity(context.packageManager) != null) {
+                        context.startActivity(intent)
                     }
```
boringcactus, line 38: From a bit more of an in-depth investigation, it looks like `resolveActivity` doesn't just check that the intent can be handled but returns the activity that will handle it, and we do not need to know which web browser the user has. It looks like if the user has no web browser installed `startActivity(intent)` will throw an `ActivityNotFoundException`, so it may be sufficient to simply `catch` and discard that exception instead of changing the manifest so we can `resolveActivity`.

EmmaSimon: Yep, that worked! Nice catch.

androidApp/src/main/AndroidManifest.xml
```diff
@@ -20,6 +20,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
+                <action android:name="android.intent.action.VIEW" />
```
boringcactus, line 23: This should probably also be taken back out.

boringcactus (approved):

PR #672, opened 2025-01-22
feat(Android): show alert details

boringcactus:
### Summary

_Ticket:_ [[android alert details view]](https://app.asana.com/0/1201654106676769/1209188093606082/f)

Adds alert details (and room for future modal sheets) to the Android app.

Sorry this PR is so big; there are a lot of alert causes.

android

### Testing

Manually verified that the Android implementation looks like the iOS implementation:

|iOS|Android|
|---|-------|
|Simulator Screenshot - iPhone 13 - 2025-01-22 at 10 03 50|Screenshot_20250122_100358|

Added the iOS alert details unit tests on Android.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredRouteView.kt
```diff
@@ -80,7 +84,23 @@ fun StopDetailsFilteredRouteView(
...
+                        alert,
+                        routeColor,
+                        modifier =
+                            Modifier.clickable(
+                                onClickLabel = stringResource(R.string.displays_more_info)
+                            ) {
+                                openAlertDetails(
+                                    ModalRoutes.AlertDetails(
+                                        alertId = alert.id,
+                                        line = patternsByStop.line,
```
EmmaSimon, line 97: Why use the whole line object but just the stop ID? On iOS we did the full objects when we probably should have just done the IDs for everything to begin with, but it seems weird to mix them like this. In #673 I added the whole stop object for consistency, though I did consider switching to IDs, but getting global data in there seemed too involved. But if you're already passing in global for the stop, maybe you also should for the line and routes.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetailsPage.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
+import kotlin.time.Duration.Companion.seconds
+
+@Composable
+fun AlertDetailsPage(
+    alertId: String,
+    line: Line?,
+    routes: List<Route>?,
+    alerts: AlertsStreamDataResponse?,
+    stopId: String?,
```
EmmaSimon, line 48: nitpick: Here and in `AlertDetails.kt`, I'd order `alerts` under `stopId` to reflect the order of `ModalRoutes.AlertDetails`

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetails.kt
```diff
@@ -0,0 +1,372 @@
...
+        AlertFooter(alert.updatedAt)
+    }
+}
+
+@Composable
+private fun AlertTitle(routeLabel: String, effectLabel: String?, causeLabel: String?) {
+    Column(verticalArrangement = Arrangement.spacedBy(10.dp)) {
+        if (effectLabel != null) {
+            Text(
+                stringResource(R.string.route_effect, routeLabel, effectLabel),
```
EmmaSimon, line 164: non-blocking: It probably makes sense to do this as a follow up, but we also need to support the [elevator closure alert behavior](https://www.figma.com/design/Y66MTe3ARhDGFxwGPhww8q/App-Wireframes?node-id=6184-32598&t=4W19fKWAFv6hWeti-4) (#673 for iOS), where this should use the stop name if the route label doesn't exist, and in the page header, show the elevator alert icon (`accessibility_icon_inaccessible`).

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetailsPage.kt
boringcactus: Makes sense.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredRouteView.kt
EmmaSimon: Yeah, we'll at least need stop name for the elevator alerts.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetails.kt
boringcactus: Yeah, I think it'll make sense to wait until it's possible to enter that state to give it the behavior it needs.

EmmaSimon (approved):

PR #678, opened 2025-01-23
fix: tweak direction picker on both iOS and Android

boringcactus:
### Summary

_Ticket:_ [🤖 | Direction picker unselected side looks wrong](https://app.asana.com/0/1205732265579288/1209226816339258/f)

Just made this task while waiting on other work to get unblocked.

Before:
Screenshot 2025-01-23 at 10 39 06 AM
After:
Screenshot 2025-01-23 at 10 31 37 AM

iOS

android

### Testing

Manually verified that this looks better than it used to.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
@@ -5,11 +5,12 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.IntrinsicSize
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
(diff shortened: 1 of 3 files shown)


iosApp/iosApp/Pages/StopDetails/DirectionPicker.swift
```diff
@@ -32,7 +32,7 @@ struct DirectionPicker: View {
     var body: some View {
         if availableDirections.count > 1 {
             let deselectedBackroundColor = deselectedBackgroundColor(route)
-            HStack(alignment: .center) {
+            HStack(alignment: .center, spacing: 2) {
```
boringcactus, line 35: This doesn't actually matter, but the default of 8 was visibly different from the Android status quo of 0, so I checked what Figma technically calls for so that we can use it on both iOS and Android.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
```diff
@@ -73,14 +78,60 @@ fun DirectionPicker(
                             contentColor =
                                 if (isSelected) Color.fromHex(route.textColor)
                                 else colorResource(R.color.deselected_toggle_text)
-                        )
+                        ),
+                    contentPadding = PaddingValues(8.dp)
```
EmmaSimon, line 82: Is the left edge getting this padding? It looks like it doesn't have quite enough, the vertical padding feels uneven compared to the horizontal.

EmmaSimon (commented):
Could you also check the [BL and CR background](https://github.com/mbta/mobile_app/blob/e8e1fe6c26e3f44424b3c21e575d1d1e7de3e8bf/iosApp/iosApp/Pages/StopDetails/DirectionPicker.swift#L71-L78) while you're at it and fix if that's wrong too?

boringcactus:
The BL/CR deselected contrast was already working, yeah.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
boringcactus: It is. If I take a screenshot and measure the pixels, there are 16 pixels between the edge of the background and the edge of the text, which would be consistent with 8 DP at a 2x pixel density. It looks better with the emulator zoomed in.

In Figma, the vertical padding is only 4px, but if I use that it looks cramped vertically (I suspect some wacky line height things are occurring). It looks better 

I wouldn't say this PR makes the direction picker perfect on Android, but it makes it better, and I do not want to spend all day counting pixels.

boringcactus: My emulator does not actually have a 2x pixel density, though; for some reason its density as reported by `LocalDensity.current` is 2.625 and the pixel count of that padding should be 21. I suspect the problem is that the gutter around the selected direction is drawn as a border around the full row instead of with padding, and the border is eating into the specified padding.

EmmaSimon: Ah, hm, okay. I'd maybe just bump it up to 12 horizontally for now, but this is definitely much better than before.

EmmaSimon (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
boringcactus: Replacing the border with padding (and updating the outer corner radius to be correct) appears to have done the trick - the padding now measures at the inexplicably-correct 21 pixels.

PR #679, opened 2025-01-23
fix(iOS): disappearing map layers when switching tabs

BrandonTR:
### Summary

_Ticket:_ [Fix: going to more tab and back in iOS removes all map layers](https://app.asana.com/0/1205732265579288/1209120993502863/f)

Usually when you have a TabView the state is preserved when switching tabs and it seemed like everything was being reset when we were switching tabs instead. This also had a pretty negative performance impact. With this workaround it seems to function as expected and perform a bit better. ~~Instead of hacking in a native TabView to the sheet I re-created it so it's just a basic control now.~~ I also made it so one main TabView persists instead of being drawn conditionally. I wanted to go a different route with this and put the custom tab bar in it's own UIWindow on top of everything but that proved to be even more hacky, needing to handle edge cases like hiding the tab bar when showing the keyboard and navigating away from nearby, passing through touch events that are not on the tab bar, handling content being covered by the tab bar, and probably more that I hadn't discovered.

iOS

android
~~- [ ] All user-facing strings added to strings resource~~
~~- [ ] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible~~

### Testing

Manual testing.


iosApp/iosApp/ComponentViews/TabBar.swift
```diff
@@ -0,0 +1,39 @@
...
+//  TabBar.swift
+//  iosApp
+//
+//  Created by Brandon Rodriguez on 1/21/25.
+//  Copyright © 2025 MBTA. All rights reserved.
+//
+
+import SwiftUI
+
+struct TabBar: View {
```
KaylaBrady, line 11: @BrandonTR one concern about using a non-native tab bar is that it doesn't have the built in accessibility features. Some that exist on the native tab bar are:
* Announcing that it is a tab bar
* Announcing which tab is selected
* Announcing "Tab 1 of 2"
* Long-press shows zoomed in version of the tab (a feature highlighted in the apple accessibility review)


If we use a custom tab bar, we'd have to implement parity for these accessibility features. I'm also not sure I understand how this fixes the problem. Playing around with this locally, if I replace the body of `nearbySheetContents` to use the old native tab bar as before, I see the route shapes persist when switching between tabs. Is there a different piece of this change that is causing the map + data to stick around?

iosApp/iosApp/ContentView.swift
```diff
@@ -107,55 +92,45 @@ struct ContentView: View {
             OnboardingPage(screens: onboardingScreensPending, onFinish: {
                 contentVM.onboardingScreensPending = []
             })
-        } else if selectedTab == .more {
+        } else {
             TabView(selection: $selectedTab) {
```
BrandonTR, line 96: @KaylaBrady I believe this is the most significant change here. We are not redrawing everything when switching tabs, instead just letting the sheet overlap and persisting the main TabView with it's subviews.

iosApp/iosApp/ComponentViews/TabBar.swift
BrandonTR: Ah didn't realize there were so many accessibility features packed into the native tab bar! Thanks for flagging.

iosApp/iosApp/ContentView.swift
KaylaBrady: Ahhh, awesome, thanks! In that case, I think worth sticking with this change & keeping the native tab bar in the nearby contents!

EmmaSimon: I think doing this also has the side effect of showing the navbar when the search overlay is open, which iirc we wanted to avoid.

KaylaBrady: Maybe we can use the [`.toolbar` modifier ](https://www.hackingwithswift.com/quick-start/swiftui/how-to-hide-the-tab-bar-navigation-bar-or-other-toolbars)to conditionally hide it in search? Or to always hide it on the nearby tab since the one in the sheet is the one that is shown

iosApp/iosApp/ContentView.swift
```diff
@@ -107,28 +92,24 @@ struct ContentView: View {
...
         }
     }
 
     @State var selectedDetent: PresentationDetent = .halfScreen
     @State var visibleNearbySheet: SheetNavigationStackEntry = .nearby
     @State private var showingLocationPermissionAlert = false
 
-    @ViewBuilder var nearbySheetContents: some View {
-        // Putting the TabView in a VStack prevents the tabs from covering the nearby transit contents
-        // when re-opening nearby transit
```
KaylaBrady, line 131: suggestion (non-blocking): I think it would be worth still keeping this comment & the ViewBuilder annotations

KaylaBrady (approved):
Awesome!

PR #685, opened 2025-01-27
feat(Android): bare trip details UX

boringcactus:
### Summary

_Ticket:_ [🤖 | stop + trip details | Trip details UX](https://app.asana.com/0/1205732265579288/1209118487744428/f)

Adds the list of upcoming predictions with a collapsible section for stops between the current and target stop.

I'm not sure how confident we are in these screen reader overrides, but it seems likely that we'll need some screen reader overrides, and we may as well be consistent between platforms. The temporary translations seem to have been skipped when these were first added.

iOS

android

### Testing

Manually verified that the stops are listed and the collapsible section works. Added equivalents to the iOS unit tests where feasible.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripDetailsView.kt
```diff
@@ -29,39 +32,51 @@ fun TripDetailsView(
...
-            val vehicleStop = getParentFor(vehicle?.stopId, globalResponse.stops)
-            val tripId = tripFilter.tripId
-            val headerSpec: TripHeaderSpec? =
-                TripHeaderSpec.getSpec(tripId, stops, terminalStop, vehicle, vehicleStop)
-
-            TripDetailsHeader(tripId, headerSpec)
-            TripStops(stopId, stops, tripFilter.stopSequence, headerSpec, now, globalResponse)
-        } else {
-            // TODO: loading
+    val stops =
```
KaylaBrady, line 35: 👍 good call

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripDetailsView.kt
```diff
@@ -29,39 +32,51 @@ fun TripDetailsView(
...
+            stops,
+            tripFilter.stopSequence,
+            headerSpec,
+            now,
+            globalResponse,
+            routeAccents
+        )
     } else {
         // TODO: loading
+        CircularProgressIndicator()
```
KaylaBrady, line 80: Added polishing the loading state explicitly to the UI ticket

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
```diff
@@ -13,5 +32,107 @@ fun TripStops(
...
+    val showFirstStopSeparately =
+        when (headerSpec) {
+            TripHeaderSpec.FinishingAnotherTrip,
+            TripHeaderSpec.NoVehicle -> true
+            else -> false
+        }
+
+    val lastStopSequence = stops.stops.lastOrNull()?.stopSequence
+
+    Column {
```
KaylaBrady, line 66: suggestion: I know UI isn't in scope for this ticket, but may be worth adding just the background color here since it is hard to read the text against the default route color background.

KaylaBrady (approved):
Nice work! One small suggestion. I'll definitely make sure this is merged before merging https://github.com/mbta/mobile_app/pull/690.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
boringcactus: Added.

PR #692, opened 2025-01-28
feat(Android): highlight selected vehicle in map

boringcactus:
### Summary

_Ticket:_ [🤖 | stop + trip details | map vehicle focus](https://app.asana.com/0/1201654106676769/1209118361132269/f)

Per Asana discussion with Kayla, the existing behavior on both iOS and Android of drawing all relevant shapes at the stop, route, and direction but not focusing on the single shape of the selected trip is acceptable.

iOS

android

### Testing

Manually verified that the selected vehicle matches the iOS behavior of being highlighted and zoomed to but not re-zoomed to as it updates.


KaylaBrady:
@boringcactus I misinterpreted the current state of the world on the iOS side. I thought we were
1. Fetching all shapes at the stop (true)
2. Filtering to only show the shape of the selected stop (false)

This PR is a huge step forward as-is, but I think worth raising for additional consideration with product & design about the priority level of drawing only the selected trip on iOS & android.

EmmaSimon (approved):
Sick, glad it ended up being so straightforward 🙌

PR #694, opened 2025-01-28
feat(android.StopDetailsFilteredDeparturesView): Cancelled trip card

KaylaBrady:
### Summary

_Ticket:_ [🤖 | stop + trip Details | Handle cancelled trip](https://app.asana.com/0/1205732265579288/1209118487744432)

What is this PR for?

Adds a card when the selected trip is cancelled

iOS

android
- [X] All user-facing strings added to strings resource

### Testing
What testing have you done?
* Added unit tests
* Ran locally and found a cancelled trip
image


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -119,7 +119,6 @@ struct StopDetailsFilteredDepartureDetails: View {
                             icon: routeSlashIcon(routeType)
                         )
                         .accessibilityHeading(.h4)
-                        .accessibilityFocused($selectedDepartureFocus, equals: cardFocusId)
```
KaylaBrady, line 122: [Discussed in slack here](https://mbta.slack.com/archives/C062QNAJZ2M/p1738083419041429?thread_ts=1738082727.891259&cid=C062QNAJZ2M): It is not a possible state for the cancelled trip card to take focus, so removing from the iOS side & not porting to android.

androidApp/src/main/res/values/strings.xml
```diff
@@ -205,4 +205,6 @@
     <string name="waiting_to_depart">Waiting to depart</string>
     <string name="weather">Weather</string>
     <string name="westbound">Westbound</string>
+    <string name="trip_cancelled">Trip cancelled</string>
+    <string name="trip_cancelled_details">This trip has been cancelled. We\'re sorry for the inconvenience.</string>
```
KaylaBrady, line 209: TODO: update iOS conversion script to escape apostrophe

boringcactus: In general, we should probably not be using bare [`'` U+0027 APOSTROPHE](https://graphemica.com/')s: text looks better with [`’` U+2019 RIGHT SINGLE QUOTATION MARK](https://graphemica.com/%E2%80%99)s, and that's also what's in the iOS string (so it's what the conversion expects to see). We may wind up using `?`s at some point, though, so I've made a [follow up task](https://app.asana.com/0/1205425564113216/1209265099183544/f) to handle the Android-level special characters.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
```diff
@@ -112,6 +112,13 @@ data class StopDetailsDepartures(val routes: List<PatternsByStop>) {
         val stopName: String
     )
 
+    /*
+    Whether the upcoming trip with the given id is cancelled in the upcoming departures
+     */
+    fun tripIsCancelled(tripId: String): Boolean {
```
boringcactus, line 118: If we're touching the iOS code in this PR anyway, we may as well call this from the iOS side, too.

androidApp/src/main/res/values/strings.xml
KaylaBrady: Oops yeah this comment was outdated from when I first noticed the localizations weren't being pulled, it has been updated to use `’` to match what is on iOS.

androidApp/src/main/res/values/strings.xml
```diff
@@ -205,4 +205,6 @@
     <string name="waiting_to_depart">Waiting to depart</string>
     <string name="weather">Weather</string>
     <string name="westbound">Westbound</string>
+    <string name="trip_cancelled">Trip cancelled</string>
+    <string name="trip_cancelled_details">This trip has been cancelled. We’re sorry for the inconvenience.</string>
```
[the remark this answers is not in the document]
KaylaBrady: Same! I've put it off more now that sorting lines doesn't Just Work since there are pluralization entries too. I'll add it to the PR template at least

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsIconCard.kt
```diff
@@ -0,0 +1,62 @@
...
+                .border(1.dp, colorResource(R.color.halo), shape = RoundedCornerShape(8.dp))
+                .background(colorResource(R.color.fill3))
+                .padding(16.dp)
+    ) {
+        Row() {
+            icon(Modifier.weight(1f).requiredSize(35.dp), accentColor)
+            CompositionLocalProvider(
+                LocalTextStyle provides
+                    LocalTextStyle.current.copy(fontSize = 24.sp, fontWeight = FontWeight.Bold)
+            ) {
```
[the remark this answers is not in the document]
KaylaBrady: Thanks for the tip! I considered passing the `Painter` to `StopDetailsIconCard`, but we would lose the ability to set an icon-specific test tag.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt
KaylaBrady: Good call!

boringcactus (approved):

PR #695, opened 2025-01-29
fix(Android): respond to non-TalkBack P0 QA

boringcactus:
### Summary

_Ticket:_ [🤖 | respond to first batch of QA feedback](https://app.asana.com/0/1205425564113216/1209263422629657/f)

I'm doing the TalkBack fixes as a separate PR so that this PR scope doesn't get out of hand.

iOS

android

### Testing

Manually checked that all the spacing, color, alignment, and icon issues are fixed and that the new shimmer loading for individual rows in nearby transit with no data looks right.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
```diff
@@ -313,7 +316,17 @@ fun UpcomingTripView(
                         textAlign = TextAlign.End
                     )
             }
-        is UpcomingTripViewState.Loading -> CircularProgressIndicator(modifier)
+        is UpcomingTripViewState.Loading ->
+            CompositionLocalProvider(IsLoadingSheetContents provides true) {
```
KaylaBrady, line 320: suggestion (non-blocking): This could use the `import com.eygraber.compose.placeholder.material3.placeholder` modifier directly rather than having to force `IsLoadingSheetContents` state.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/TransitHeader.kt
```diff
@@ -33,37 +30,31 @@ fun TransitHeader(
...
+            modifier = Modifier.semantics { heading() }.placeholderIfLoading(),
+            style =
+                LocalTextStyle.current.copy(
+                    color = textColor,
+                    fontSize = 17.sp,
+                    fontWeight = FontWeight.Bold
+                )
+        )
+        if (rightContent != null) {
+            Spacer(modifier = Modifier.weight(1.0F))
```
KaylaBrady, line 56: thought (non-blocking): @EmmaSimon noted in a comment [here](https://github.com/mbta/mobile_app/pull/697#discussion_r1936037987) about having some issues with using a spacer before a view. I still need to play around with it more on that PR, but wanted to flag here in case it is also relevant, or to learn more about why not.

KaylaBrady (approved):
Nice! Really appreciate the way you broke down these commits & broke out the Talk Back changes.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/UpcomingTripView.kt
boringcactus: If we ever decide we want our placeholders to be styled in some more specific way, leaving it like this will mean we don't have to come back and change this specific one separately.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/TransitHeader.kt
boringcactus: I'm not sure what the issue Emma encountered was, but I suspect it may have been a case where multiple elements within a Row or Column had a weight assigned, and so the extra space in the container was being handled weirdly. I haven't had issues like that when the spacer is the only element with a weight.

EmmaSimon: It seems to cause problems when the text needs a line break, so if a route name overflows, it eats up the right content entirely.

Screenshot_20250131_154337

boringcactus: I don't understand how that's possible - the documentation says that `weight` only considers unused space in the container - but if I move the weight from the spacer to the text then the rightContent gets the space it needs again. How does anyone do Android development when the standard library is this broken?

boringcactus: (This spacer also wasn't added in this PR (curse Git for not treating indentation changes more reasonably) and wasn't identified as an issue in the P0 QA this is responding to.)

EmmaSimon: Yeah totally, the layout system feels really unintuitive and janky.

PR #697, opened 2025-01-29
feat(android.StopDetailsNoTripCard): Special messaging when no trips

KaylaBrady:
### Summary

_Ticket:_ [🤖 | stop + trip details | no trips cases](https://app.asana.com/0/1205732265579288/1209118508062807/f)

What is this PR for?
Copies the iOS implementation of the StopDetailsNoTripCard for android.

iOS

android
- [X] All user-facing strings added to strings resource in alphabetical order

### Testing

What testing have you done?
* Ported unit tests
* Ran locally (forcing each case for simplicity)

image
image
image
image


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/ObjectCollectionBuilder.kt
```diff
@@ -93,11 +93,11 @@ class ObjectCollectionBuilder {
 
     class LineBuilder : ObjectBuilder<Line> {
         var id = uuid()
-        var color = ""
+        var color = "FFFFFF"
```
KaylaBrady, line 96: Got an error about the default hex code not being 6 characters. We are already defaulting to these on the RouteBuilder.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsIconCard.kt
```diff
@@ -44,14 +45,17 @@ fun StopDetailsIconCard(
...
+        ) {
             CompositionLocalProvider(LocalContentColor provides accentColor) {
-                icon(Modifier.weight(1f).requiredSize(35.dp))
+                icon(Modifier.requiredSize(35.dp))
             }
             ProvideTextStyle(TextStyle(fontSize = 24.sp, fontWeight = FontWeight.Bold)) {
                 header(Modifier.semantics { heading() })
             }
-            Spacer(modifier = Modifier.weight(1.0F))
+            Spacer(Modifier.weight(1f))
```
EmmaSimon, line 58: Not really relevant here, but when I was working on the filtered stop details header change, I realized that a spacer with a weight of 1 could result in views after the spacer being shrunk down awkwardly (the close button was being shrunk into nothing), but instead putting the weight = 1 on the Text modifier and removing the spacer worked perfectly. Since we'll probably never have more things at the end of this row, it doesn't matter, but I think in general it might be best to avoid Spacers, and just put the weights directly on the text elements. Though I'm unsure what would happen if a translation here is very long and spills onto multiple lines, it might still be best to remove the spacer anyway.

EmmaSimon (approved):

PR #701, opened 2025-01-31
feat(android.feedbackFormUrl): Separate android feedback form

KaylaBrady:
### Summary

_Ticket:_ [Name](URL)

What is this PR for?
No ticket, [request on slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1738340696065619) to change the feedback URL for the beta.

iOS

android

### Testing

What testing have you done?
* Updated unit tests
* Ran locally & confirmed taken to the android form as expected
* Confirmed iOS link still goes to /appfeedback

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/morePage/feedbackFormUrl.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/morePage/feedbackFormUrl.kt
@@ -1,12 +1,14 @@
 package com.mbta.tid.mbta_app.model.morePage
 
-fun feedbackFormUrl(translation: String) =
-    when (translation) {
-        "es" -> "https://mbta.com/appfeedback?lang=es-US"
-        "ht" -> "https://mbta.com/appfeedback-ht"
-        "pt-BR" -> "https://mbta.com/appfeedback?lang=pt-BR"
-        "vi" -> "https://mbta.com/appfeedback?lang=vi"
-        "zh-Hans-CN" -> "https://mbta.com/appfeedback?lang=zh-Hans"
-        "zh-Hant-TW" -> "https://mbta.com/appfeedback?lang=zh-Hant"
-        else -> "https://mbta.com/appfeedback"
+fun feedbackFormUrl(translation: String): String {
+    val formLink = "https://mbta.com/androidappfeedback"
+    return when (translation) {
+        "es" -> "${formLink}?lang=es-US"
+        "ht" -> formLink
+        "pt-BR" -> "${formLink}?lang=pt-BR"
+        "vi" -> "${formLink}?lang=vi"
+        "zh-Hans-CN" -> "${formLink}?lang=zh-Hans"
+        "zh-Hant-TW" -> "${formLink}?lang=zh-Hant"
+        else -> formLink
     }
+}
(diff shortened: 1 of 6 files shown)


EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/morePage/feedbackFormUrl.kt
```diff
@@ -1,12 +1,14 @@
...
-    when (translation) {
-        "es" -> "https://mbta.com/appfeedback?lang=es-US"
-        "ht" -> "https://mbta.com/appfeedback-ht"
-        "pt-BR" -> "https://mbta.com/appfeedback?lang=pt-BR"
-        "vi" -> "https://mbta.com/appfeedback?lang=vi"
-        "zh-Hans-CN" -> "https://mbta.com/appfeedback?lang=zh-Hans"
-        "zh-Hant-TW" -> "https://mbta.com/appfeedback?lang=zh-Hant"
-        else -> "https://mbta.com/appfeedback"
+fun feedbackFormUrl(translation: String): String {
+    val formLink = "https://mbta.com/androidappfeedback"
```
boringcactus, line 4: We do not want to send iOS users to the Android app feedback form.

boringcactus (changes requested):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/morePage/feedbackFormUrl.kt
KaylaBrady: 😱 Should have checked that this is in shared, thank you!

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/morePage/FeedbackFormUrlTest.kt
```diff
@@ -3,44 +3,44 @@ package com.mbta.tid.mbta_app.model.morePage
...
     fun testES() {
-        assertEquals("https://mbta.com/appfeedback?lang=es-US", feedbackFormUrl("es"))
+        assertEquals("baseUrl?lang=es-US", localizedFeedbackFormUrl("baseUrl","es"))
     }
 
     @Test
     fun testHT() {
-        assertEquals("https://mbta.com/appfeedback-ht", feedbackFormUrl("ht"))
+        assertEquals("baseUrl", localizedFeedbackFormUrl("baseUrl", "ht"))
     }
```
boringcactus, line 20: This should be returning `"baseUrl-ht"` on iOS; I'm not sure how we want to handle that on Android, but it might be easiest to create an `androidappfeedback-ht`.

EmmaSimon: I think it's still technically possible for Android to support Haitian Creole, it's just that most phones don't include it in the stock rom, so this is probably a good idea anyway.

boringcactus (approved):

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/morePage/FeedbackFormUrlTest.kt
KaylaBrady: Oh @EmmaSimon I missed your comment before making this change, will raise in the slack thread.

EmmaSimon: It's possible to select it in android studio at least, but I'm not sure that means much.

KaylaBrady:
Ran locally with all NearbyViewModelTests passing. These changes did not touch the iOS NearbyViewModel, so I think that failure is a red herring. Going to merge to get the android build going & will re-run these tests to verify.
image

PR #702, opened 2025-01-31
feat(iOS): elevator alert UI

boringcactus:
### Summary

_Ticket:_ [ | elevator alerts UI](https://app.asana.com/0/1205732265579288/1209230846082281)

This was not as straightforward as I had expected.

In alert details, the designs call for showing the elevator facility info, but we don't have that. The one currently available elevator closure alert has the facility info and cause in the description, though, so it seems like showing that may be close enough for now.

Some of the provisional translations of "N elevators closed" are the same as the old provisional translations of "N elevator closures" - I'm not sure if those languages really don't have that distinction or if our provisional translations are just not all that great.

iOS

android

### Testing

Manually verified that things work in roughly the way they're supposed to. Added and updated unit tests.


iosApp/iosApp/Pages/AlertDetails/AlertDetails.swift
```diff
@@ -268,12 +279,17 @@ struct AlertDetails: View {
...
+                if !isElevatorClosure {
+                    alertPeriod
+                    VStack(alignment: .leading, spacing: 16) {
+                        affectedStopCollapsible
+                        tripPlannerLink
+                    }
                 }
                 alertDescription
+                if isElevatorClosure {
+                    alertPeriod
```
KaylaBrady, line 291: The alert period lis a bit squished. I see this also is an issue for other alerts. If minor, could you include the fix here?
IMG_0222

KaylaBrady (approved):
LGTM! I realize the AC for this ticket included onboarding UI, but this PR very reasonably doesn't have that since the onboarding UX ticket isn't done.

iosApp/iosApp/Pages/AlertDetails/AlertDetails.swift
boringcactus: I did wonder when I implemented this on Android why the structure was the way that it was on iOS. This seems better:
Screenshot 2025-01-31 at 2 43 40 PM

PR #706, opened 2025-02-03
feat: onboarding and settings for elevator alerts

boringcactus:
### Summary

_Ticket:_ [Elevator Status | UX + UI | Onboarding page + move feature flag to settings](https://app.asana.com/0/1205732265579288/1209144608947397)

Adds the onboarding screen and moves the feature flag into settings.

I really like implementing things on both platforms at once.

iOS

android

### Testing

Added and updated unit tests. Manually verified that the new screen appears and looks correct on both Android and iOS and that the setting is displayed in the correct section.


iosApp/iosApp/Icons.xcassets/accessibility-icon-accessible.imageset/Glyph=ISA, Background=Shape.svg
boringcactus, line 1: Since iOS does not process .svg assets as vectors, rendering this 24x24 icon at 192x192 looks [bad](https://mbta.slack.com/archives/C062QNAJZ2M/p1738259599261969), so I've scaled it up in Inkscape as `-large`; I'm leaving the normal sized version here even though this PR doesn't use it, because we'll want it eventually.

EmmaSimon: I'm probably being overcautious, but It might be worth asking Adam for another export of this at the larger size. I've had issues with inkscape adding weird properties that cause problems in the past (I'm mostly concerned about the `style="stroke-width:8"` ones that were added). I tried exporting it from the onboarding page in figma, but that also included some drop shadow stuff that could also have compatibility issues.

And just to double check, we don't have a similar issue on Android? I'm a little surprised they would be doing something more sensible with image rendering.

iosApp/iosApp/Icons.xcassets/accessibility-icon-accessible.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images": [
+    {
+      "filename": "Glyph=ISA, Background=Shape.svg",
```
EmmaSimon, line 4: nitpick: This doesn't matter even a little bit, but this file name makes me sad

EmmaSimon (approved):

iosApp/iosApp/Icons.xcassets/accessibility-icon-accessible.imageset/Contents.json
boringcactus: It makes me sad too.

PR #713, opened 2025-02-05
feat(iOS): Combined stop + trip promo UX

EmmaSimon:
### Summary

_Ticket:_ [ | Combined Stop + Trip Details: Launch promo](https://app.asana.com/0/1205732265579288/1209309195022071)

Adds the UX for the combined stop + trip promo page. We don't have a finalized graphic yet, so that's left as a todo, and it's only set to show up once we update the version number to 1.2.0, so currently you can only make it show up by manually updating the version locally.

Simulator Screenshot - iPhone 15 - 2025-02-05 at 16 20 01

iOS

### Testing

Added some basic UI tests

### DIFF
--- a/androidApp/src/main/res/values/strings.xml
+++ b/androidApp/src/main/res/values/strings.xml
@@ -140,7 +140,7 @@
     <string name="police_action">Police Action</string>
     <string name="police_activity">Police Activity</string>
     <string name="power_problem">Power Problem</string>
-    <string name="predictions_unavailable_details">Service is running, but predicted arrival times aren’t available. The map shows where %1$s on this route currently are.</string>
+    <string name="predictions_unavailable_details">Service is running, but predicted arrival times aren’t available. Check the map to see where %1$s are right now.</string>
     <string name="predictions_unavailable_details_hide_maps">Service is running, but predicted arrival times aren’t available.</string>
     <string name="real_time_arrivals_updating_live">Real-time arrivals updating live</string>
     <string name="recenter" tools:ignore="MissingTranslation">Recenter</string>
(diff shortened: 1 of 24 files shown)


androidApp/src/main/res/values/strings.xml
```diff
@@ -140,7 +140,7 @@
     <string name="police_action">Police Action</string>
     <string name="police_activity">Police Activity</string>
     <string name="power_problem">Power Problem</string>
-    <string name="predictions_unavailable_details">Service is running, but predicted arrival times aren’t available. The map shows where %1$s on this route currently are.</string>
+    <string name="predictions_unavailable_details">Service is running, but predicted arrival times aren’t available. Check the map to see where %1$s are right now.</string>
```
boringcactus, line 143: Probably worth running the Android build to update the Android translations so they aren't stale in-tree.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/FeaturePromo.kt
```diff
@@ -2,7 +2,8 @@ package com.mbta.tid.mbta_app.model
 
 enum class FeaturePromo(val addedInAndroidVersion: AppVersion, val addedInIosVersion: AppVersion) {
     // ktfmt does not like empty enums as of the version we get with spotless
-    Nothing(AppVersion(0u, 0u, 0u));
+    Nothing(AppVersion(0u, 0u, 0u)),
+    CombinedStopAndTrip(AppVersion(0u, 0u, 0u), AppVersion(1u, 2u, 0u));
```
boringcactus, line 6: Now that there's a real feature in the list, we shouldn't need `Nothing` or the comment.

boringcactus (approved):

androidApp/src/main/res/values/strings.xml
EmmaSimon: I didn't actually bother updating the translations for this, since we're just about to send out the strings to get translated, and it's a relatively minor copy change on text that won't be released until after we get back official translations.

PR #714, opened 2025-02-05
feat(Android): departure tiles

boringcactus:
### Summary

_Ticket:_ [🤖 | stop + trip details | departure tiles ](https://app.asana.com/0/1205732265579288/1209118480622196)

TileData probably belongs in the shared code, but we can worry about that later.

iOS

android

### Testing

Manually verified that the tiles behave properly when on the Green Line and when headsigns all match. Ported unit tests over from iOS.


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
```diff
@@ -0,0 +1,9 @@
+package com.mbta.tid.mbta_app.android
+
+import androidx.compose.ui.semantics.SemanticsActions
+import androidx.compose.ui.test.SemanticsMatcher
+
+fun hasClickActionLabel(expected: String?) =
```
KaylaBrady, line 6: Nice!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DepartureTile.kt
```diff
@@ -0,0 +1,202 @@
...
+        verticalArrangement = Arrangement.spacedBy(4.dp, Alignment.CenterVertically)
+    ) {
+        CompositionLocalProvider(
+            LocalContentColor provides
+                if (isSelected) colorResource(R.color.text)
+                else colorResource(R.color.deselected_toggle_text)
+        ) {
+            // If you can figure out a better way to make the pill left aligned to the text and the
+            // status right aligned to the text whether the text is wider or narrower than the pill
+            // + status, please do.
```
KaylaBrady, line 88: Messing around with it a bit locally, I think `width(IntrinsicSize.Max)` on the Column and `fillMaxWidth()` on the headsign text will have the effect you're looking for

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
```diff
@@ -94,12 +104,21 @@ fun StopDetailsFilteredDeparturesView(
...
     val routeTextColor: Color = Color.fromHex(textHex)
 
+    // keys are trip IDs
+    val bringIntoViewRequesters = remember { mutableStateMapOf<String, BringIntoViewRequester>() }
+
+    LaunchedEffect(tripFilter) {
+        if (tripFilter != null) {
+            bringIntoViewRequesters[tripFilter.tripId]?.bringIntoView()
+        }
+    }
```
KaylaBrady, line 114: I noticed that after scrolling the list a bit and selecting a tile in the middle of the scroll view, it is scrolled so that the selected tile is at the end of the view which felt a little unexpected. Is there a way around that?

https://github.com/user-attachments/assets/10679c0a-00a5-4636-9a5e-935033bd01cd

EmmaSimon: Unrelated but noticed from the video that the tiles should also scroll from edge to edge, without getting cut off by side padding. But that'll involve broader padding/layout changes, so maybe it would be best to save it for a final QA pass where we tweak the padding of everything to get it to play nice in different states and match the spec.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DepartureTile.kt
boringcactus: That did it, thanks! I don't think I'd used `IntrinsicSize` before; I'll definitely have to remember that.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
boringcactus: I don't know why the scroll position is being reset like that. I thought it might be that the `rememberScrollState` was too far down in the hierarchy and it was being forgotten when the `DepartureTiles` helper component's identity was changing somehow, but moving it further up didn't appear to help.

Scrolling from edge to edge isn't all that much work, though, I can sneak that in here.

boringcactus: I was right, I just didn't hoist it high enough. Since the trip filter is an argument to everything from the stop details page on down, I think the component identity of every single thing in the sheet resets when the trip filter changes. That's going to be annoying for the rest of the time we spend working on the app, so it might be worth considering other approaches, but certainly not now.

PR #725, opened 2025-02-09
Disable RTL support

amire80:
At the moment, the app doesn't support any languages that are written from right to left, such as Hebrew or Arabic.

However, AndroidManifest.xml says android:supportsRtl="true", which is the default value. This causes the app to display incorrectly in English on phones that are set to work in a right to left language.

This patch sets rtlandroid:supportsRtl="false",
so that the app won't try to flip itself for
right to left languages unnecessarily.

If full support for such languages is ever added,
set it back to "true".


KaylaBrady (approved):
Nice, thanks @amire80!

amire80:
> Nice, thanks @amire80!

Thanks for approving! Is there anything I can do to fix the CI failure?

KaylaBrady:
@amire80 I am working on a fix for the CI issue, sorry for the delay!

KaylaBrady:
Thanks again @amire80!

amire80:
Thank you for the quick review and the CI fixes!

amire80:
I have the updated version on my phone, and it's fixed! Thank you again!

PR #726, opened 2025-02-10
enable edge-to-edge map coverage of the screen

JackVCurtis:
### Summary

_Ticket:_ [Android | Map extends to top of screen](https://app.asana.com/0/1205732265579288/1209343093134602)

What is this PR for?

Removes the black bar behind the top system icons and extends the map to cover the whole screen.

Screenshot_20250210_104304
Screenshot_20250210_103837

iOS

android

### Testing

What testing have you done?

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
@@ -32,12 +32,12 @@ class MainActivity : ComponentActivity() {
             navigationBarStyle =
                 if (isDarkModeOn()) {
                     SystemBarStyle.dark(
-                        scrim = Color(0xFF192026).toArgb(),
+                        scrim = getColor(R.color.fill2),
                     )
                 } else {
                     SystemBarStyle.light(
-                        scrim = Color(0xFFF5F4F2).toArgb(),
-                        darkScrim = Color(0xFF192026).toArgb()
+                        scrim = getColor(R.color.fill2),
+                        darkScrim = getColor(R.color.fill2)
                     )
                 },
             statusBarStyle =
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
@@ -1,5 +1,6 @@
 package com.mbta.tid.mbta_app.android
 
+import android.content.res.Configuration
 import android.os.Bundle
 import android.util.Log
 import androidx.activity.ComponentActivity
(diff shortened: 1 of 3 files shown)


boringcactus:
With 3-button navigation instead of gesture navigation (Settings > System > Gestures > System navigation), this has an issue I expected and an issue I didn't:
|main|this PR|
|----|-------|
|Screenshot_20250211_121408|Screenshot_20250211_121059|

The bottom navigation bar looks wrong, which I thought it might, but also, the halo around the search bar is in the wrong position. It looks like `Modifier.navigationBarsPadding()` might fix the navigation if you put it in the right place, and if that doesn't fix the halo then it may be necessary to use something like `Modifier.onPlaced` on the SearchBar to dynamically size the halo to match (although there's like a 15% chance that our `Modifier.haloContainer` will magically work on the SearchBar and we can ditch the separate halo box completely).

JackVCurtis:
Should be fixed now @boringcactus 
Screenshot_20250214_105229
Screenshot_20250214_104846

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
```diff
@@ -22,10 +28,35 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
         initSentry()
         fusedLocationClient = LocationServices.getFusedLocationProviderClient(this)
+        enableEdgeToEdge(
+            navigationBarStyle =
+                if (isDarkModeOn()) {
+                    SystemBarStyle.dark(
+                        scrim = Color(0xFF192026).toArgb(),
```
boringcactus, line 35: I'm like 80% sure that `Color(0xFF192026).toArgb()` will be `0xFF192026`, and also, I think `getColor(R.color.fill2)` would be a little more legible.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/SearchBarOverlay.kt
```diff
@@ -59,6 +61,10 @@ fun SearchBarOverlay(
         remember(currentNavEntry) {
             currentNavEntry?.let { it is SheetRoutes.NearbyTransit } ?: true
         }
+    val navigationBarHeight =
+        with(LocalDensity.current) {
+            WindowInsets.navigationBars.getBottom(Density(LocalContext.current)).toDp()
+        }
```
boringcactus, line 67: Is this variable used?

JackVCurtis: Nope, was used for the now-removed background

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
JackVCurtis: Couldn't find `getColor`, thanks

boringcactus (approved):

PR #732, opened 2025-02-10
WIP

KaylaBrady:
### Summary
No ticket

What is this PR for?
Based on discussion [in slack](https://mbta.slack.com/archives/C062QNAJZ2M/p1739216016282209) to always do android debug builds as part of PRs instead of release builds.

iOS

android

### Testing

What testing have you done?


.github/workflows/android-staging-deploy.yaml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Build Android app
+        env:
+          FIREBASE_KEY: ${{ secrets.FIREBASE_KEY }}
+          GOOGLE_APP_ID_ANDROID: ${{ secrets.GOOGLE_APP_ID_ANDROID_STAGING }}
+          KEYSTORE_FILE: "${{ github.workspace }}/androidApp/upload-keystore.jks"
+          KEYSTORE_PASSWORD: ${{ env.MOBILE_APP_ANDROID_UPLOAD_KEY_PASSPHRASE }}
+          KEY_ALIAS: "upload"
+          KEY_PASSWORD: ${{ env.MOBILE_APP_ANDROID_UPLOAD_KEY_PASSPHRASE }}
+          MAPBOX_SECRET_TOKEN: ${{ secrets.MAPBOX_SECRET_TOKEN }}
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
```
EmmaSimon, line 146: You'll hit a conflict when you merge in #728, but you'll need to add [these lines](https://github.com/mbta/mobile_app/pull/728/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR150-R151) here.

fastlane/Fastfile
```diff
@@ -24,29 +24,34 @@ platform :android do
       else
         raise "Unknown flavor #{options[:flavor]}"
       end
+
     version_codes = google_play_track_version_codes(
       package_name: package_name,
       track: "internal",
     )
     last_version_code = version_codes.max
     next_version_code = last_version_code + 1
```
boringcactus, line 33: Fetching the existing version codes from Google Play requires GCP authentication, which won't be possible for PRs from external contributors. I think if the build type is debug we can just set `next_version_code = 1`.

.github/workflows/ci.yaml
```diff
@@ -123,68 +119,22 @@ jobs:
...
-          bundle exec fastlane android build flavor:Staging
+          bundle exec fastlane android build flavor:Staging build_type:Debug
       - uses: actions/upload-artifact@v4
         with:
           name: android-apk
-          path: androidApp/build/outputs/apk/staging/release/androidApp-staging-release.apk
+          path: androidApp/build/outputs/apk/staging/debug/androidApp-staging-debug.apk
       - uses: actions/upload-artifact@v4
         with:
           name: android-aab
```
boringcactus, line 138: Uploading the aab is only useful for deploying, and uploading the apk is only useful for giving our non-Google-Play-authenticated beta testers a build to sideload. Neither of these is a consideration in these debug builds, so we should probably skip the uploads.

.github/workflows/ci.yaml
```diff
@@ -123,68 +119,22 @@ jobs:
           ruby: true
           gcp-provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
           gcp-service-account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
```
boringcactus, line 121: We should probably skip GCP auth in this workflow.

.github/workflows/android-staging-deploy.yaml
```diff
@@ -0,0 +1,185 @@
...
+          SENTRY_DSN_ANDROID: ${{ secrets.SENTRY_DSN_ANDROID }}
+          SENTRY_ENVIRONMENT: "staging"
+        run: |
+          bundle exec fastlane android build flavor:Staging build_type:Release
```
boringcactus, line 150: The Android prod deployment workflow is also going to need `build_type:Release`.

.github/workflows/android-staging-deploy.yaml
```diff
@@ -0,0 +1,185 @@
...
+          path: androidApp/build/outputs/apk/staging/release/androidApp-staging-release.apk
+      - uses: actions/upload-artifact@v4
+        with:
+          name: android-aab
+          path: androidApp/build/outputs/bundle/stagingRelease/androidApp-staging-release.aab
+  deploy-android:
+    name: Upload to Google Play
+    concurrency:
+      group: deploy-android
+      cancel-in-progress: false
```
boringcactus, line 163: There is an existing bug that we may as well fix while we're in here: this concurrency group is the same as the one used in prod deploys, meaning an in-progress staging deploy will block a prod deploy until it finishes, despite that not being necessary. Setting this to `deploy-android-staging` or removing the `jobs.deploy-android.concurrency` block from the prod deployment workflow would fix that; I think the latter might be slightly better since prod deploys already have a workflow-level concurrency block so the job-level concurrency block is redundant at best.

fastlane/Fastfile
```diff
@@ -24,29 +24,34 @@ platform :android do
...
-    properties = {
+    properties = if options[:build_type] == "Debug"
+        {"android.injected.version.code" => next_version_code}
+    else
+        {
       "android.injected.signing.store.file" => ENV["KEYSTORE_FILE"],
       "android.injected.signing.store.password" => ENV["KEYSTORE_PASSWORD"],
```
boringcactus, line 39: The entries in this hash literal should continue to be indented further than the curly braces surrounding them.

.github/workflows/android-staging-deploy.yaml
```diff
@@ -0,0 +1,185 @@
+name: Android staging deploy
```
KaylaBrady, line 1: It seemed a bit cleaner to me to pull out a separate workflow for staging deploy rather than add more checks for `github.event_name == 'pull_request'`. Some of these common pieces should probably be pulled out into a composable workflow to reduce duplication.

.github/workflows/ci.yaml
boringcactus: Deleting the lines will have the same effect, since the default is an empty string, but if you'd rather explicitly pass `''` that's fine.

boringcactus (approved):

PR #733, opened 2025-02-11
feat: show track change alerts

boringcactus:
### Summary

_Ticket:_ [CR information | Track Change Alerts](https://app.asana.com/0/1205732265579288/1209129050348919)

I probably could have done less tangential refactoring here to keep the PR smaller, but I did not.

iOS

android

### Testing

Manually verified that track change alerts are correctly displayed on both Android and iOS, and that they are correctly hidden at core stations. Added unit tests for new functionality.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/pages/MorePageTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/pages/MorePageTests.kt
@@ -1,20 +1,16 @@
 package com.mbta.tid.mbta_app.android.pages
 
-import androidx.compose.ui.test.ComposeTimeoutException
 import androidx.compose.ui.test.ExperimentalTestApi
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
-import androidx.compose.ui.test.onRoot
 import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performScrollTo
-import androidx.compose.ui.test.printToLog
 import com.mbta.tid.mbta_app.repositories.ISettingsRepository
 import com.mbta.tid.mbta_app.repositories.MockSettingsRepository
 import com.mbta.tid.mbta_app.repositories.Settings
 import kotlin.test.assertTrue
-import kotlinx.coroutines.test.runTest
 import org.junit.Rule
 import org.junit.Test
 import org.koin.compose.KoinContext
@@ -35,7 +31,7 @@ class MorePageTests : KoinTest {
 
     @OptIn(ExperimentalTestApi::class)
     @Test
-    fun testSettings() = runTest {
+    fun testSettings() {
         var hideMapToggleCalled = false
 
         val koinApplication = koinApplication {
@@ -57,17 +53,12 @@ class MorePageTests : KoinTest {
             KoinContext(koinApplication.koin) { MorePage(bottomBar = {}) }
         }
 
-        try {
-            composeTestRule.waitUntilExactlyOneExists(hasText("Settings"))
-        } catch (ex: ComposeTimeoutException) {
-            composeTestRule.onRoot().printToLog("ci-keep")
-            throw ex
-        }
+        composeTestRule.waitUntilExactlyOneExists(hasText("Settings"))
         composeTestRule.onNodeWithText("Settings").performScrollTo()
         composeTestRule.onNodeWithText("Settings").assertIsDisplayed()
         composeTestRule.onNodeWithText("Hide Maps").performClick()
 
-        composeTestRule.awaitIdle()
+        composeTestRule.waitForIdle()
         composeTestRule.waitUntil { hideMapToggleCalled }
 
         assertTrue { hideMapToggleCalled }
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -47,6 +47,7 @@ data class Alert(
             // service changes are always secondary
             Effect.ServiceChange -> AlertSignificance.Secondary
             Effect.ElevatorClosure -> AlertSignificance.Accessibility
+            Effect.TrackChange -> AlertSignificance.Minor
             else -> AlertSignificance.None
         }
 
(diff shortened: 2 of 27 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/pages/MorePageTests.kt
```diff
@@ -57,17 +53,12 @@ class MorePageTests : KoinTest {
             KoinContext(koinApplication.koin) { MorePage(bottomBar = {}) }
         }
 
-        try {
-            composeTestRule.waitUntilExactlyOneExists(hasText("Settings"))
-        } catch (ex: ComposeTimeoutException) {
-            composeTestRule.onRoot().printToLog("ci-keep")
-            throw ex
-        }
+        composeTestRule.waitUntilExactlyOneExists(hasText("Settings"))
```
boringcactus, line 56: This test has been flaky for months now, I think, and I added this debugging as part of a prior PR to figure out what was actually going on. The output showed me that the ViewModel was never loading, and I figured out that it was because `runTest` makes all coroutines run in the main thread regardless of dispatcher (for more predictable event sequencing during tests, I think) but `waitUntilExactlyOneExists` blocks instead of suspending, meaning if the ViewModel didn't load before `waitUntilExactlyOneExists` was called, it'd never get the chance to.

Using `runBlocking` instead of `runTest` would probably have been sufficient to let the ViewModel load in the background while waitUntilExactlyOneExists is blocking, but I think it's a little bit clearer to standardize on the blocking functions and avoid `runTest`/`runBlocking` entirely, since `runTest` certainly looks like it's what should be used for tests.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -205,36 +285,38 @@ data class Alert(
...
-                        } && it.activities.contains(Alert.InformedEntity.Activity.Board)
+                    alert.anyInformedEntitySatisfies {
+                        checkActivity(InformedEntity.Activity.Board)
+                        checkDirection(directionId)
+                        checkRouteIn(routeIds)
+                        if (stopIds != null) {
+                            checkStopIn(stopIds)
+                        }
+                        checkTrip(tripId)
                     }
```
boringcactus, line 320: It's possible that the `InformedEntity.PredicateBuilder` is too complicated, but I think this code is both clearer and more efficient (comparing the direction O(N²) times is a waste), and this isn't even why I introduced it.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -278,15 +360,24 @@ data class Alert(
...
+                        it.anyInformedEntitySatisfies {
+                            checkActivityIn(
+                                InformedEntity.Activity.Exit,
+                                InformedEntity.Activity.Ride
+                            )
+                            checkDirection(trip.directionId)
+                            checkRoute(trip.routeId)
+                            checkStopIn(targetStopWithChildren)
+                        }
+                    }
```
boringcactus, line 380: The idea with the `PredicateBuilder` was to get the clarity of `applicableAlerts` with enough flexibility to require different activities instead in the context of downstream alerts.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/AlertSignificance.kt
```diff
@@ -1,19 +1,22 @@
...
 
-/**
- * | [AlertSignificance]     | [Major] | [Secondary] | [Accessibility] | [Minor] | [None] |
- * |-------------------------|---------|-------------|-----------------|---------|--------|
- * | Shown on map            | Yes     | No          | No              | No      | No     |
- * | Replaces upcoming trips | Yes     | No          | No              | No      | No     |
- * | Shown in nearby transit | Yes     | Yes         | Yes             | No      | No     |
- * | Shown in stop details   | Yes     | Yes         | Yes             | Yes     | No     |
- * | Shown in trip details   | Yes     | No          | No              | No      | No     |
- */
```
boringcactus, line 11: I don't think this table is all that clear - trip details is halfway to not existing anymore - so I figured now was a good time to swap in some more specific documentation on each value separately.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/ExtrinsicProperties.kt
boringcactus, line 1: I think what lines being grouped and CR stations being core have in common is that they could conceivably be defined in the data itself, but they are too presentation-specific for that to have been done. Maybe that's a stretch that doesn't justify moving line grouping in here, though.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
```diff
@@ -49,11 +49,16 @@ data class PatternsByStop(
...
+                                        alerts.toList(),
+                                        null,
+                                        listOf(it.route.id),
+                                        it.stopIds,
+                                        null
+                                    )
+                                    .filterNot {
+                                        it.effect == Alert.Effect.TrackChange &&
+                                            staticData.stop.isCRCore
+                                    },
```
boringcactus, line 61: I was struggling to figure out where to discard track changes at CR core stations, and I'm not sure this is the best place.

JackVCurtis: Yeah, semantically within the `applicableAlerts` function makes more sense to me, maybe just pass that function a boolean that determines whether to filter the track change alerts?

JackVCurtis (approved):
Seems like a pretty reasonable amount of refactor tbh

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/ExtrinsicProperties.kt
JackVCurtis: The thing that stands out to me about these properties is that they rely on hardcoded values, which is just another way of coming around to "this could be in the data". I think as a practical matter having the magic IDs isolated is good.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/PatternsByStop.kt
boringcactus: I think that `applicableAlerts` only looking at the informed entities and not considering the effect makes a lot of intuitive sense, and moving this in there would make that a little less coherent. I think moving it up into `withRealtimeInfo` makes more sense, given how much other filtering we're also doing at that level.

PR #734, opened 2025-02-11
feat(Android): Single direction destination label

EmmaSimon:
### Summary

_Ticket:_ No ticket

Noticed this was missing while working on commuter rail track numbers.

Screenshot_20250211_152816 Screenshot_20250211_152436 Screenshot_20250211_152553

android

### Testing

Added a unit test for the single direction case


KaylaBrady (approved):
Nice find & fix!

PR #738, opened 2025-02-12
feat: Add CR track numbers

EmmaSimon:
### Summary

_Ticket:_ [CR information | Preserve predicted platform](https://app.asana.com/0/1205732265579288/1209128916172062)

Display predicted track numbers at core CR stations when they exist, on both iOS and Android.

Screenshot_20250212_145655 Screenshot_20250212_145914

iOS

android

### Testing

Added unit tests, updated existing tests


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripHeaderCard.kt
```diff
@@ -309,17 +309,14 @@ private fun VehicleStatusDescription(
 private fun vehicleStatusString(
     context: Context,
     vehicleStatus: Vehicle.CurrentStatus,
-    stopEntry: TripDetailsStopList.Entry?
+    atTerminal: Boolean
```
KaylaBrady, line 312: 👍 Nice, this is clearer!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStopRow.kt
```diff
@@ -83,21 +85,44 @@ fun TripStopRow(
...
+                                    MaterialTheme.typography.bodyMedium.copy(
+                                        fontWeight =
+                                            if (targeted) FontWeight.Bold else FontWeight.Normal
+                                    ),
+                            )
+                            val trackNumber = stop.trackNumber
+                            if (
+                                trackNumber != null &&
+                                    routeAccents.type == RouteType.COMMUTER_RAIL &&
+                                    stop.stop.isCRCore
```
KaylaBrady, line 116: suggestion (non-blocking): This is used in a few places, might be nice to have a `shouldShowTrackNumber` helper! Could the vehicle_type of the stop be used? It is nullable, but should be [populated for all platform stops](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#stopstxt)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -16,9 +16,13 @@ constructor(val tripId: String, val stops: List<Entry>, val startTerminalEntry:
         val alert: Alert?,
         val schedule: Schedule?,
         val prediction: Prediction?,
+        val predictionStop: Stop?,
```
KaylaBrady, line 19: question: Do we only need the track number from the prediction stop? If so, it might be nice to pass that around directly instead of the stop since it is a little confusing looking only at this class why stop would be different from predictionStop. Alternatively, or maybe a single `PredictionWithStop` or `PredictionWithTrackNumber` field that packages them together would help to simplify since you can't have a predictionStop without a prediction.

Same thought applies to `UpcomingTrip`

KaylaBrady (commented):
Nice! Wasn't able to find example of this while running locally, but logic & screenshots look good! Just one structural question.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStopRow.kt
EmmaSimon: After looking closer at that test failure, I realized that it's caused by the test always looking at the combined semantic text, and the track number isn't in there, so I started some updates to handle that, including this change. I was passing in a route type, but I like the idea of using the one in the stop, I'll try that.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
EmmaSimon: If we start using the stop's vehicle type, then that means we need the whole stop object. I decided to keep the whole thing for cases like this, it seemed like it would be better to have as much context as possible for as long as possible. But I agree that it's confusing why it exists, I could just leave a comment? I'm a little worried about handling predictions when there is no associated stop, even though in theory that shouldn't be possible, I'm skeptical we'd never hit a case where it does happen.

KaylaBrady: Yeah I think a comment would be fine! I think the only way for there to be no associated stop is if something is messed up in the global data since stop is a required part of a prediction, which is unlikely but I agree not impossible. 

I forgot you can't extend a data class in kotlin so something like `PredictionWithStop` would be a bit ugly: 
```
prediction: Prediction,
stop: Stop?
```

I don't hate it, but also don't think it is nice enough to warrant making a change instead of adding a comment.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStopRow.kt
EmmaSimon: 😮‍💨 Ok well, that took much longer than I'd hoped. I moved the check to see if a track number should be displayed to the stop object, and then switched to just having the presence of the track number on the stop list entry determine whether or not it should be displayed. Also refactored some of the accessibility label stuff a little bit.

KaylaBrady (approved):
Awesome!

PR #740, opened 2025-02-13
fix: select vehicle in auto filter if assigned late

boringcactus:
### Summary

_Ticket:_ none
[Slack thread](https://mbta.slack.com/archives/C062QNAJZ2M/p1739388088874429)

Both Android and iOS go back to loading for a frame or two, but since it happens on both platforms and state management is a different kind of confusing on iOS than it is on Android I'm disinclined to chase it down any further.

I'm also adding in an Android DebugView change to inherently only render if debug mode is enabled, which makes it substantially easier to use from places that aren't already loading settings. If that's going to be confusingly subtle, I can take it back out, though.

iOS

android

### Testing

Manually verified that this change works for late-assigned vehicles on iOS and Android. Added a unit test for the central logic.


EmmaSimon (approved):
Nice job tracking this down!

PR #743, opened 2025-02-13
fix: tweak outside service area behavior

boringcactus:
### Summary

_Ticket:_ [🤖 | Outside service area behavior slightly off](https://app.asana.com/0/1205732265579288/1209310342616041)

Fixes the text color, restores the crosshair when pressing "View transit near Boston", and resets data to loading on both iOS and Android (since that wasn't actually working on iOS before).

iOS

android

### Testing

Manually verified that the appearance and behavior on Android is correct.


iosApp/iosApp/ContentView.swift
```diff
@@ -125,10 +125,13 @@ struct ContentView: View {
                     noNearbyStops: { NoNearbyStopsView(
                         hideMaps: contentVM.hideMaps,
                         onOpenSearch: { searchObserver.isFocused = true },
-                        onPanToDefaultCenter: { viewportProvider.animateTo(
-                            coordinates: ViewportProvider.Defaults.center,
-                            zoom: 13.75
-                        ) }
+                        onPanToDefaultCenter: {
```
KaylaBrady, line 128: 👍 nice catch on iOS consistency!

KaylaBrady (approved):

PR #744, opened 2025-02-14
feat: use route color for major alert icons in UpcomingTripView

boringcactus:
### Summary

_Ticket:_ [Disruptions UI updates in Nearby Transit and unfiltered Stop Details](https://app.asana.com/0/1205732265579288/1208091797262212) & [Fix handling of Snow Route alerts](https://app.asana.com/0/1205732265579288/1209342902372065)

Using the same solution for major alert icons that we used for secondary alert icons seems like a really good approach to me.

iOS

android

### Testing

Manually verified that disruption icons appear with the correct color. Added unit tests to verify this behavior on Android and iOS.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
@@ -1,9 +1,57 @@
 package com.mbta.tid.mbta_app.android
 
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toPixelMap
 import androidx.compose.ui.semantics.SemanticsActions
 import androidx.compose.ui.test.SemanticsMatcher
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.captureToImage
+import kotlin.math.abs
+import org.junit.Assert.fail
 
 fun hasClickActionLabel(expected: String?) =
     SemanticsMatcher("has click action label $expected") { node ->
         node.config[SemanticsActions.OnClick].label == expected
     }
+
+@OptIn(ExperimentalStdlibApi::class)
+private fun colorToHex(color: Color): String {
+    fun valToHex(value: Float): String {
+        val byteValue = (value * 255).toInt().toByte()
+        return byteValue.toHexString(HexFormat.UpperCase)
+    }
+    return "#${valToHex(color.red)}${valToHex(color.green)}${valToHex(color.blue)}"
+}
+
+private fun Color.isGrey(): Boolean =
+    abs(this.red - this.green) < 0.01 && abs(this.green - this.blue) < 0.01
+
+/**
+ * It's difficult to determine if a view hierarchy contains the correct image when the image is
+ * semantically invisible, but it's easy to capture a screenshot of the view and check it pixel by
+ * pixel.
+ */
+fun SemanticsNodeInteraction.assertHasColor(targetColor: Color) {
+    val pixelMap = this.captureToImage().toPixelMap()
+    val pixelCounts = mutableMapOf<Color, Int>()
+    for (x in 0.rangeUntil(pixelMap.width)) {
+        for (y in 0.rangeUntil(pixelMap.height)) {
+            val colorHere = pixelMap[x, y]
+            if (colorHere == targetColor) {
+                return
+            }
+            pixelCounts[colorHere] = pixelCounts.getOrDefault(colorHere, 0) + 1
+        }
+    }
+    // never returned so did not find the target
+    val legibleResult =
+        pixelCounts.entries
+            .groupBy(
+                { if (it.key.isGrey()) "grey" else "not grey" },
+                { colorToHex(it.key) to it.value }
+            )
+            .mapValues { it.value.sortedByDescending { it.second } }
+            .toSortedMap(reverseOrder())
+
+    fail("Did not contain specified color, but did have $legibleResult")
+}
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
@@ -43,7 +43,9 @@ fun SemanticsNodeInteraction.assertHasColor(targetColor: Color) {
             pixelCounts[colorHere] = pixelCounts.getOrDefault(colorHere, 0) + 1
         }
     }
-    // never returned so did not find the target
+    // never returned so did not find the target. hopefully the color that is present instead of the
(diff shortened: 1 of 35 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
```diff
@@ -1,9 +1,57 @@
...
+private fun Color.isGrey(): Boolean =
+    abs(this.red - this.green) < 0.01 && abs(this.green - this.blue) < 0.01
+
+/**
+ * It's difficult to determine if a view hierarchy contains the correct image when the image is
+ * semantically invisible, but it's easy to capture a screenshot of the view and check it pixel by
+ * pixel.
+ */
+fun SemanticsNodeInteraction.assertHasColor(targetColor: Color) {
+    val pixelMap = this.captureToImage().toPixelMap()
```
boringcactus, line 35: I knew there was a chance this existed. I may have been too eager to use it, but I'm tired of setting tags on things.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
```diff
@@ -1,9 +1,57 @@
...
+    val legibleResult =
+        pixelCounts.entries
+            .groupBy(
+                { if (it.key.isGrey()) "grey" else "not grey" },
+                { colorToHex(it.key) to it.value }
+            )
+            .mapValues { it.value.sortedByDescending { it.second } }
+            .toSortedMap(reverseOrder())
+
+    fail("Did not contain specified color, but did have $legibleResult")
```
boringcactus, line 56: When one of these tests fails, it's going to be confusing why it failed, so I'm reporting the most common non-grey color at the front of the exception as a hint (if I run the test with MapStopRoute.RED instead, I get the red hex code, which would hopefully be enough to debug if it was a real confusing error).

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -13,7 +13,7 @@ constructor(val tripId: String, val stops: List<Entry>, val startTerminalEntry:
     data class Entry(
         val stop: Stop,
         val stopSequence: Int,
-        val alert: Alert?,
+        val disruption: RealtimePatterns.Format.Disruption?,
```
boringcactus, line 16: This data class is too big already, so I didn't want to add another field to it; since the `Disruption` now contains both the `alert` and the `iconName`, and since we're calling the same `UpcomingTripView` widget that we'd call from nearby transit, I figured I'd just reuse `RealtimePatterns.Format.Disruption` directly.

EmmaSimon (approved):
Just a few minor clarifications/nitpicks, but overall looks good.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
boringcactus: The problem is that sorting colors by frequency will give you all the backgrounds before it gives you the colors that are actually useful in understanding why the test is failing. The idea is to partition the colors into non-grey and grey and then sort each of those by frequency, because in cases like the one I'm using this for right now, the most common non-grey color is supposed to be the one that's being searched for. The `mapValues` sorts the colors by frequency, and the `toSortedMap(reverseOrder())` puts `"non-grey"` before `"grey"`. (I saw `toSortedMap` in my autocomplete and it seemed like a reasonable choice, but if it's too magic I can just do `.entries.sortedDescending()`.)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -398,33 +400,40 @@ constructor(val tripId: String, val stops: List<Entry>, val startTerminalEntry:
             }
         }
 
-        private fun getAlert(
+        private fun getDisruption(
             entry: WorkingEntry,
             alertsData: AlertsStreamDataResponse?,
             globalData: GlobalResponse,
+            route: Route?,
```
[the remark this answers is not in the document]
boringcactus: I may have forgotten to read this function while I was changing it, oops.

The only edge case I can think of is that if you have an added trip (so the entries have no schedules) and there's an alerting stop that the vehicle has already passed (so the entry has no prediction), we can't look up a route ID from the individual entry, but we still can from the entire set of predictions. We don't show stops that the vehicle has already passed, though, so it is extremely unlikely to actually matter.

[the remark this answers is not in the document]
boringcactus: Actually, though, I think in principle it's better to resolve the route once for the whole list rather than at each entry separately.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
EmmaSimon: I don't love relying on "non-grey" alphabetically as the sort mechanism, but I don't necessarily have a more straightforward alternative, so I don't have an issue with keeping it. I assume the reason you're not just filtering out all the greys is for cases where there are only greys?

boringcactus: I thought about explicitly mapping grey and non-grey to 2 and 1 so that it was more obvious what was happening; that might be more clear than reverse alphabetical order.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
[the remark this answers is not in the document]
EmmaSimon: Since the route can be different from the entryRoute, it might make sense to only pass in the `MapStopRoute` so there's less chance for confusion.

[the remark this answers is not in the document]
EmmaSimon: Oh, sorry, I was still seeing the original changeset when I sent that, and was confused, these are always guaranteed to be the same route because they're downstream stops, I was thinking it was the list of departures at a stop, which could mix multiple GL routes.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
boringcactus: Swapped that in and expanded the comment so hopefully it's clearer if someone comes back to this code in a few months.

EmmaSimon (approved):

PR #747, opened 2025-02-14
feat: overhaul alert text handling

boringcactus:
### Summary

_Ticket:_ [Disruptions UI updates in Nearby Transit and unfiltered Stop Details](https://app.asana.com/0/1205732265579288/1208091797262212) & [Fix handling of Snow Route alerts](https://app.asana.com/0/1205732265579288/1209342902372065)

I could probably have just skipped these temporary translations on the grounds that most of them will not be seen before we get the next real translations back. Oops.

iOS
- [x] If you added any user-facing strings on iOS, are they included in Localizable.xcstrings?
  - [x] Add temporary machine translations, marked "Needs Review"

android
- [x] All user-facing strings added to strings resource in alphabetical order
- [x] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible

### Testing

Updated existing unit tests when values have changed. Checked in the Xcode/Android Studio preview that the effects specified in Figma have correct text and (thanks to #744) correct images.

PR #748, opened 2025-02-14
fix: tweak disruption styles to match designs

boringcactus:
### Summary

_Ticket:_ [Disruptions UI updates in Nearby Transit and unfiltered Stop Details](https://app.asana.com/0/1205732265579288/1208091797262212)

Drops the chevron for alerting departure rows, tweaks text color on Android, and fine tunes spacing on both iOS and Android to match the designs.

Stacked on #747.

|Android|iOS|
|-------|---|
|Screenshot_20250214_151420|IMG_0047|

iOS

android

### Testing

Manually verified new styles in IDE previews and full app. Added iOS unit test for elided chevron (this would be tougher on Android and I'm not sure checking for the color would be sufficient).

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/PredictionRowView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/PredictionRowView.kt
@@ -22,11 +22,15 @@ import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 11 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/PredictionRowView.kt
```diff
@@ -153,74 +159,85 @@ private fun PredictionRowViewPreview() {
         }
     val trip = objects.trip()
 
-    Column {
-        PredictionRowView(
-            predictions =
+    MyApplicationTheme {
```
KaylaBrady, line 162: 👍 thanks for adding the theme to previews!

PR #749, opened 2025-02-14
fix: improve font consistency between platforms

boringcactus:
### Summary

_Ticket:_ [🤖 | Fix incorrect text sizes](https://app.asana.com/0/1205732265579288/1209369563910304)

Who says we need to adhere to Android's definition of which font names exist?

Where iOS and Android diverge, I have for the most part just assumed that iOS is correct, and only checked in Figma a handful of things that I found uniquely suspicious.

iOS

android

### Testing

Went through the app and confirmed that everything still looks reasonable.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStopRow.kt
```diff
@@ -103,11 +104,8 @@ fun TripStopRow(
                                     }
                                     .placeholderIfLoading(),
                                 color = colorResource(R.color.text),
-                                style =
-                                    MaterialTheme.typography.bodyMedium.copy(
-                                        fontWeight =
-                                            if (targeted) FontWeight.Bold else FontWeight.Normal
-                                    ),
+                                fontWeight = if (targeted) FontWeight.Bold else FontWeight.Normal,
+                                style = Typography.body,
```
EmmaSimon, line 108: It might be slightly cleaner to just do the `if (targeted)` on `style = bodySemibold / body`. Or was it specced to actually be bold rather than semibold?

EmmaSimon (approved):
This is great, I appreciate the consistency 🙏

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStopRow.kt
boringcactus: It is specified as bold, but specifically it's specified as headline bold, which is coincidentally the same size as body.

PR #753, opened 2025-02-18
Android recenter on vehicle

JackVCurtis:
### Summary

_Ticket:_ [🤖 | stop + trip details | Recenter on vehicle button](https://app.asana.com/0/1205732265579288/1209118824865491)

What is this PR for?
Adds the button to show the overview of your selected vehicle/trip

iOS

android

### Testing

What testing have you done?

Automated testing for reproducible map states, manual testing for the states we can't currently create in the test environment (going off existing test coverage).

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
@@ -249,7 +249,7 @@ class HomeMapViewTests {
         viewModel.loadConfig()
         composeTestRule
             .onNodeWithContentDescription("Recenter map on my location")
-            .assertIsDisplayed()
+            .assertDoesNotExist()
     }
 
     @Test
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
@@ -188,6 +188,71 @@ class HomeMapViewTests {
     }
 
     @Test
+    fun testOverviewNotShownWhenNoPermissionsStopDetails() {
+
+        val locationManager = MockLocationDataManager(Location("mock"))
+
+        locationManager.hasPermission = false
+
+        composeTestRule.setContent {
+            HomeMapView(
+                lastNearbyTransitLocation = null,
+                nearbyTransitSelectingLocationState = mutableStateOf(false),
+                locationDataManager = locationManager,
+                viewportProvider = ViewportProvider(MapViewportState()),
+                currentNavEntry = SheetRoutes.StopDetails("stopId", null, null),
+                handleStopNavigation = {},
+                vehiclesData = emptyList(),
+                stopDetailsDepartures = null,
+                viewModel =
+                    MapViewModel(ConfigUseCase(MockConfigRepository(), MockSentryRepository()), {}),
+                searchResultsViewModel =
+                    SearchResultsViewModel(
+                        MockAnalytics(),
+                        MockSearchResultRepository(),
+                        VisitHistoryUsecase(MockVisitHistoryRepository())
+                    )
+            )
+        }
+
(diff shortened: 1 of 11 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/TripCenterButton.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.semantics
+import com.mbta.tid.mbta_app.android.R
+import com.mbta.tid.mbta_app.android.component.routeIcon
+import com.mbta.tid.mbta_app.model.RouteType
+
+@Composable
+fun TripCenterButton(modifier: Modifier = Modifier, routeType: RouteType, onClick: () -> Unit) {
```
KaylaBrady, line 16: question(non-blocking): Why not use the existing RecenterButton like we do on iOS?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/SearchBarOverlay.kt
```diff
@@ -154,7 +153,7 @@ fun SearchBarOverlay(
                     onExpandedChange = {},
                 ) {
                     LazyColumn(
-                        modifier = Modifier.fillMaxSize().background(colorResource(R.color.fill1)),
+                        modifier = Modifier.fillMaxSize().background(colorResource(R.color.fill2)),
```
KaylaBrady, line 156: suggestion: This should stay fill1  based on [this design QA](https://www.figma.com/design/OVO5Z0eYUtnnRvbxsIRpO3/App-QA?node-id=1013-221&t=tPCPdpf79064Fjj7-0)

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
```diff
@@ -145,13 +174,84 @@ class HomeMapViewTests {
...
+                        MockAnalytics(),
+                        MockSearchResultRepository(),
+                        VisitHistoryUsecase(MockVisitHistoryRepository())
+                    )
+            )
+        }
+        viewModel.loadConfig()
+        composeTestRule
+            .onNodeWithContentDescription("Recenter map on my location")
+            .assertDoesNotExist()
```
KaylaBrady, line 252: question: Is there a new test case needed for where the new TripCenterButton does exist?

OK to continue using this accessibility text pending design input in slack

KaylaBrady (commented):
Nice! Just a few minor comments.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/TripCenterButton.kt
JackVCurtis: Mostly because it didn't make much sense to me why the existing recenter button was reused due to the copy not being clear. I also didn't want to crowd the HomeMapView component any more than it already is.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
JackVCurtis: I looked into it, but the existing tests don't provide any examples - for the recenter button there's only the negative cases, no tests for when it actually is displayed. In order to show either button we need to pan the map within the test or somehow trick the component into thinking that the map is not currently focused on the user's location/the vehicle and trip. I kinda figured the last dev gave up.

KaylaBrady: 👍 I think there _could_ be a way by passing a trip filter with `currentNavEntry` and some other magic combination of mocked data, but this does feel like the kind of thing that may just be easier to test once we have visual snaphot-based tests set up.

KaylaBrady (approved):

PR #756, opened 2025-02-18
feat(ios): Various accessibility fixes

KaylaBrady:
### Summary

_Ticket:_ [ | Fix NCAM flagged issues](https://app.asana.com/0/1205732265579288/1209402431191060)

What is this PR for?
This PR includes a number of improvements to the VoiceOver behavior. 

Note: Hiding the tab bar from VoiceOver in StopDetails is surprisingly challenging - I didn't find a solution in this PR, but added a note to [ | Fix: Extra tab bar when hide maps enabled](https://app.asana.com/0/1205732265579288/1209391330487797). I think addressing that ticket will resolve the voice over problem.

iOS

android

### Testing

What testing have you done?
* Ran locally


iosApp/iosApp/ViewModels/SearchViewModel.swift
```diff
@@ -177,4 +184,19 @@ class SearchViewModel: ObservableObject {
             routePills: routePills.removingDuplicates()
         )
     }
+
+    private func stopRouteContentDescription(isStation: Bool, route: Route) -> String {
+        if silverRoutes.contains(route.id), isStation {
+            let routeName = "Silver Line"
```
JackVCurtis, line 190: I think the current convention is to hardcode strings like this in place but an 🎱 for the future might be to centralize our magic strings into an enum.

JackVCurtis:
Just one question about the odd test copy!

iosApp/iosApp/ViewModels/SearchViewModel.swift
KaylaBrady: Yeah good point. I could pull them into the strings.xml file and just add `tools:ignore="MissingTranslation"` so it is consistent with how we pull strings elsewhere

JackVCurtis (approved):
👍

PR #758, opened 2025-02-18
i18n: Add new translations

EmmaSimon:
### Summary

_Ticket:_ [Update translations](https://app.asana.com/0/1205732265579288/1209433346935170)

Imports the new translation files for iOS and Android. I needed to mark quite a few of the Android strings (mostly new statuses) as missing translations to get rid of warnings, these will be included in the next upcoming batch of translations.

Also adds `fr` as a new option for the feedback form URL. 

iOS

android


androidApp/src/main/res/values/strings.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <string name="access_issue">Access Issue</string>
-    <string name="access_issue_sentence_case">Access issue</string>
+    <string name="access_issue" tools:ignore="MissingTranslation">Access Issue</string>
+    <string name="access_issue_sentence_case" tools:ignore="MissingTranslation">Access issue</string>
```
KaylaBrady, line 4: @EmmaSimon I see there are placeholder translations for these ones missing translations on main. If the import removed them, is it possible to add them back with manual editing of the iOS strings file?

KaylaBrady (commented):
Looks good! Just one question about keeping some existing translation placeholders

androidApp/src/main/res/values/strings.xml
KaylaBrady: Ah, got it! I was worried there was an issue with importing overwriting things, glad that isn't the case.

KaylaBrady (approved):

PR #763, opened 2025-02-19
fix(Android): handle NNBSP in 12-hour times when testing on recent API levels

boringcactus:
### Summary

_Ticket:_ none

I think it's Android 14 (API 34) that introduced this change, but it's hard to tell.

iOS

android

### Testing

Ran the tests on a recent API level AVD and confirmed that they now pass instead of failing.


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
```diff
@@ -14,6 +16,18 @@ fun hasClickActionLabel(expected: String?) =
         node.config[SemanticsActions.OnClick].label == expected
     }
 
+fun hasTextMatching(regex: Regex): SemanticsMatcher {
```
KaylaBrady, line 19: appreciate the helper!

KaylaBrady (approved):

PR #764, opened 2025-02-19
fix(iOS): Double tab bar fixes

EmmaSimon:
### Summary

This fixes an extra tab bar showing up when hide maps is enabled, and also fixes some voiceover behavior where a tab bar is announced even though it isn't visible on the screen. I can't figure out a perfect fix though (see [this figma thread](https://www.figma.com/design/OVO5Z0eYUtnnRvbxsIRpO3?node-id=1193-35157#1140278167)), and though the tab bar is no longer included on the stop page, an extra tab bar is announced on the nearby transit page, and a tab bar is also announced in the search overlay, where no tab bar should be visible. The alternative is that the tab bar is hidden from VoiceOver, but the entire search bar and overlay is too.

iOS

### Testing

Manually tested with VoiceOver enabled


JackVCurtis (approved):
👍

PR #767, opened 2025-02-20
fix(iOS): Never hide background tab bar

EmmaSimon:
### Summary

Follow up removing the partial fix for the background tab bar showing up in VoiceOver when it isn't visible on the page, since hiding it was breaking the mapbox attribution positioning and the more page tab bar.

iOS

### Testing

Verified that the mapbox attributions are now positioned properly, but I could never reproduce the missing tab bar background on More.


JackVCurtis (approved):
👍

PR #754, opened 2025-02-18
center elevator accessibility button text, change font weight to 600 to match Figma

JackVCurtis:
### Summary

_Ticket:_ [🤖 | Fix elevator onboarding alignment -> onboarding punchlist](https://app.asana.com/0/1205732265579288/1209433125172607)

What is this PR for?

Small styling fixes, I can undo the FontWeight change if there's some reason we avoided it earlieer.

Screenshot_20250218_125526

iOS
- [x] If you added any user-facing strings on iOS, are they included in Localizable.xcstrings?
  - [x] Add temporary machine translations, marked "Needs Review"

android
- [x] All user-facing strings added to strings resource in alphabetical order
- [x] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible

### Testing

What testing have you done?

PR #770, opened 2025-02-21
fix(Android): hide onboarding images that are likely to collide with text

boringcactus:
### Summary

_Ticket:_ [🤖 | Fix dynamic text in onboarding ](https://app.asana.com/0/1201654106676769/1209447650055976)

Given the variety of screen sizes on Android, these thresholds may only be correct at the size Android Studio rendered the preview at while I was working on this, and a more robust solution would be to use the layout subsystem that I don't really understand to actually use the size of the text to show or hide the images, but in the interest of shipping in the short term, this seems like it's better than nothing, which is the main objective.

I'm also opportunistically fixing the bug where the button heights are fixed at 52 and so a second line of text will get cut off - on iOS that's a minimum height, and on Android it seems like it should be too.

iOS

android

### Testing

Checked in the Android Studio preview (which is the easiest place to apply arbitrary text scale values) that the images are hidden before the text clips into them.


JackVCurtis (approved):
👍

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
```diff
@@ -258,10 +264,7 @@ fun OnboardingScreenView(
                     }
                     Spacer(modifier = Modifier.weight(1f))
                     Button(
-                        modifier =
-                            Modifier.fillMaxWidth()
-                                .height(52.dp)
-                                .padding(start = 32.dp, end = 32.dp),
+                        modifier = buttonModifier.padding(start = 32.dp, end = 32.dp),
```
EmmaSimon, line 267: Do we still want the 32 padding on all the buttons if that was the thing causing the line breaks that messed up formatting? 16 seems reasonable and safer, but not a huge deal either way, these all need to just be their own composable, which we shouldn't spend time on now 🤷

EmmaSimon (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
boringcactus: I guess we don't; 16 inherent in the button modifier might be the better move there.

EmmaSimon: Yeah I considered that, but wasn't sure if it was ok to be adding back some of the padding around the Text that was removed in #754 for the elevator accessibility buttons (not sure why just those had the padding on the text rather than the button, maybe that was part of the problem).

boringcactus: Actually, no, the weird inconsistencies between these buttons are going to be annoying to disentangle here, and the right fix is a more comprehensive refactor anyway. The one that's most likely to wrap is "Show elevator closures", and that one's getting its padding removed in #754 anyway.

PR #772, opened 2025-02-21
fix(iOS): Hide tab bar properly when map is hidden

EmmaSimon:
### Summary

Follow up to #767, we do still want to hide the background tab bar sometimes.

iOS

### Testing

Manual testing that the tab bar only appears on more and nearby transit when maps are hidden or not, and that the mapbox attribution icons are in the correct place.


KaylaBrady (approved):

iosApp/iosApp/ContentView.swift
```diff
@@ -114,6 +120,7 @@ struct ContentView: View {
         } else {
             TabView(selection: $selectedTab) {
                 nearbyTab
+                    .toolbar(baseTabBarVisibility, for: .tabBar, .bottomBar)
```
boringcactus, line 123: This does not look like it's passing two values to `for:`, but I guess that's just how Swift handles varargs. Thanks, Apple.

EmmaSimon: Oh, thanks for mentioning that, it's not actually doing anything and I meant to remove it.

PR #777, opened 2025-02-21
feat(android.BottomNavbar): Use native tab bar for accessibility

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Fix NCAM flagged tab bar & direction switcher](https://app.asana.com/0/1205732265579288/1209433476007815)

What is this PR for?

This PR updates only the bottom tab bar, since that is the piece I'm most confident in changing. May follow up with making the direction picker a tab bar too depending on the timing.

iOS

android

### Testing

What testing have you done?
* Ran with Talk Back on, confirmed appropriate tab descriptions.
    * It seemed a little odd to me that the focus stayed on the selected tab when switching between pages, but that was the behavior before this change too so considering it out of scope to fix. 
* Checked it out in dark & light modes 
image
image


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/DirectionPicker.kt
```diff
@@ -69,22 +71,24 @@ fun DirectionPicker(
...
                     onClick = action,
-                    shape = RoundedCornerShape(6.dp),
-                    colors =
-                        ButtonDefaults.buttonColors(
-                            containerColor =
+                    modifier =
+                        Modifier.fillMaxHeight()
+                            .background(Color.fromHex(route.color))
+                            .background(deselectedBackgroundColor)
+                            .background(
```
KaylaBrady, line 81: Adding 3 background modifiers in a row feels particularly strange, but adding the route color background first was needed in order for the fill color to bleed through the slightly transparent grey deselectedBackgroundColor

JackVCurtis (approved):
🎉

PR #778, opened 2025-02-21
chore: bump version to 1.2.0 and release combined stop + trip details

boringcactus:
### Summary

_Ticket:_ none, I think

🎉 

I took the quickest and dirtiest conceivable approach to dummying out the feature flag, but I put in a bit of extra effort when fixing the end-to-end test.

iOS

android

### Testing

Verified that the version in the More screen is displayed as 1.2.0 in iOS and Android, and on iOS verified that most of the tests still pass locally (we should eventually fix all the tests on Xcode 16) and that the feature promo is displayed and combined stop + trip details is correctly used.


iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -321,12 +321,12 @@ extension HomeMapView {
         }
 
         if previousVehicle == nil || previousVehicle?.id != nextVehicle.id {
-            if nearbyVM.combinedStopAndTrip, let stop = nearbyVM.getTargetStop(global: globalData) {
+            if true, let stop = nearbyVM.getTargetStop(global: globalData) {
```
KaylaBrady, line 324: question (non-blocking): Are these `true` values placeholders for if we need to revert for some reason?

KaylaBrady (approved):

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
boringcactus: They're me trying to keep the diff minimal since we already plan to have a cleanup ticket later.

EmmaSimon: It might be worth adding a comment for each of these replacements, now when we do the cleanup ticket, we can no longer search for `combinedStopAndTrip` to see where it needs to be removed.

EmmaSimon: I think that would also allow all the failing tests to continue setting the flag to false through the constructor to keep them passing.

boringcactus: I've got fixes or at least patches that lead to the test passing for most of those failing tests, and I think it's worth making sure the functionality they're trying to test still works with the flag set to true.

boringcactus: The one I can't figure out is `HomeMapViewTests.testVehicleChanging`; @EmmaSimon as the original author of that test, do you think you can take a look at that one and either get it passing or decide it's not worth fixing?

EmmaSimon: Oh I see what's happening, the `viewportProvider.followedVehicle` behavior is only relevant to the legacy trip details, since we no longer follow vehicles at all, just update to the vehicle overview once when the vehicle first changes. So this test is no longer relevant for combined stop and trip details, I guess we should disable or remove it.

boringcactus: Makes sense. I can just delete it.

PR #779, opened 2025-02-21
fix(Android): fix some map loading performance

BrandonTR:
### Summary

_Ticket:_ [🤖 | Investigate: slow map data loading](https://app.asana.com/0/1205732265579288/1209367654923217)

Prevent some unnecessary transformation of globalMapData by passing it into the functions that require it, also ensure that route lines and stops are processed after globalMapData has been produced. This hopefully improves performance at least slightly and also makes stop rendering more reliable. I'm going to write up my findings during all of this but wanted to get this up in case it's useful for launch.

iOS

android

### Testing

What testing have you done?


JackVCurtis (approved):
This looks great!

EmmaSimon (approved):

PR #780, opened 2025-02-21
fix(Android): Info icon layout

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Fix scheduled to depart info icon](https://app.asana.com/0/1205732265579288/1209465916321750)

Workaround for an annoying and old [Android bug](https://issuetracker.google.com/issues/206039942).

Screenshot_20250221_155708 Screenshot_20250221_155607 Screenshot_20250221_155633

android

### Testing

Tested layout at different text sizes.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripHeaderCard.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripHeaderCard.kt
@@ -23,6 +23,8 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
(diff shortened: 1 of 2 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripHeaderCard.kt
```diff
@@ -23,6 +23,8 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.setValue
```
boringcactus, line 27: These probably aren't used.

boringcactus (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripHeaderCard.kt
EmmaSimon: I still don't understand why spotless only removes _some_ unused imports 😐

boringcactus: I think it's because `getValue` and `setValue` are extension operator functions that get used implicitly when you do `var ... by` and so it's harder to know statically that they aren't used.

PR #784, opened 2025-02-24
fix: always exclude shuttles from Green Line grouping

boringcactus:
### Summary

_Ticket:_ [Fix: Remove shuttles from Green Line groups](https://app.asana.com/0/1205425564113216/1208091797262204)

This was so easy that we probably shouldn't've waited this long to do it.

iOS

android

### Testing

Added a unit test. Verified by subtracting two days when loading schedules and when calculating nearby transit that this would have worked during the most recent shuttle instance.


EmmaSimon (approved):
🙌

PR #790, opened 2025-02-24
fix button toggling and add some extra padding around vehicle overview viewport

JackVCurtis:
### Summary

_Ticket:_ [Fix: stop details map buttons ](https://app.asana.com/0/1205732265579288/1209485723122809)

What is this PR for?

Fixes an issue where both recenter and trip center buttons would disappear when toggling between them. Also adds a 20px padding for when the selected vehicle is in the top right of the screen.

Screenshot_20250224_154858

iOS

android

### Testing

What testing have you done?
Manual, unfortunately hard to test map interactions currently

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
@@ -101,6 +101,7 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
 
     fun vehicleOverview(vehicle: Vehicle, stop: Stop?, density: Density) {
         isVehicleOverview = true
+        isFollowingPuck = false
         if (stop == null) {
             animateTo(vehicle.position.toMapbox())
         } else {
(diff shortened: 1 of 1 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
```diff
@@ -214,7 +215,7 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
             fun overviewPadding(density: Density) =
                 with(density) {
                     EdgeInsets(
-                        75.dp.toPx().toDouble(),
+                        95.dp.toPx().toDouble(),
                         50.dp.toPx().toDouble(),
                         75.dp.toPx().toDouble(),
```
boringcactus, line 220: It's probably worth keeping these symmetric so the center doesn't drift away from the center.

boringcactus (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
EmmaSimon: I think that would actually have the opposite effect, since the top padding doesn't include the status bar height by default, this pushes the center closer to the apparent center, and adding bottom padding would push it back up to off center.

boringcactus: Ah, yeah, that makes sense.

PR #555, opened 2024-11-22
ci: run iOS tests in GitHub Actions

boringcactus:
### Summary

_Ticket:_ [Spike: github actions CI replacement](https://app.asana.com/0/1205732265579288/1208710412256724/f)

What is this PR for?

- [ ] If you added any user facing strings on iOS, are they included in Localizable.xcstrings?

### Testing

What testing have you done?

PR #791, opened 2025-02-25
ci: run iOS tests in GitHub Actions

boringcactus:
### Summary

_Ticket:_ [Move iOS CI/CD to GitHub Actions](https://app.asana.com/0/1205732265579288/1209226816339255)

I think this piece works and could be merged as-is to replace our PR CI workflow in Xcode Cloud. It's been a while, though; it might not still work. Also, there's risk of drift between our GitHub Actions CI for PRs and our Xcode Cloud CD for staging and production deploys, and if something works from GitHub Actions but fails from Xcode Cloud then our misery will be ceaseless, so it might make sense to leave the Xcode Cloud CI running until we get deployments from GitHub Actions working properly.

iOS

android

### Testing

Manually verified that on my local machine `bundle exec fastlane ios test` works (I think the tests that fail are the tests that always fail in Xcode 16). Checked that in #555 at this stage the workflow was working.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -27,6 +27,9 @@ runs:
           android-test)
             tools="java kotlin-cache rsvg"
             ;;
+          ios-test)
+            tools="java kotlin-cache ruby xcode"
+            ;;
           android-build)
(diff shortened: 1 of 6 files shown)


BrandonTR (approved):
Nice!

PR #793, opened 2025-02-26
fix: keep showing BRD if vehicle is still present after prediction elapses

boringcactus:
### Summary

_Ticket:_ [Show BRD even if prediction is in past](https://app.asana.com/0/1205732265579288/1209072793746974)

There are a lot of cases that we want to not change here, so I think this has made `TripInstantDisplay.from` more complicated. Perhaps it'd be worthwhile to have more explicitly diverging code paths per route type at some point.

iOS

android

### Testing

Added unit test for new case.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
@@ -1,0 +1,16 @@
+package com.mbta.tid.mbta_app.model
+
+import kotlinx.datetime.Instant
+
+interface TripStopTime : Comparable<TripStopTime> {
+    val arrivalTime: Instant?
+    val departureTime: Instant?
+
+    val stopTime
+        get() = arrivalTime ?: departureTime
+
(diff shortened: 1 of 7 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
```diff
@@ -0,0 +1,16 @@
+package com.mbta.tid.mbta_app.model
+
+import kotlinx.datetime.Instant
+
+interface TripStopTime : Comparable<TripStopTime> {
+    val arrivalTime: Instant?
+    val departureTime: Instant?
```
boringcactus, line 7: This is an abstraction that I thought of months ago but haven't yet had a compelling reason to add. If I'd been more eager to implement it, I would still remember more places where we could take advantage of it; as it stands, there's only one existing piece of code that I've changed to use this abstraction.

KaylaBrady (approved):
Nice, and nice abstraction! To your comment about explicitly diverging code paths - I do think we're at a point where that would be valuable; as-is this is a long function and it is hard to quickly understand exactly how the order of the branches matters here. Could you make a ticket for it?

PR #794, opened 2025-02-27
Android location issues

JackVCurtis:
### Summary

_Ticket:_ [Fix: Android location issues](https://app.asana.com/0/1205732265579288/1209485724244302)

What is this PR for?

This PR fixes the issue in the second bullet of the linked ticket. The issue in the first bullet did not reproduce, but [Paul found another bug with the recenter button](https://mbta.slack.com/archives/C05QMG9GS9M/p1740672500855899?thread_ts=1740672061.625109&cid=C05QMG9GS9M) which this PR corrects.

iOS

android

### Testing

What testing have you done?

Manual testing since we don't have a good way of testing location behavior

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
@@ -14,7 +14,6 @@ import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.CameraState
 import com.mapbox.maps.EdgeInsets
 import com.mapbox.maps.extension.compose.animation.viewport.MapViewportState
-import com.mapbox.maps.extension.style.expressions.dsl.generated.zoom
 import com.mapbox.maps.plugin.animation.MapAnimationOptions
 import com.mapbox.maps.plugin.viewport.data.DefaultViewportTransitionOptions
 import com.mapbox.maps.plugin.viewport.data.FollowPuckViewportStateOptions
@@ -124,7 +123,7 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
         }
     }
 
-    fun stopOverview(stop: Stop) {
+    fun stopCenter(stop: Stop) {
         isFollowingPuck = false
         isVehicleOverview = false
         animateTo(stop.position.toMapbox())
@@ -204,7 +203,6 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
     }
 
     fun restoreNearbyTransitViewport() {
-        // TODO preserve zoom
         isFollowingPuck = savedNearbyTransitViewport?.isFollowingPuck ?: false
         isVehicleOverview = false
         isAnimating = true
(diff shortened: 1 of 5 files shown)


EmmaSimon (changes requested):
I think the location changing after backgrounding the app is working, but I'm still seeing the recenter button stay visible even after you go back from stop details.

I also noticed that there was previously a flicker of the vehicle position button underneath the recenter button when you navigated to stop details, but after this change, it flickers in first, then switches to the recenter button. Maybe it makes sense to hide both of them until the page is done loading?

| Before | After |
| -------- | ------- |
| [Screen_recording_20250227_153941.webm](https://github.com/user-attachments/assets/af827d44-4b81-4d29-82bf-95f893da3659) | [Screen_recording_20250227_154124.webm](https://github.com/user-attachments/assets/ee03fb4f-2f8c-40b3-9dd3-79224815d184) |

JackVCurtis:
Yeah, I think hiding both makes sense. I'm going to try to factor more of this boolean logic into the view model or something, it's getting hard to keep the state straight.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
```diff
@@ -192,14 +205,19 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
 
     fun restoreNearbyTransitViewport() {
         // TODO preserve zoom
```
EmmaSimon, line 207: Might as well get rid of this comment, I think we decided that we don't want to preserve zoom

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -324,7 +331,26 @@ fun HomeMapView(
...
                 LaunchedEffect(currentNavEntry) { updateDisplayedRoutesBasedOnStop() }
-
+                LaunchedEffect(
+                    locationDataManager.hasPermission,
+                    currentLocation,
+                    viewportProvider.isAnimating,
+                    viewportProvider.isFollowingPuck,
+                    selectedVehicle,
+                    searchResultsViewModel.expanded,
+                    viewportProvider.isVehicleOverview
```
EmmaSimon, line 341: nitpick: I'd maybe alphabetize these dependencies, just for readability

androidApp/src/main/java/com/mbta/tid/mbta_app/android/location/ViewportProvider.kt
```diff
@@ -120,14 +124,20 @@ class ViewportProvider(var viewport: MapViewportState, isManuallyCentering: Bool
         }
     }
 
+    fun stopOverview(stop: Stop) {
```
EmmaSimon, line 127: nitpick: This might make more sense as "stopCamera" or "stopCenter" or something, since the "Overview" in "vehicleOverview" is referring to a specific type of Mapbox viewport type, which this doesn't use.

EmmaSimon (approved):
Just a few nitpicks, but overall this looks great, nice work!

PR #796, opened 2025-02-28
fix(StopDetailsDepartures): Use same Predictions Unavailable logic as Nearby Transit

KaylaBrady:
### Summary

_Ticket:_ [Name](URL)

What is this PR for?

iOS

android

### Testing

What testing have you done?


JackVCurtis (approved):
👍 Seems like a good step

PR #798, opened 2025-02-28
refactor: move alert affected stops calculation into shared code

boringcactus:
### Summary

_Ticket:_ [Fix: Alert Details is slow/broken](https://app.asana.com/0/1205732265579288/1209485722332828)

The performance issue was only present on iOS, and a brief profiler session revealed that the app was spending 2.67s in a single frame bridging Kotlin Maps to Swift Dictionaries in the context of calculating affected stops:
Screenshot 2025-02-28 at 1 59 07 PM

We probably should have had this in shared code all along.

iOS

android

### Testing

Verified that if the 216-informed-entity Green Line alert is opened the stop list is not perceptibly delayed in loading and scrolling is smooth.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/GlobalResponse.kt
```diff
@@ -83,4 +84,15 @@ data class GlobalResponse(
             }
             .distinct()
     }
+
+    fun getAlertAffectedStops(alert: Alert?, routes: List<Route>?): List<Stop>? {
```
KaylaBrady, line 88: suggestion: I know this wasn't tested specifically in either platform, but would be nice to have a unit test or two for it now that it is pulled out!

KaylaBrady (approved):
Good find!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/GlobalResponse.kt
boringcactus: Added.

PR #804, opened 2025-03-04
feat: remove old TripHeadsigns feature flag & add new GroupByDirection flag

KaylaBrady:
### Summary

_Ticket:_ [Group by direction | Feature flags](https://app.asana.com/0/1205732265579288/1209536961544694)

What is this PR for?

Removes old TripHeadsigns feature flag & adds a new GroupByDirection flag. This leaves the old TripHeadsigns functions in place so that we can easily reference them while building up the group by direction functionality, but removes all entry points from the UI to those functions so that we can freely break them as needed.

Also removes the now unused `CombinedStopAndTrip` feature flag.

iOS

android

### Testing

What testing have you done?
* Ran iOS application, android application, confirmed both run.

### DIFF
--- a/androidApp/src/main/res/values/strings.xml
+++ b/androidApp/src/main/res/values/strings.xml
@@ -96,6 +96,7 @@
     <string name="fog">Fog</string>
     <string name="freight_train_interference">Freight Train Interference</string>
     <string name="full_description">Full Description</string>
+    <string name="group_by_direction" translatable="false">Group by Direction</string>
     <string name="hazmat_condition">Hazmat Condition</string>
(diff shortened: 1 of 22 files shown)


androidApp/src/main/res/values/strings.xml
```diff
@@ -96,6 +96,7 @@
     <string name="fog">Fog</string>
     <string name="freight_train_interference">Freight Train Interference</string>
     <string name="full_description">Full Description</string>
+    <string name="group_by_direction" translatable="false">Group by Direction</string>
```
KaylaBrady, line 99: Only exposed in staging, so skipping translations

EmmaSimon (approved):

PR #805, opened 2025-03-04
refactor: simplify Koin mock repo usage, especially on Android

boringcactus:
### Summary

_Ticket:_ [🤖 Simplify dependency injection in instrumented tests](https://app.asana.com/0/1205732265579288/1209010630021062)

I'm happy about this. Setting arbitrary properties on a `MockRepositories` instance feels a lot nicer than having to define a handful of factory method arguments (since if we provide all of them SKIE will generate 2^n variants), and passing the `ObjectCollectionBuilder` to set a bunch of repositories at once makes a lot of intuitive sense to me.

iOS

android

### Testing

Checked that all the Android tests still pass and that most of the iOS tests still pass (shout out to Xcode 16).


BrandonTR (approved):
🎉

PR #806, opened 2025-03-04
fix: Handle unknown Alert Cause and Effect values

EmmaSimon:
### Summary

_Ticket:_ [Update clients to handle new & unknown alert causes/effects](https://app.asana.com/0/1205732265579288/1209341739588168)

This includes the piece for safely parsing unknown enum values, using [coerceInputValues](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-builder/coerce-input-values.html), along with a new batch of alert causes from https://github.com/mbta/api/pull/856, and their associated machine translations.

If there is an unknown non-nullable case outside of the cause and effect handled here, we'll still get crashes. I looked into possibilities for parsing the list by handling each element individually and dropping any that weren't parseable, but there's no obvious good solution. There is an [open issue](https://github.com/Kotlin/kotlinx.serialization/issues/1205) for it, and a handful ([1](https://github.com/Kantis/ks3/pull/103/files), [2](https://github.com/illuzor/Kind-List-Serializer/blob/main/src/main/kotlin/com/illuzor/kindlistserializer/KindListSerializer.kt)) of existing implementations from random people, which we could adopt, but I'm not sure if we should, when using `coerceInputValues` gives us basically everything we need for the scope of this ticket, but a more lenient list serializer could protect us from future issues that we haven't predicted yet.

### Testing

Added a unit test for parsing unknown values, and manually checked that with an undefined cause coming from the backend, alert parsing did not crash.


androidApp/src/main/res/values/strings.xml
```diff
@@ -256,6 +263,7 @@
     <string name="track_work">Track Work</string>
     <string name="traffic">Traffic</string>
     <string name="train">train</string>
+    <string name="train_traffic">Train Traffic</string>9
```
boringcactus, line 266:
```suggestion
    <string name="train_traffic">Train Traffic</string>
```

boringcactus (approved):

PR #809, opened 2025-03-05
wip: draft of foundational data types with a bunch of placeholders

KaylaBrady:
### Summary

What is this PR for?

Currently a draft to seek feedback on this approach before I sink more time into actually getting all the pieces together. 

This introduces a new structure, `RouteCardData` as a replacement for the existing `NearbyStaticData`, `PatternsByStop`, etc. classes which have grown in complexity over time as we organically added functionality. These new structures aim to simplify the data representation with simplified naming and hiearchy. 

I've tried on using the builder pattern here. It felt like a good option since we have a complex series of steps; hopefully pulling all those steps into discrete functions will help with readability. 

The supporing `Builder` classes are strongly based in the `NearbyHierarchy` classes. I opted to make a new file instead of modifying the old pieces for simplicity. There are a few key differences though:
* NearbyHierarchy was intended to support adding headsigns to the list of headsigns at a stop for a station that might not otherwise be there. We don't need that functionality since we are no longer grouping by headsigns, which I think allows us to simplify some of the functionality.
    * NearbyHiearchy uses generics to achieve merging together hierarchies of different types. With the above simplification, I think there will be few enough steps where we can omit the use of generics and instead use a `LeafBuilder` type to represent the leaf data that we are incrementally building up at each step.
* NearbyHiearchy was an internal data type that was ultimately transformed into PatternsByStop. In this approach, I tried to keep the internal pieces limited to the supporting `Builder` classes and expose a top-level structure that we could use comfortably in views models + views.
    
Please leave any feedback you have at this stage - structural, naming, nitpicks, new ideas, all welcome! 

iOS

android

### Testing

What testing have you done?

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
@@ -1,0 +1,184 @@
+package com.mbta.tid.mbta_app.model
+
+import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
+import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
+import com.mbta.tid.mbta_app.model.response.ScheduleResponse
+import io.github.dellisd.spatialk.geojson.Position
+import kotlin.time.Duration
+import kotlinx.datetime.Instant
+
+// type aliases can't be nested :(
+
+typealias ByDirection = Map<Int, RouteCardData.Leaf>
+
+private typealias ByDirectionBuilder = Map<Int, RouteCardData.LeafBuilder>
+
+private typealias ByStopIdBuilder = Map<String, RouteCardData.RouteStopDataBuilder>
+
+private typealias ByLineOrRouteBuilder = Map<String, RouteCardData.Builder>
+
+/**
+ * Contain all data for presentation in a route card. A route card is a snapshot of service for a
+ * route at a set of stops. It has the general structure: Route (orLine) => Stop(s) => Direction =>
+ * Upcoming Trips / reason for absence of upcoming trips
+ */
+data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: List<RouteStopData>) {
+    data class RouteStopData(
+        val stop: Stop,
+        val directions: List<Direction>,
+        val data: ByDirection
+    )
+
+    data class Leaf(val upcomingTrips: List<UpcomingTrip>?, val alertsHere: List<Alert>?)
+
+    sealed interface LineOrRoute {
+        data class Line(
+            val line: com.mbta.tid.mbta_app.model.Line,
+            val routes: Set<com.mbta.tid.mbta_app.model.Route>
+        ) : LineOrRoute
+
+        data class Route(val route: com.mbta.tid.mbta_app.model.Route) : LineOrRoute
+    }
+
+    companion object {
+        fun routeCardsForStopList(
+            stopIds: List<String>,
+            globalData: GlobalResponse?,
+            sortByDistanceFrom: Position?,
+            schedules: ScheduleResponse?,
+            predictions: PredictionsStreamDataResponse?,
+            alerts: AlertsStreamDataResponse?,
+            filterAtTime: Instant,
+            showAllPatternsWhileLoading: Boolean,
+            hideNonTypicalPatternsBeyondNext: Duration?,
+            filterCancellations: Boolean,
+            includeMinorAlerts: Boolean,
+            pinnedRoutes: Set<String>
+        ): List<RouteCardData>? {
+
+            // if predictions or alerts are still loading, this is the loading state
+            if (predictions == null || alerts == null) return null
+
+            // if global data was still loading, there'd be no nearby data, and null handling is
+            // annoying
+            if (globalData == null) return null
+
+            val cutoffTime = hideNonTypicalPatternsBeyondNext?.let { filterAtTime + it }
+
+            return ListBuilder()
+                .addStaticStopsData(stopIds, globalData)
+                .addUpcomingTrips(schedules, predictions, filterAtTime)
+                .filterIrrelevantData(cutoffTime, showAllPatternsWhileLoading, filterCancellations)
+                .addAlerts(alerts, includeMinorAlerts, filterAtTime)
+                .build()
+                .sort(sortByDistanceFrom, pinnedRoutes)
+        }
+    }
+
+    class ListBuilder() {
+        val data: ByLineOrRouteBuilder = emptyMap()
+
+        fun addStaticStopsData(stopIds: List<String>, globalData: GlobalResponse?): ListBuilder {
+            // TODO
+            return this
+        }
+
+        fun addUpcomingTrips(
+            schedules: ScheduleResponse?,
+            predictions: PredictionsStreamDataResponse,
+            filterAtTime: Instant
+        ): ListBuilder {
+            // TODO build upcoming trips
+            // transform into map route => stop => direction => upcoming trips
+            // merge into data
+            return this
+        }
+
+        fun addAlerts(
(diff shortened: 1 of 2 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
```diff
@@ -0,0 +1,184 @@
...
+private typealias ByStopIdBuilder = Map<String, RouteCardData.RouteStopDataBuilder>
+
+private typealias ByLineOrRouteBuilder = Map<String, RouteCardData.Builder>
+
+/**
+ * Contain all data for presentation in a route card. A route card is a snapshot of service for a
+ * route at a set of stops. It has the general structure: Route (orLine) => Stop(s) => Direction =>
+ * Upcoming Trips / reason for absence of upcoming trips
+ */
+data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: List<RouteStopData>) {
```
KaylaBrady, line 26: We have used `Transit` in some places to mean `LineOrRoute`, could adopt that here also. TBH I always call the UI component RouteCard in my head when thinking about it and just keep the caveat in my head that it also is a line card b/c GL is weird. We could also consider calling it RouteCard and just document that sometimes we are treating lines like routes sometimes.

Very to other naming ideas!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
```diff
@@ -0,0 +1,184 @@
...
+        data class Line(
+            val line: com.mbta.tid.mbta_app.model.Line,
+            val routes: Set<com.mbta.tid.mbta_app.model.Route>
+        ) : LineOrRoute
+
+        data class Route(val route: com.mbta.tid.mbta_app.model.Route) : LineOrRoute
+    }
+
+    companion object {
+        fun routeCardsForStopList(
```
KaylaBrady, line 45: pulled props from `withRealtimeInfoViaTripHeadsigns`

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
```diff
@@ -0,0 +1,184 @@
...
+            globalData: GlobalResponse?,
+            sortByDistanceFrom: Position?,
+            schedules: ScheduleResponse?,
+            predictions: PredictionsStreamDataResponse?,
+            alerts: AlertsStreamDataResponse?,
+            filterAtTime: Instant,
+            showAllPatternsWhileLoading: Boolean,
+            hideNonTypicalPatternsBeyondNext: Duration?,
+            filterCancellations: Boolean,
+            includeMinorAlerts: Boolean,
```
EmmaSimon, line 56: I don't totally understand the utility of `includeMinorAlerts` being on the top level of creating `NearbyStaticData` in the existing code, is this still necessary?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
KaylaBrady: Yeah the existing pattern made sense to me while we were doing it, but I like the idea of moving towards a spec! We should def keep that in mind as we do the barebones UX - I don't have a sense yet of whether we'd have an opportunity to make that change then while introducing new components, or if we'll just want to make a separate refactor ticket for after these data types are fully adopted

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
KaylaBrady: It looks like we only set it to true for [stop details](https://github.com/mbta/mobile_app/blob/ce011a551655236d233d5aa7ffacf9250c5e1303/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsDepartures.kt#L193). I assume some of these other booleans are also driven by the page type - maybe we can condense them all down into a context variable like we use for `TripInstantDisplay` to simplify the number of params here

KaylaBrady: Just did a push with that change to pass 1 context var instead of 3 booleans.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
```diff
@@ -0,0 +1,195 @@
...
+        val directions: List<Direction>,
+        val data: ByDirection
+    )
+
+    data class Leaf(val upcomingTrips: List<UpcomingTrip>?, val alertsHere: List<Alert>?)
+
+    sealed interface LineOrRoute {
+        data class Line(
+            val line: com.mbta.tid.mbta_app.model.Line,
+            val routes: Set<com.mbta.tid.mbta_app.model.Route>
```
boringcactus, line 43: As far as I remember, `LineOrRoute.Line` containing the routes is what causes the Kenmore bug: the static data knows that only B, C, D go to Kenmore, but the realtime data doesn't know that, and it can't build a `LineOrRoute.Line` without a list of routes, so it assumes B, C, D, E which then fails to match the static data because the `LineOrRoute.Line`s are different.

In the final object, we'll probably still need the list of routes included in a line, but in the intermediate state it might work to use just the line as a key and then have a separate `Map<String, List<Route>>` for the routes-by-line attested by the static data at any given stop. It might be simpler to just count all GL branches as always included at every GL stop, though; I'm not sure if that'd actually break anything but if it wouldn't it'd be way simpler.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
boringcactus: What if we had a `RouteGroup` that contains a usually-single-element `routes: List<Route>` and a usually-null `line: Line?`

boringcactus:
I think builder pattern makes a lot of sense here to keep the state manageable, and I agree that it makes sense to try to subsume the entire old code path including `NearbyStaticData` and `RealtimePatterns` rather than only defining temporary structures for within the call to `withRealtimeInfo` - I think I was approaching trip headsigns as just the feature itself rather than as a refactor opportunity to enable the feature and pay down some tech debt.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
KaylaBrady: Ah thanks for that detail, I forgot the specifics of the Kenmore bug. Will think through if there are any further issues from always including all GL routes

KaylaBrady:
Thanks for the review @boringcactus!

>  I think I was approaching trip headsigns as just the feature itself rather than as a refactor opportunity to enable the feature and pay down some tech debt.

I 100% think this was the right approach at the time of trip headsigns! The larger refactor was definitely out of scope at that point w/ the release timeline. Glad we have some more time to revisit it now that the initial releases are behind us!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardHierarchy.kt
KaylaBrady: Oh interesting, so we don't have to do as much special casing? Though I guess then we'd probably end up doing conditional expressions in the UI depending on whether line is null or not, so I'd lean towards still preferring representing them as distinct data types. But I am open to this option - maybe whoever picks up the barebones UX ticket can use that as an opportunity to re-consider if this will make things easier to work with & switch if so, since it should be a fairly minor change.

KaylaBrady:
Closing this Draft PR to preserve initial conversation, will open a separate PR to cover the functionality being built out more.

PR #810, opened 2025-03-05
build: deploy for iOS from GitHub Actions

boringcactus:
### Summary

_Ticket:_ [Move iOS CI/CD to GitHub Actions](https://app.asana.com/0/1205732265579288/1209226816339255)

I can't wait.

iOS

android

### Testing

Staging deployments have been confirmed to be working from the GitHub side, although something weird may be happening - I see the build from GitHub in TestFlight at the top level, but I can't choose it or several of the other recent builds from the TestFlight version history.


.github/actions/setup/action.yml
```diff
@@ -100,6 +106,12 @@ runs:
         create_credentials_file: true
         workload_identity_provider: ${{ inputs.gcp-provider }}
         service_account: ${{ inputs.gcp-service-account }}
+    - name: Configure AWS Credentials
+      if: inputs.aws-role != ''
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: ${{ inputs.aws-role }}
+        aws-region: us-east-1
```
KaylaBrady, line 114: question (non-blocking): For my own understanding, why would we want to skip this if aws-role is empty? Is this conditional for the case where it is run from dependabot / external contributors?

.github/workflows/build-ios.yaml
```diff
@@ -0,0 +1,122 @@
+name: Build for iOS
+
+on:
+  workflow_call:
+    inputs:
+      deploy:
+        description: "Whether or not to deploy the app to the App Store"
+        required: true
+        type: boolean
```
KaylaBrady, line 9: question (non-blocking): Looks like this is only used in deploy flows as of this PR which makes sense. Appreciate it being controlled with an input var, but wondering if you envision this being used outside deploy flows?

fastlane/Fastfile
```diff
@@ -81,6 +76,68 @@ platform :android do
...
+      "com.mbta.tid.mbtaapp.staging"
+    else
+      raise "Unknown scheme #{scheme}"
+    end
+  end
+
+  def keychain_path()
+    # For reasons, KEYCHAIN_PATH is not the actual keychain path created by setup_ci,
+    # it's missing a "-db" suffix that macOS adds for some reason,
+    # and this helper will both add that suffix to a path and turn `"login"` into a full path of its own
```
KaylaBrady, line 93: oof, good find & workaround

KaylaBrady (commented):
Nice work piecing these tricky systems together! Happy to see this significant chunk of work merged & iterated on as needed.

To your comment about not being able to select versions in TestFlights, which builds would you expect to see as selectable that are not? Is 2435 one? Noticed that it shows 10 invites but 0 groups in app store connect, may be related to the internal test group error?

boringcactus:
2437 was the last build created by the GitHub Actions deployment process before I moved that back out of CI to get this PR ready for review.

.github/actions/setup/action.yml
boringcactus: We use the setup action for everything, including jobs that don't need AWS access and runs from external contributors that don't have secrets access at all, so this lets us skip trying to log in to AWS in both of those cases.

.github/workflows/build-ios.yaml
boringcactus: Seems like we don't envision this being used outside deploy flows, but keeping this argument around makes this workflow more similar to the build-android workflow, and it doesn't really add a meaningful amount of complexity, so I'm inclined to leave it in anyway.

KaylaBrady: Yeah I'm cool with leaving it in place!

KaylaBrady (approved):
Oops, meant to approve initially!

PR #816, opened 2025-03-10
feat(RouteCardData): Foundation of new data type to support grouping by direction

KaylaBrady:
### Summary

What is this PR for?

See https://github.com/mbta/mobile_app/pull/809 for some initial review of the new data structure approach. 

RouteCardData uses the builder pattern to construct the data representation that contains all data needed in a route card. 

This PR adds the foundational data types and implements the first piece of the builder logic: `addStaticStopsData`. This is largely a port of the logic from `NearbyStaticData` with some exceptions, which I'll call out in comments.

Intentionally excluded from this PR, and will be included in follow-ups:
* Directions
* Upcoming trips
* Filtering
* Sorting

iOS

android

### Testing

What testing have you done?
* Ported unit tests of NerabyStaticData


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -0,0 +1,349 @@
...
+            private set
+
+        /**
+         * Construct a map of the route/line-ids served by the given stops. Uses the order of the
+         * stops in the given list to determine the stop ids that will be included for each route.
+         *
+         * A stop is only included at a route if it serves any route pattern that is not served by
+         * an earlier stop in the list.
+         */
+        fun addStaticStopsData(stopIds: List<String>, globalData: GlobalResponse): ListBuilder {
```
KaylaBrady, line 119: This is mostly a port of [this NearbyStaticData constructor](https://github.com/mbta/mobile_app/blob/7a09634054172e6cb20144c9742dd98d64015d10/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NearbyStaticData.kt#L202), building the new data types instead of NearbyStaticData. I did modify the logic a bit to handle routes and lines in the same way rather than as separate steps.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -0,0 +1,349 @@
...
+                    )
+
+                    val stopKey =
+                        stop.parentStationId?.let { parentStationId ->
+                            fullStopIds
+                                .getOrPut(parentStationId) { mutableSetOf(parentStationId) }
+                                .add(stop.id)
+                            // Parents should be disjoint, but if somehow a parent has its own
+                            // patterns,
+                            // find it in the regular stops list
```
EmmaSimon, line 150: nitpick: This would fit on one line 😛

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -0,0 +1,349 @@
...
+                            .filterNot {
+                                routePatternsUsed.containsAll(
+                                    it.value.map { pattern -> pattern.id }.toSet()
+                                )
+                            }
+                    routePatternsUsed.addAll(
+                        patternsByRouteOrLine.flatMap { it.value }.map { it.id }
+                    )
+
+                    val stopKey =
```
EmmaSimon, line 143: nitpick: This might make more sense named `keyStop`/`keyedStop` or something, I was a little confused that it was being assigned an actual stop, because the name made me expect it to be a string or some distinct key type derived from a stop (like we have for `UpcomingTripKey`).

EmmaSimon (approved):
Looks a lot more straightforward than the old code!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
KaylaBrady: I definitely need to get better at paying attention to line breaks when copypasta-ing comments!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
KaylaBrady: Good call! I ended up reworking this a bit to separate defining the key from populating the map, and renamed the key to `stopOrParent` which is hopefully a bit more descriptive.

PR #825, opened 2025-03-12
feat(RouteCardData): Add upcoming trips & filter

KaylaBrady:
### Summary

What is this PR for?
This PR adds upcoming trips to the RouteCardData structure and filters using the same rules as `NearbyStaticData.withRealtimeInfo` (filter cancellations, non-typical w/out upcoming trips, arrival only).

There are a few minor adjustments to this logic based on filtering by direction, I'll add specific comments where that is the case.

Still remaining work:
* Adding directions
* Sorting
* Adding equivalent of `hasSchedulesToday` (thinking of punting this entirely to the special service states ticket to unblock the barebones UX work)

iOS

android

### Testing

What testing have you done?
* Ported tests of `NearbyStaticData.withRealtimeInfo` from `NearbyResponseTests` to the new `RouteCard` data type. Made some minor adjustments, will call out where so.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -64,6 +64,13 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
                 is Route -> this.route.type.isSubway()
             }
         }
+
+        fun id(): String {
+            return when (this) {
+                is Line -> this.line.id
+                is Route -> this.route.id
+            }
+        }
     }
 
     companion object {
@@ -374,7 +381,9 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
                                 )
                             }
                 }
+                byStopId.stopData = byStopId.stopData.filterNot { it.value.data.isEmpty() }
             }
+            this.data = this.data.filterNot { it.value.stopData.isEmpty() }
             return this
         }
 
@@ -388,7 +397,7 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
         }
     }
 
-    data class Builder(val lineOrRoute: LineOrRoute, val stopData: ByStopIdBuilder) {
+    data class Builder(val lineOrRoute: LineOrRoute, var stopData: ByStopIdBuilder) {
 
         fun build(): RouteCardData {
             return RouteCardData(this.lineOrRoute, stopData.values.map { it.build() })
@@ -439,9 +448,9 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
                 directionId,
                 checkNotNull(routePatterns),
                 checkNotNull(stopIds),
-                checkNotNull(this.upcomingTrips),
+                this.upcomingTrips ?: emptyList(),
                 alertsHere ?: emptyList(),
-                checkNotNull(allDataLoaded)
+                allDataLoaded ?: false
             )
         }
 
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -6,6 +6,7 @@ import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.ScheduleResponse
 import io.github.dellisd.spatialk.geojson.Position
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
 import kotlinx.datetime.Instant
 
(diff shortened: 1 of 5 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -85,28 +106,27 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
...
+                .filterIrrelevantData(filterAtTime, cutoffTime, context, globalData)
+                .addAlerts(alerts, includeMinorAlerts = context.isStopDetails(), filterAtTime)
                 .build()
                 .sort(sortByDistanceFrom, pinnedRoutes)
         }
     }
 
     class ListBuilder() {
-        var data: ByLineOrRouteBuilder = emptyMap()
+        var data: ByLineOrRouteBuilder = mutableMapOf()
```
KaylaBrady, line 129: My original implementation plan called out making code functional where possible to help with readability. I think the use of mutability here is limited enough while building intermediate state, and hopefully the other readability improvements made throughout these classes makes things generally easier to follow. If there are pieces of the mutations that are hard to follow though, we can change approach.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -259,11 +349,43 @@ data class RouteCardData(private val lineOrRoute: LineOrRoute, val stopData: Lis
...
+                                    showAllPatternsWhileLoading,
+                                    isSubway,
+                                    globalData
+                                )
+                            }
+                            .mapValues {
+                                val (directionId, leafBuilder) = it
+                                leafBuilder
+                                    .filterCancellations(isSubway, context)
+                                    .filterArrivalOnly()
```
KaylaBrady, line 383: note: `NearbyStaticData.withRealtimeInfo` doesn't appear to filter out arrival only upcoming trips, it would only remove entire headsign entries where all are arrival only. 

It seems worth doing this because the Leaf for a single direction could have a mix of arrival-only trips and trips with departures.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -638,10 +662,1577 @@ class RouteCardDataTest {
...
+            // should be sorted after the pattern 2 prediction under Harvard
+            val stop1Pattern1Prediction =
+                objects.prediction {
+                    arrivalTime = time + 5.seconds
+                    departureTime = time + 15.seconds
+                    stopId = stop1.id
+                    trip = trip1
+                }
+
+            // should *not* be ignored since pattern 1 shows at stop, but pattern 3 does not
```
KaylaBrady, line 724: This is a change from `NearbyStaticData.withRealtimeInfo`

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -638,10 +662,1577 @@ class RouteCardDataTest {
...
+        val routePattern1 = objects.routePattern(route1) { representativeTrip { headsign = "A" } }
+        val trip1 = objects.trip(routePattern1)
+
+        val route2 = objects.route { sortOrder = 2 }
+        val routePattern2 = objects.routePattern(route2) { representativeTrip { headsign = "B" } }
+        val trip2 = objects.trip(routePattern2)
+
+        // change from before: this will be included because it has a matching
+        // route & direction, even if it is an unexpected route pattern.
+        val trip3 = objects.trip(routePattern2) { routePatternId = "not the right id" }
```
KaylaBrady, line 1542: This is a change from `NearbyStaticData.withRealtimeInfo` that seems desirable based on the theory behind the initial trip headsigns work - if there is a trip in a known route + direction, we should show that trip.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -638,10 +662,1577 @@ class RouteCardDataTest {
...
+                        )),
+                        allDataLoaded = true,
+                        // TODO: incorporate alert
+                        alertsHere = emptyList()
+
+                    ),RouteCardData.Leaf(
+                        directionId = 1,
+                        routePatterns = listOf(orangeNorthboundTypical, orangeNorthboundDiversion),
+                        stopIds = setOf(northStation.id, northStationSouthboundPlatform.id, northStationNorthboundPlatform.id),
+                        upcomingTrips = listOf(),
```
KaylaBrady, line 2037: This is a change from before: we now don't include the arrival-only trips in the list of upcoming trips, but we do still keep the leaf so that we can show the alert.

BrandonTR (approved):
This is looking really good! Love the extension functions and clear comments.

PR #828, opened 2025-03-14
feat: Handle start/end of service and "later today" for alert active periods

EmmaSimon:
### Summary

_Ticket:_ [Alert Details | format later today & start/end of service](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1208572284247951?focus=true)

Add special handling for alert active periods which start or end at the bounds of a service day, and ones which have an estimated duration certainty. This required a bit of time zone handling to make sure the times being checked were in Eastern time.

This also includes a bit of refactoring of the active period composable on Android, since the text was longer it was more likely to break onto a new line, which messed up the formatting. And since Compose has no provided table component, I needed to add some ugly code to measure the width of start/end text before drawing it, to make the times line up in neat columns and rows.

Simulator Screenshot - iPhone 16 - 2025-03-14 at 10 43 23 Simulator Screenshot - iPhone 16 - 2025-03-14 at 10 45 13
Screenshot_20250314_104750 Screenshot_20250314_104818

iOS

android

### Testing

Added unit tests for the new behavior. I'm pretty sure they'll work properly when run in any time zone.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -10,13 +10,20 @@ data class Alert(
...
     @SerialName("effect_name") val effectName: String?,
     val header: String?,
     @SerialName("informed_entity") val informedEntity: List<InformedEntity>,
     val lifecycle: Lifecycle,
     @SerialName("updated_at") val updatedAt: Instant
 ) : BackendObject {
+    init {
+        // This is done on init to avoid having to pass it in for any call to format an ActivePeriod
+        activePeriod.forEach { period -> period.durationCertainty = durationCertainty }
+    }
```
KaylaBrady, line 25: thought (non-blocking): I looked a bit at alternative approaches for this since it feels a little funky for `durationCertainty on `ActivePeriod` to be nullable, but didn't find any clear cut options besides writing a custom serializer which seems like it could be more complicated than it is worth. 

It could be easier to pass durationCertainty as part of the activePeriods on the backend, but that may also be overkill. 

I think a comment in ActivePeriod about durationCertainty being null could help.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertActivePeriodFormatting.kt
```diff
@@ -1,33 +1,64 @@
...
+import java.util.TimeZone
+import kotlin.time.Duration.Companion.hours
 import kotlinx.datetime.Instant
 
-private fun format(instant: Instant): AnnotatedString {
+private fun format(
+    context: Context,
+    instant: Instant,
+    durationCertainty: Alert.DurationCertainty?,
+    start: Boolean
```
EmmaSimon, line 22: I felt weird about using booleans to handle start/end, but it seemed awkward to add a new shared enum to handle this, if it was only ever used in this one spot.

KaylaBrady (approved):
Sweet! A few minor non-blocking comments.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -10,13 +10,20 @@ data class Alert(
...
     val effect: Effect = Effect.UnknownEffect,
     @SerialName("effect_name") val effectName: String?,
     val header: String?,
     @SerialName("informed_entity") val informedEntity: List<InformedEntity>,
     val lifecycle: Lifecycle,
     @SerialName("updated_at") val updatedAt: Instant
 ) : BackendObject {
+    init {
+        // This is done on init to avoid having to pass it in for any call to format an ActivePeriod
+        activePeriod.forEach { period -> period.durationCertainty = durationCertainty }
```
EmmaSimon, line 24: I considered using a custom serializer here, but it seemed like a ton of overhead for something that shouldn't be that complicated, and this works pretty well. I think even with a custom serializer, it would still need to do effectively the same thing, since the duration certainty applies to the active periods but always exists one level up, the active periods would always need to be serialized first then have the certainty added in afterwards.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetails.kt
```diff
@@ -222,21 +224,55 @@ private fun AlertTitle(
...
+    // period to be in rows together in case the period needs to break onto multiple lines.
+    val density = LocalDensity.current
+    val style = Typography.callout
+    val textMeasurer = rememberTextMeasurer()
+    val columnTexts = listOf(stringResource(R.string.start), stringResource(R.string.end))
+
+    val columnWidth = remember {
+        columnTexts.maxOf { text ->
+            with(density) {
+                textMeasurer.measure(text = AnnotatedString(text), style = style).size.width.toDp()
```
EmmaSimon, line 242: I considered using some other method to measure the width of the elements, but this works pretty well and is relatively straightforward.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetails.kt
```diff
@@ -222,21 +224,55 @@ private fun AlertTitle(
 
 @Composable
 private fun AlertPeriod(currentPeriod: Alert.ActivePeriod?) {
+    var startWidth by remember { mutableStateOf<Int?>(null) }
+    var endWidth by remember { mutableStateOf<Int?>(null) }
+
+    // This is all because Jetpack Compose has no built in table layout, we need the "Start" and
+    // "End" label texts to take up the same column width, but they can be variable widths in
+    // different languages, so we can't use a fixed size, but we also need the label and formatted
+    // period to be in rows together in case the period needs to break onto multiple lines.
```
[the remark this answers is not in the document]
EmmaSimon: I didn't find any obvious 3rd party libraries that would fit exactly what we need. I feel like most table libraries are more geared towards data tables and have a bunch of extra functionality around filtering and sorting, and most grid libraries are based around the assumption that you want them to lazy load arbitrarily long amount of info.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertActivePeriodFormatting.kt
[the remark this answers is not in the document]
EmmaSimon: Yeah good call!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetails.kt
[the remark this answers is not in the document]
EmmaSimon: It's only very slightly different, but Vietnamese looks like one of the only ones where end is longer. This is still useful even if end is always shorter because the longest will always be a different length in different languages/text sizes that needs to be set the same for both labels, and the date and time text needs to be able to break without messing up the positioning of the start/end labels.

Screenshot_20250317_112326

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
KaylaBrady: Hah yeah, glad we were thinking along the same lines!

PR #830, opened 2025-03-17
perf: move expensive calculations to background in shared code

boringcactus:
### Summary

_Ticket:_ [Move expensive calculations to background thread in shared code](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209077852914470?focus=true)

I'm removing the explicit dispatchers from the Android side in the places where I'm adding them in shared code instead.

Since `runBlocking` implicitly returns the last value in its block and `assertNotNull(x)` returns x, I've had to explicitly mark some tests as returning `Unit` to make the compiler happy.

iOS

android

### Testing

Updated tests. Checked that the app isn't obviously completely broken in a way that the tests aren't catching.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
@@ -23,6 +23,8 @@ import com.mbta.tid.mbta_app.utils.resolveParentId
 import io.github.dellisd.spatialk.geojson.LineString
 import io.github.dellisd.spatialk.turf.ExperimentalTurfApi
 import io.github.dellisd.spatialk.turf.lineSlice
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import kotlinx.serialization.json.put
 
 data class RouteLineData(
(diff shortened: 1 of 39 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
```diff
@@ -48,36 +50,40 @@ object RouteFeaturesBuilder {
     val propRouteColor = FeatureProperty<Color>("routeColor")
     val propAlertStateKey = FeatureProperty<String>("alertState")
 
-    fun generateRouteLines(
+    suspend fun generateRouteLines(
         routeData: List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>,
         routesById: Map<String, Route>?,
         stopsById: Map<String, Stop>?,
         alertsByStop: Map<String, AlertAssociatedStop>?
     ): List<RouteLineData> {
```
BrandonTR, line 58: I noticed here and a few other places we're not using the shorthand `): List<RouteLineData> = withContext(Dispatchers.Default) {`. Is there a distinction in functionality?

boringcactus: There's not; for functions that already have an explicit `return` I'm more likely to leave it in place than I am to rewrite to `fun x(): T =`, although I'm not consistent with that.

PR #836, opened 2025-03-18
fix: Make the cache not erase unknown backend fields

EmmaSimon:
### Summary

_Ticket:_ [Fix stale parsed cache data](https://app.asana.com/0/1205732265579288/1209717539725073/f)

Fixes a couple related issues with the cache, the first was that all data was being parsed into the cache object type before being written to disk, which means that any data which existed in the response but was not included in the object would be erased, while leaving the etag the same, which resulted in the missing fields never being reloaded even if they were added to the object type. Now we write the response body directly to the disk cache without parsing it into an object first (though a parse does still happen prior to the write, so if the parse fails, nothing will be written).

Then, even with that fixed, we needed a way for the client to know that it was possible for the current cache to be invalid, even if the etag matches the backend, so I added a hardcoded `invalidationKey`, which is an arbitrary string which gets written to disk along with the data. If the loaded invalidation key doesn't match the one that the cache expects, it clears the data and reloads from the backend.

### Testing

Added unit tests for the new behavior. Manually tested locally by loading the data on the 1.2.1 build, then switching to this branch and verifying that it was re-requesting and displaying the added fields properly.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
@@ -16,7 +16,6 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.encodeToString
 import kotlinx.serialization.serializer
 import okio.FileSystem
 import okio.Path
@@ -68,7 +67,7 @@ class ResponseCache<T : Any>(
 
     private val lock = Mutex()
 
-    private fun decodeData(body: String): T {
+    private fun decodeString(body: String): T {
         return json.decodeFromString(serializer, body)
     }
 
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
```diff
@@ -27,21 +26,26 @@ import org.koin.core.component.inject
 data class ResponseMetadata(
     val etag: String?,
     @Serializable(with = TimeMarkSerializer::class)
-    var fetchTime: TimeSource.Monotonic.ValueTimeMark
+    var fetchTime: TimeSource.Monotonic.ValueTimeMark,
+    val invalidationKey: String? = null
```
KaylaBrady, line 30: suggestion(non-blocking): Maybe worth a little doc about using this param!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/GlobalRepository.kt
```diff
@@ -22,7 +22,8 @@ interface IGlobalRepository {
 }
 
 class GlobalRepository(
-    val cache: ResponseCache<GlobalResponse> = ResponseCache.create(cacheKey = "global")
+    val cache: ResponseCache<GlobalResponse> =
+        ResponseCache.create(cacheKey = "global", invalidationKey = "2025-03-18")
```
KaylaBrady, line 26: question(non-blocking): Should the rail shapes cache get an invalidationKey too? We aren't updating the returned data type anytime soon AFAIK, but might be nice to do at this point just so that we don't run into this same problem in the future for existing users.

KaylaBrady (approved):
Awesome! Nice testing 🎉

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/GlobalRepository.kt
EmmaSimon: Oh hmm, maybe yeah. It might be a slightly better experience to only add it if/when we make any changes to the returned data types on the frontend, just so clients don't have to refetch unless strictly necessary, but I'm 90% sure we'd forget to actually set this whenever we decide to change that.

KaylaBrady: I could go either way on it... it doesn't seem likely that we will end up changing that data anytime soon. We could always just make a ticket for it so that it is documented & we can included in a subsequent release if needed.

EmmaSimon: I added it, this way we know that we're starting from a fresh slate and can always rely on the cached data to match the backend response going forward.

PR #837, opened 2025-03-19
feat: Display subway delay alerts

KaylaBrady:
### Summary

_Ticket:_ [Display subway delays](https://app.asana.com/0/1205732265579288/1208643635656244/f)

What is this PR for?

This adds subway delay alerts to iOS and android. Most of the volume of this change comes from adding translations for the lowercased alert causes so that they can be popped into the "Delays due to X" string.

The overhead of adding the strings in ios & android did make me wish a bit that we were sharing localizations across platform, but it wasn't all too time consuming to add in both places.

## NOTE: 
Backend must be redeployed before this is released

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran on iOS & android. Confirmed alert for a severe subway delay showed, informational subway and all bus delays did not.
IMG_0250Screenshot_20250319_120322


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCard.kt
```diff
@@ -34,7 +34,8 @@ enum class AlertCardSpec {
     Major,
     Downstream,
     Secondary,
-    Elevator
+    Elevator,
+    Delay
```
KaylaBrady, line 38: I was on the fence about adding `Delay` as the AlertCardSpec or `Minor`, but `Delay` seemed to make the logic the simplest.

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -2471,6 +2518,53 @@
...
+            "state" : "needs_review",
+            "value" : "Tráfego de trem Amtrak"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Giao thông đường sắt Amtrak"
+          }
+        },
```
EmmaSimon, line 2553: Shouldn't everything besides "Amtrak" be lowercase here?

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -4177,6 +4412,53 @@
...
+            "state" : "needs_review",
+            "value" : "Restrição da Guarda Costeira"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Hạn chế của Cảnh sát biển"
+          }
+        },
```
EmmaSimon, line 4447: Same thing here?

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -8137,37 +9453,84 @@
...
+            "state" : "needs_review",
+            "value" : "manutenção"
           }
         },
         "vi" : {
           "stringUnit" : {
-            "state" : "translated",
-            "value" : "Bảo trì"
+            "state" : "needs_review",
+            "value" : "BẢO TRÌ"
```
EmmaSimon, line 9486: Why is this all caps?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCard.kt
```diff
@@ -51,6 +52,9 @@ fun AlertCard(
         when (spec) {
             AlertCardSpec.Downstream -> formattedAlert.downstreamEffect
             AlertCardSpec.Elevator -> alert.header ?: formattedAlert.effect
+            AlertCardSpec.Delay ->
+                formattedAlert.cause?.let { stringResource(R.string.delays_due_to_cause, it) }
+                    ?: stringResource(R.string.delays_unknown_reason)
```
boringcactus, line 57: As I recall, "X between Y and Z due to A" (or "X due to A" or "X between Y and Z") is the way we're interested in presenting a lot of alerts, and delays are just the first alert that we're handling that way. I was about to suggest making this more generic, but since calculating the boundaries of the alert is the actual hard part and delays generally apply to the whole route, it may not be worth being slightly more generic now if we're still going to have to do all the hard work later.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/FormattedAlert.kt
```diff
@@ -30,6 +32,10 @@ data class FormattedAlert(
         @Composable
         get() = stringResource(R.string.effect_ahead, stringResource(downstreamEffectRes))
 
+    // TODO: Decide if should be nullable or not
+    val cause
```
boringcactus, line 36: I think naming this `cause` might be confusing later if we want to move the title-case cause text into `FormattedAlert` at some point - the way this PR is set up right now, `@StringRes val dueToCauseRes: Int?` and `val delaysDueToCause @Composable get()` might be the structure that rhymes best with the `effect`/`downstreamEffect` distinction that's in here now.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCard.kt
KaylaBrady: Yeah that is true, happy to hold off on making it more generic for now 😄

boringcactus (approved):

PR #840, opened 2025-03-21
feat: Initial direction grouping UX

EmmaSimon:
### Summary

_Ticket:_ [Group by direction | 🤖 Barebones Nearby Transit UX](https://app.asana.com/0/1205732265579288/1209536961544700/f)

Sorry about the line count, this PR looks scary but it's almost entirely renames/field changes, tests, and shifting existing code around.

Notes:
* This moves the `RealtimePatterns.Format` class out into an independent `UpcomingFormat`, which is used by both `RouteCardData.Leaf` and `RealtimePatterns`. I also updated `UpcomingFormat.Some.FormattedTrip` to store the full `UpcomingTrip` that was used to create it, rather than just the trip ID. This is so that the `Departures` component can display the trip headsign alongside each formatted departure, and for future use in departure tiles to replace `TripAndFormat`, which was created to work around `UpcomingTrip` not being included in the format object.
  * We should consider moving the headsign to `UpcomingFormat.Some.FormattedTrip` relationship into `UpcomingFormat`, possibly with something like `UpcomingFormat.SomeMatchingHeadsigns` and `UpcomingFormat.SomeDistinctHeadsigns`, but that's a bit difficult right now because of the way that `HeadsignRowView`/`PredictionRowView` work, they can only handle a single text label for each row, so we'd need to do some refactoring to have an arbitrary number of headsigns associated with their individual `UpcomingFormat`s.
* I added `Context` and `Instant`  fields to `RouteCardData`. The context was needed in the `RouteCard` to determine whether to render the stop name and to format the leaves. It seemed weird to pass the context in separately, so I included it in the object. The time wasn't totally necessary, we're not using it yet, but I think including something to check if the time the data was created is drastically different from the current time will be a good way to prevent some issues with stale data. 
  * We should also refactor `routeCardsForStopList` to not pass in the `context` or `now` to any of the builder functions that use it, since it's now included in the base object, but I was trying to avoid any more refactoring than is already here.
* I moved a good chunk of the convoluted existing `RouteCard`/`TransitCard` composables into a `legacyRouteCard` package, and created a new `routeCard` package. I'm hoping that the new structure will be much easier to work with, and won't involve awkwardly piecing together all the separate parts like we're doing in `NearbyStopView` and `StopDetailsRouteView` (which should eventually be removed too, I didn't move them into `legacyRouteCard` because they exist in their own `nearbyTransit` and `stopDetails` page packages).
* I updated `LineOrRoute` to use computed properties rather than functions that take no arguments, and resolved all the overlapping line/route properties there, now we can just use the properties on `LineOrRoute` and not really worry about which one it is.

### Testing

Updated existing tests to pass, added basic tests for newly added functionality (`Leaf.format` and `RouteCard` components).


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/routeCard/Departures.kt
```diff
@@ -0,0 +1,93 @@
...
+        fun analyticsTappedDeparture(predictions: UpcomingFormat) {
+            val noTrips =
+                when (predictions) {
+                    is UpcomingFormat.NoTrips -> predictions.noTripsFormat
+                    else -> null
+                }
+            analytics.tappedDeparture(
+                cardData.lineOrRoute.id,
+                stopData.stop.id,
+                // TODO: Handle pinning
```
KaylaBrady, line 43: 👍 Could you add this into one of the follow-up tickets (maybe polish?) I worry since it isn't core functionality it might get lost and our analytics could become misleading

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/routeCard/RouteCard.kt
```diff
@@ -0,0 +1,35 @@
...
+    data: RouteCardData,
+    now: Instant,
+    showElevatorAccessibility: Boolean = false,
+    onOpenStopDetails: (String, StopDetailsFilter) -> Unit
+) {
+    Column(Modifier.haloContainer(1.dp)) {
+        TransitHeader(data.lineOrRoute)
+
+        data.stopData.forEach {
+            if (data.context == RouteCardData.Context.NearbyTransit) {
```
KaylaBrady, line 23: 👍 this is nice, good call keeping the context as part of the data

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
```diff
@@ -91,80 +97,137 @@ fun NearbyTransitView(
...
         Text(
             text = stringResource(R.string.nearby_transit),
             modifier = Modifier.semantics { heading() }.padding(horizontal = 16.dp),
             style = Typography.title3Semibold,
         )
         ErrorBanner(errorBannerViewModel)
-        if (nearbyWithRealtimeInfo == null) {
-            CompositionLocalProvider(IsLoadingSheetContents provides true) {
+        if (groupByDirection) {
+            rememberSuspend(
```
KaylaBrady, line 108: suggestion: Since the result of this isn't stored anywhere and is stored in the VM (which is a great improvement!) I don't think we need to do rememberSuspend anymore and can instead do a LaunchedEffect.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
```diff
@@ -10,54 +10,62 @@ import kotlinx.coroutines.suspendCancellableCoroutine
...
+
     suspend fun getNearby(global: GlobalResponse, location: Position): ApiResult<NearbyStaticData>
 }
 
 class NearbyRepository : KoinComponent, INearbyRepository {
-    override suspend fun getNearby(global: GlobalResponse, location: Position) =
-        ApiResult.runCatching {
-            val searchPosition =
-                Position(latitude = location.latitude, longitude = location.longitude)
+    override fun getStopIdsNearby(global: GlobalResponse, location: Position): List<String> {
```
KaylaBrady, line 19: suggestion: Maybe worth a simple test case for this new fn

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/routeCard/Departures.kt
```diff
@@ -0,0 +1,93 @@
...
+        ) {
+            when (formatted) {
+                is UpcomingFormat.Some -> {
+                    DirectionLabel(
+                        direction,
+                        Modifier.padding(start = 8.dp, top = 8.dp),
+                        showDestination = false
+                    )
+                    formatted.trips.forEach {
+                        HeadsignRowView(
```
KaylaBrady, line 76: 👍 using the existing headsign row view makes sense for now. I suspect we'll want to move away from it when we get to distinguishing between branching and non-branching routes in the UI, but we can cross that bridge later.

KaylaBrady (commented):
This is looking good! A few small comments. 
This probably would have been a bit easier to review if the renaming / refactoring bits were in a separate PR from the new UI changes, but hindsight 20/20! Something to keep in mind for next time.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/routeCard/Departures.kt
EmmaSimon: I think it makes sense for it to be in polish, but we could maybe also make a ticket just for getting pinning functionality in place?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
EmmaSimon: Ah, yes, I meant to clean this up then forgot, thanks for catching it!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/routeCard/Departures.kt
EmmaSimon: Yeah, I think so too. Not sure if we necessarily will even want `PredictionRowView`, since the way that's structured will make refactoring things difficult.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/DirectionLabel.kt
```diff
@@ -31,11 +31,12 @@ private fun directionNameFormatted(direction: Direction) =
...
     direction: Direction,
     modifier: Modifier = Modifier,
-    textColor: Color = LocalContentColor.current
+    textColor: Color = LocalContentColor.current,
+    showDestination: Boolean = true
 ) {
     val destination = direction.destination
     Column(modifier = modifier) {
-        if (destination != null) {
+        if (destination != null || !showDestination) {
```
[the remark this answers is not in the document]
EmmaSimon: Splitting it out into composables makes it cleaner, but now I'm wondering if I should just get rid of `showDestination` entirely and just use `DirectionTo` by itself in the route card.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
EmmaSimon: I just simplified all the existing test cases in `NearbyRepositoryTest` to test this directly, since in practice that's what they were doing, then added a separate test case to test the `NearbyStaticData` creation.

EmmaSimon:
> This probably would have been a bit easier to review if the renaming / refactoring bits were in a separate PR from the new UI changes, but hindsight 20/20! Something to keep in mind for next time.

I know I've done the same thing before, but I always get far enough into the UI before realizing that refactoring needs to happen, then don't realize exactly how much refactoring will need to happen until I'm a good chunk of the way through integrating them with the UI changes, and by then the change sets are intertwined enough that it would be painful to split them up 😭

KaylaBrady (approved):
🎉

PR #847, opened 2025-03-26
feat(iOS): Wheelchair boarding in nearby transit & stop details

BrandonTR:
### Summary

_Ticket:_ [ | Elevator Status | Wheelchair boarding in nearby transit & stop details](https://app.asana.com/0/1205732265579288/1209144605811987/f)

Adds wheelchair boarding information in nearby transit & stop details.
Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 08 42 34
Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 08 42 47
Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 08 42 22

iOS

android

### Testing

Added some automated tests for wheelchair boarding and fixed elevator tests.


iosApp/iosApp/Pages/StopDetails/StopDetailsUnfilteredView.swift
```diff
@@ -90,31 +94,37 @@ struct StopDetailsUnfilteredView: View {
...
+                                                        elevator: true
+                                                    )
+                                                }
+                                            )
+                                            .padding(.horizontal, 16)
+                                            .padding(.bottom, 16)
+                                        }
+                                    }
+                                } else {
+                                    NotAccessibleCard()
```
KaylaBrady, line 125: issue: I don't think the NotAccessibleCard should be showing up at Government Center - it correctly gets the ISA icon in nearby transit. On android there is no NotAccessibleCard as expected, maybe some small difference in the logic there! 

thought(non-blocking): - this makes me think if there are ways we could approach android implementation differently in the future to share more logic with iOS. Would a common function given accessibility inputs that returns an enum value depending on the conditional branch case you're in have helped? Would love to know if you have any ideas about that from your experience porting to iOS!

KaylaBrady (approved):

iosApp/iosApp/Pages/StopDetails/StopDetailsUnfilteredView.swift
KaylaBrady: Yeah, I'm with you on being nervous about the feasibility of fully shared VMs! It is a good callout though - if more of the logic is in VMs, I think it will be easier to recognize & extract utils that could be common to the VMs, even if the representation of state / lifecycle objects stay native.

PR #848, opened 2025-03-26
fix(TripInstantDisplay): BRD / ARR when predicted arrival in the past

KaylaBrady:
### Summary

_Ticket:_ [BRD bug when train dwelling at station](https://app.asana.com/0/1205732265579288/1209702051776246/f)

What is this PR for?

When a train is dwelling at a station, show it as BRD to match in-station countown clocks. 
Similarly, if we predict that a train is already be at a station (but it doesn't have the StoppedAt boarding status), show ARR, even if departure is predicted more than 30 seconds out.

**Note:** This PR description suggested considering whether to refactor out `TripInstantDisplay.from` in conjunction with this change. I considered it but ultimately this change was so simple I don't think it is necessary to tie it to this change. After thinking more about how I would approach that refactor, I'm also not totally sure we'd want to strictly split up this logic by mode, as there is some shared logic & entangled pieces that I think worth be worth putting more thought into the refactor.

iOS

android

### Testing

What testing have you done?
* Added unit tests for each new case

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
@@ -138,14 +138,14 @@ sealed class TripInstantDisplay {
                 vehicle?.currentStatus == Vehicle.CurrentStatus.StoppedAt &&
                     vehicle.stopId == prediction.stopId &&
                     vehicle.tripId == prediction.tripId &&
-                    timeRemaining <= BOARDING_CUTOFF
+                    (timeRemaining <= BOARDING_CUTOFF || prediction.hasArrivedButNotDeparted(now))
             ) {
                 return Boarding
             }
             if (timeRemaining.isNegative()) {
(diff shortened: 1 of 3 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
```diff
@@ -154,4 +154,6 @@ sealed class TripInstantDisplay {
             return Minutes(minutes)
         }
     }
+
+    fun scheduleBasedDisplay() {}
```
boringcactus, line 158:
```suggestion
```

boringcactus (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
KaylaBrady: TY! I started thinking about the refactor and forgot to back it all out

PR #849, opened 2025-03-26
refactor: Check start/end of service in shared

EmmaSimon:
### Summary

_Ticket:_ Precursor to [Include affected stop bounds in downstream disruption summary](https://app.asana.com/0/1205732265579288/1209527076283980/f)

Splitting this out from the alert templating changes because it can stand on its own, I didn't realize when I did this initially that it was possible to do entirely within Kotlin.

iOS

android

### Testing

Existing tests should cover all of this

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -1,6 +1,8 @@
 package com.mbta.tid.mbta_app.model
 
 import kotlinx.datetime.Instant
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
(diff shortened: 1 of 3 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -77,6 +79,23 @@ data class Alert(
             }
             return instant in start..end
         }
+
+        val fromStartOfService: Boolean
+            get() {
+                val localTime = start.toLocalDateTime(TimeZone.of("America/New_York"))
```
BrandonTR, line 85: Nitpick: Could be worth making a small private convenience func on Instant:
```
    fun Instant.toBostonTime(): LocalDateTime {
        return this.toLocalDateTime(TimeZone.of("America/New_York"))
    }
```

EmmaSimon: Good idea, I'm doing this even more in my current changes, this would simplify things!

PR #850, opened 2025-03-27
refactor(RouteCardData): simplify addStaticData step while working on GL park st bug

KaylaBrady:
### Summary

_Ticket:_ Kind of [Group by direction | Check on old trip headsign issues](https://app.asana.com/0/1205732265579288/1209536961544705/f) since the spirit of the problem is the same as the old Kenmore bug, but also bug report from @boringcactus.

What is this PR for?

This PR restructures the `addStaticData` step to actually group all routes for a line into a single record. I wrote `addStaticData` originally and had kind of a hard time reading through it, so I did a bit of refactoring to help my own understanding of it and help me isolate where the root of the bug was. 

The refactoring did help to fix the bug a little, but not entirely. The core of the problem is that the map of Line => Patterns was being built at the *child stop* level, so when keying by `LineOrRoute`, the list of routes for the GL weren't matching up at Park St because different child stops serve different routes. Will split out the actual fix as a separate PR so this refactor can be reviewed in isolation.

iOS

android

### Testing

What testing have you done?
* Confirmed no regressions
* Ran the app locally. This does actually help the Park St bug but doesn't resolve the issue entirely. Will add a new test when the issue is fully resolved.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -306,90 +306,51 @@ data class RouteCardData(
          */
         fun addStaticStopsData(stopIds: List<String>, globalData: GlobalResponse): ListBuilder {
 
-            val routePatternsUsed = mutableSetOf<String>()
-
-            val patternsGrouped =
-                mutableMapOf<LineOrRoute, MutableMap<Stop, MutableList<RoutePattern>>>()
-
-            val parentToAllStopIds = mutableMapOf<String, MutableSet<String>>()
-
-            globalData.run {
-                stopIds.forEach { stopId ->
-                    val stop = stops[stopId] ?: return@forEach
-                    val patternsByRouteOrLine =
-                        patternsByRouteOrLine(stopId, globalData)
-                            // filter out a route if we've already seen all of its patterns
-                            .filterNot {
-                                routePatternsUsed.containsAll(
-                                    it.value.map { pattern -> pattern.id }.toSet()
-                                )
-                            }
-                    routePatternsUsed.addAll(
-                        patternsByRouteOrLine.flatMap { it.value }.map { it.id }
-                    )
-
-                    stop.parentStationId?.let { parentStationId ->
-                        parentToAllStopIds
-                            .getOrPut(parentStationId) { mutableSetOf(parentStationId) }
-                            .add(stop.id)
+            // A map of a stop to itself and any children.
+            // for standalone stops, an entry will be <standaloneStop, [standaloneStopId]>.
+            // for stations, an entry will be <station, [stationId, child1Id, child2Id, etc.]>
+            val parentToAllStops: Map<Stop, Set<String>> =
+                stopIds
+                    .mapNotNull { stopId ->
+                        val stop = globalData.stops[stopId]
+                        if (stop != null) {
+                            Pair(stop.resolveParent(globalData), stop)
+                        } else {
+                            null
+                        }
                     }
+                    .groupBy({ it.first }, { it.second.id })
+                    .mapValues { it.value.toSet().plus(it.key.id) }
 
-                    val stopOrParent = stop.resolveParent(stops)
-
-                    for ((routeOrLine, routePatterns) in patternsByRouteOrLine) {
-                        val routeStops = patternsGrouped.getOrPut(routeOrLine) { mutableMapOf() }
-                        val patternsForStop = routeStops.getOrPut(stopOrParent) { mutableListOf() }
-                        patternsForStop += routePatterns
-                    }
-                }
-            }
+            val patternsGrouped = patternsGroupedByLineOrRouteAndStop(globalData, parentToAllStops)
 
             val builderData =
                 patternsGrouped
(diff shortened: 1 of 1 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -306,90 +306,51 @@ data class RouteCardData(
...
-                    )
-
-                    stop.parentStationId?.let { parentStationId ->
-                        parentToAllStopIds
-                            .getOrPut(parentStationId) { mutableSetOf(parentStationId) }
-                            .add(stop.id)
+            // A map of a stop to itself and any children.
+            // for standalone stops, an entry will be <standaloneStop, [standaloneStopId]>.
+            // for stations, an entry will be <station, [stationId, child1Id, child2Id, etc.]>
+            val parentToAllStops: Map<Stop, Set<String>> =
```
KaylaBrady, line 312: pulling this up to be the first step rather than building parentToAllStopIds incrementally while iterating over all stops I think will result in some extra looping, but on a relatively small dataset, and I think worth it for the readability gains.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -306,90 +306,51 @@ data class RouteCardData(
...
             val builderData =
                 patternsGrouped
-                    .map { byLineOrRoute ->
-                        val key =
-                            when (val cardType = byLineOrRoute.key) {
-                                is LineOrRoute.Line -> cardType.line.id
-                                is LineOrRoute.Route -> cardType.route.id
-                            }
-                        // TODO: Directions list with actual values.
+                    .map { (lineOrRoute, patternsByStop) ->
```
KaylaBrady, line 329: I think I forgot on my initial writing of this that you can de-structure map entries like this!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -306,90 +306,51 @@ data class RouteCardData(
...
-                                                is LineOrRoute.Route ->
-                                                    Direction.getDirections(
-                                                        globalData,
-                                                        byStop.key,
-                                                        lineOrRoute.route,
-                                                        byStop.value
-                                                    )
-                                            }
-                                        byStop.key.id to
+                                            lineOrRoute.directions(globalData, stop, patterns)
```
KaylaBrady, line 337: Also forgot to use the helper directions fn!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -407,40 +368,83 @@ data class RouteCardData(
...
                             if (line != null && !route.isShuttle && line.isGrouped) {
                                 // set routes empty for now, will populate once all routes
                                 // of the line at this stop are known in the next step
                                 LineOrRoute.Line(line, routes = emptySet())
                             } else LineOrRoute.Route(route)
                         } else null
                     }
-                    .mapNotNull { entry ->
-                        when (val key = entry.key) {
+                    .mapNotNull { (partialLineOrRoute, patternRoutePairs) ->
```
KaylaBrady, line 397: Now that we're using the full list of `stopIds` for a stop, we know `patternRoutePairs` is comprehensive and can safely build the LineOrRoute.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -407,40 +368,83 @@ data class RouteCardData(
...
+                            .filterNot {
+                                routePatternsUsed.containsAll(
+                                    it.value.map { pattern -> pattern.id }.toSet()
+                                )
+                            }
+                    routePatternsUsed.addAll(
+                        patternsByRouteOrLine.flatMap { it.value }.map { it.id }
+                    )
+                    for ((routeOrLine, routePatterns) in patternsByRouteOrLine) {
+                        val routeStops = patternsGrouped.getOrPut(routeOrLine) { mutableMapOf() }
```
KaylaBrady, line 440: shoot I think this could theoretically still be an issue where the GL `RouteOrLine.Line` might not exactly match at different stops. Going to rework a little bit more

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
KaylaBrady: https://github.com/mbta/mobile_app/pull/851

EmmaSimon: This is more a comment for the follow up, but would it be a huge pain in the ass to write a test case for this? It might be helpful to have that covered so we don't accidentally break it again.

EmmaSimon (commented):
Nice, this is definitely easier to follow!

EmmaSimon (approved):

PR #851, opened 2025-03-27
fix(RouteCardData): Grouping data by Line / Park St missing Westbound data bug

KaylaBrady:
### Summary

Ticket: Kind of [Group by direction | Check on old trip headsign issues](https://app.asana.com/0/1205732265579288/1209536961544705/f) since the spirit of the problem is the same as the old Kenmore bug, but also bug report from @boringcactus. Follow-up to #850 

What is this PR for?

This PR actually fixes the GL line grouping issue in a more robust way to prevent any other edge cases. @boringcactus suggested [here](https://github.com/mbta/mobile_app/pull/809#discussion_r1984069356):

> It might be simpler to just count all GL branches as always included at every GL stop, though; I'm not sure if that'd actually break anything but if it wouldn't it'd be way simpler.

As far as I can tell, this doesn't actually break anything - we only use the routes to support `sortOrder` and `isSubway`, which does not depend on the particular routes included. 

iOS

android

### Testing

What testing have you done?
* Confirmed I can see both directions now at Park St
Screenshot_20250327_113724

* Added a unit test that fails prior to #850


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -372,41 +372,37 @@ data class RouteCardData(
...
+                allPatternsAtStopWithRoute.groupBy(
+                    { (route, _pattern) ->
+                        val line = route.lineId?.let { globalData.lines[it] }
+                        if (line != null && !route.isShuttle && line.isGrouped) {
+                            LineOrRoute.Line(
+                                line,
+                                routes =
+                                    globalData.routesByLineId
+                                        .getOrElse(line.id) { emptyList() }
+                                        .filterNot { it.isShuttle }
```
EmmaSimon, line 399: Couldn't this be removed since it's happening in `routesByLineId`?

KaylaBrady: Ah yeah, I went back and forth on where the better place for it would be, but I'm sold on putting it in `routesByLineId` for simplicity since right now we never want to include shuttle routes in a line.

boringcactus (approved):

PR #852, opened 2025-03-27
refactor(Android): accept objects in StopDetailsViewModel.mocked

boringcactus:
### Summary

_Ticket:_ [Group by direction | 🤖 Barebones Unfiltered Stop Details UX](https://app.asana.com/0/1205732265579288/1209536961544712/f)

Saw the opportunity for this as part of writing the tests for this ticket, and it would be hypocritical of me not to split this out.

iOS
- ~~[ ] If you added any user-facing strings on iOS, are they included in Localizable.xcstrings?~~
  - ~~[ ] Add temporary machine translations, marked "Needs Review"~~

android
- ~~[ ] All user-facing strings added to strings resource in alphabetical order~~
- ~~[ ] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible (ideally in shared code)~~

### Testing

Checked that all the relevant tests still pass.

PR #853, opened 2025-03-27
feat(Android): group by direction in unfiltered stop details

boringcactus:
### Summary

_Ticket:_ [Group by direction | 🤖 Barebones Unfiltered Stop Details UX](https://app.asana.com/0/1205732265579288/1209536961544712/f)

Stacked on #852. This was not as free as I had thought it would be, but it's not all that much code.

iOS

android

### Testing

Added new unit tests and manually verified that stop details still works (aside from the bug fixed in #851).


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsUnfilteredView.kt
```diff
@@ -40,7 +41,9 @@ fun StopDetailsUnfilteredView(
 
     val analytics: Analytics = koinInject()
 
-    val departures by viewModel.stopDepartures.collectAsState()
+    val groupByDirection by viewModel.groupByDirection.collectAsState()
+    val departures = viewModel.stopDepartures.collectAsState().value
+    val routeCardData = viewModel.routeCardData.collectAsState().value
```
boringcactus, line 46: I'm confident that `= state.value` is equivalent to `by state` from a functionality perspective and also simplifies the null handling later on (since the getter is allowed to do all sorts of wacky stuff, and under the hood, the `SnapshotMutableStateImpl.value` getter does, but it [WOULD PROBABLY](https://www.rfc-editor.org/rfc/rfc6919#section-5) be idempotent within a single invocation of the component).

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
```diff
@@ -559,6 +572,45 @@ fun stopDetailsManagedVM(
...
+                            globalResponse.getStop(stopId)?.childStopIds.orEmpty(),
+                            globalResponse,
+                            sortByDistanceFrom = null,
+                            schedules,
+                            stopData?.predictionsByStop?.toPredictionsStreamDataResponse(),
+                            alertData,
+                            now,
+                            pinnedRoutes,
+                            // TODO do something about filtered stop details
+                            context = RouteCardData.Context.StopDetailsUnfiltered
```
boringcactus, line 606: Since the context is an argument to `routeCardsForStopList` but we use the same data for unfiltered and filtered stop details, this is going to be a problem when we're ready to move filtered stop details to consuming the RouteCardData, but it's not a problem yet, so my inclination is to not worry about it.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -254,7 +256,7 @@ data class RouteCardData(
          * [filterAtTime] and [filterAtTime] + [hideNonTypicalPatternsBeyondNext] are omitted.
          * Cancelled trips are also omitted when [context] = NearbyTransit.
          */
-        fun routeCardsForStopList(
+        suspend fun routeCardsForStopList(
```
boringcactus, line 259: Per the PR checklist, since this will likely be expensive (although hopefully not very expensive), it belongs in a background thread.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -264,33 +266,34 @@ data class RouteCardData(
...
-                when (context) {
-                    Context.NearbyTransit -> 120.minutes
-                    Context.StopDetailsFiltered -> 120.minutes
-                    Context.StopDetailsUnfiltered -> null
-                }
+                val hideNonTypicalPatternsBeyondNext: Duration? =
+                    when (context) {
+                        Context.NearbyTransit -> 120.minutes
+                        Context.StopDetailsUnfiltered -> 120.minutes
+                        Context.StopDetailsFiltered -> null
```
boringcactus, line 283: There's an acceptance criterion that says that unfiltered stop details should have the same rules as nearby transit, which doesn't match what was previously implemented here.

BrandonTR (approved):
Nice!

PR #855, opened 2025-03-28
feat: create new dev-orange variant

boringcactus:
### Summary

_Ticket:_ [iOS & Android dev orange apps](https://app.asana.com/0/1205732265579288/1208842736619571/f)

I’m really glad we worked on managing the Xcode project first; this was easy.

Deploys to Android won’t work until the Android console listing is created; I’ve filed the infra request for that.

iOS

android

### Testing

Checked that the new apps have the right name and icon, and that alerts shown in staging are not visible in dev-orange.

Checked that iOS deploys work now and Android deploys will probably work once the Google Play console is set up.


.github/workflows/android-dev-orange-deploy.yaml
```diff
@@ -0,0 +1,53 @@
...
+      build-type: "Release"
+      deploy: true
+      flavor: "DevOrange"
+      sentry-environment: "dev-orange"
+    secrets:
+      AWS_ROLE_ARN: ${{ secrets.AWS_ROLE_ARN }}
+      FIREBASE_KEY: ${{ secrets.FIREBASE_KEY }}
+      GCP_WORKLOAD_IDENTITY_PROVIDER: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+      GCP_SERVICE_ACCOUNT: ${{ secrets.GCP_SERVICE_ACCOUNT }}
+      GOOGLE_APP_ID_ANDROID: ${{ secrets.GOOGLE_APP_ID_ANDROID_DEV_ORANGE }}
```
boringcactus, line 51: This secret and its iOS cousin will need to be populated - I’ve created the apps in Firebase but I don’t have access to the GitHub repo settings.

KaylaBrady: Added!

.github/workflows/android-dev-orange-deploy.yaml
```diff
@@ -0,0 +1,53 @@
+name: Android dev-orange deploy
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - synchronize
+      - labeled
+  schedule:
+    - cron: "13 12 * * MON"
```
KaylaBrady, line 10: question (non-blocking): Why different cron than backend?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/more/MoreSectionView.kt
```diff
@@ -38,7 +38,7 @@ fun MoreSectionView(section: MoreSection, toggleSetting: ((Settings) -> Unit)) {
 
     val note = section.note
 
-    if (!(section.requiresStaging && appVariant != AppVariant.Staging)) {
```
KaylaBrady, line 41: suggestion (non-blocking): Might be worth renaming `requiresStaging` at this point too. We could invert to `requiresProd`

fastlane/Fastfile
```diff
@@ -34,10 +36,17 @@ platform :android do
...
-          package_name: package_name,
-          track: 'internal'
-        ).max
+        begin
+          google_play_track_version_codes(
+            package_name: package_name,
+            track: 'internal'
+          ).max
+        rescue Exception => e
+          unless e.message == "Google Api Error: Invalid request - Package not found: #{package_name}."
```
KaylaBrady, line 45: question (non-blocking): Is this just temporary handling until the app is added?

fastlane/Fastfile
```diff
@@ -188,6 +199,13 @@ platform :ios do
     app_store_connect
     begin
       case options[:scheme]
+      when 'DevOrange'
+        upload_to_testflight(
+          ipa: 'iosApp/build/iosApp.ipa',
+          distribute_external: true,
+          changelog: '',
+          groups: ['Dev Orange Users']
```
KaylaBrady, line 207: question(non-blocking): Is there a reason we'd want to manage this group separately from our existing groups? Wondering if we only need one group with access to all non-prod apps for simplicity

KaylaBrady (approved):
Sweet!

boringcactus:
I figured separate labels would be easier, but apparently adding two labels to a PR at the same time is simultaneously non-atomic (each workflow runs twice) and atomic (all four runs see both labels as having already been added). If we have a combined label, the logic for the scheduled deploys gets a lot more complicated, so it’s probably easier to make the second runs cancel the first runs.

.github/workflows/android-dev-orange-deploy.yaml
boringcactus: I picked these times arbitrarily, and then I picked the backend times with more attention to normal working hours just in case something goes wrong, and then I did not update these to match that.

fastlane/Fastfile
boringcactus: This required a temporary workaround when we created the staging app, it requires a temporary workaround now, and it will require a temporary workaround in the not inconceivable future where we want dev-silver. I don’t think it increases the maintenance burden to leave it in long-term.

fastlane/Fastfile
boringcactus: It doesn’t look like TestFlight groups carry over from one app to another.

PR #856, opened 2025-03-28
feat(RouteCard): Pin button with analytics

KaylaBrady:
### Summary

_Ticket:_ [Group by direction | 🤖 Add back pinning](https://app.asana.com/0/1205732265579288/1209786469572857/f)

What is this PR for?
Adds the current favorites/pinning behavior to the group by direction route cards.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally, confirmed pinning / unpinning reflected as expected
Screenshot_20250328_162733
Screenshot_20250328_162827


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/routeCard/DeparturesTest.kt
```diff
@@ -141,10 +148,100 @@ class DeparturesTest {
...
+        var onClickCalled = false
+
+        val koinApplication =
+            testKoinApplication(
+                analytics = MockAnalytics({ event, props -> tapAnalytics = Pair(event, props) })
+            )
+
+        composeTestRule.setContent {
+            KoinContext(koinApplication.koin) {
+                Departures(stopData, cardData, now, pinned = true) { onClickCalled = true }
```
KaylaBrady, line 225: I didn't catch in initial review, but adding this test, it felt a little funny passing stopData & cardData both when cardData contains stopData. @EmmaSimon what do you think about passing only in the cardData props that are needed - context & lineOrRoute?

boringcactus (approved):

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/routeCard/DeparturesTest.kt
KaylaBrady: Could point on the extra props. Maybe we could make a separate struct to represent all the route metadata so that it is all bundled & easy to pass separately

EmmaSimon: Yeah I was thinking about that too, but at least right now, when we only have 3 fields, it seems a bit overkill. 🤷 Can you think of any stuff we're likely to add on the top level? I was thinking that favorites will be included, but those should all be on a lower level.

PR #866, opened 2025-04-02
feat: detect service ended in group by direction

boringcactus:
### Summary

_Ticket:_ [Group by direction | 🤖 Special service states](https://app.asana.com/0/1205732265579288/1209536961544714/f)

This was not a two point task.

iOS

android

### Testing

Added new unit tests to ensure that leaves with the correct properties are generated when there is no service, service only earlier in the day, or service scheduled for later in the day (which causes "predictions unavailable" when we don’t just show the service). Checked that there are already unit tests to ensure that leaves with the correct properties are formatted in the correct way. Manually verified that on the Green Line near the start of trips the message is “predictions unavailable” rather than “no service today” as was previously the case.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -525,23 +519,32 @@ data class RouteCardData(
...
+                        ::mutableListOf
+                    )
                     .add(upcomingTrip)
             }
 
-            this.updateLeaves<List<UpcomingTrip>>(partialHierarchy) { leafBuilder, upcomingTrips ->
-                leafBuilder.upcomingTrips = upcomingTrips
+            val hasSchedulesTodayByPattern = NearbyStaticData.getSchedulesTodayByPattern(schedules)
+
+            forEachLeaf { routeOrLineId, stopId, directionId, leafBuilder ->
```
boringcactus, line 539: We need to visit all of the leaves, not just the ones which have upcoming trips, so I’ve turned `updateLeaves` into `forEachLeaf` and flattened the `partialHierarchy`. Defining a `data class HierarchyPath` instead of using `Triple` might be a little bit more legible, but unless we start wanting it somewhere else it’s almost certainly not worth it.

EmmaSimon: I think I'd actually prefer creating a data class here, just looking at the triple definition, it's unclear what the values represent, and then you could just create the `HierarchyPath` in `forEachLeaf` and pass it into `process` as a single param. It seems a little unnecessarily verbose to pass all three of those in separately when they're only ever accessed to create the triple. But that's all nitpicking that I don't feel incredibly strongly about.

EmmaSimon (approved):
Just the one nitpick, otherwise lgtm!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
boringcactus: Created.

PR #868, opened 2025-04-03
test(Android): fix some flaky tests, hopefully

boringcactus:
### Summary

_Ticket:_ [[flaky android tests again]](https://app.asana.com/0/1205732265579288/1209645505139426/f)

I’m choosing to blame the StopDetailsViewModelTest failures on runTest - runTest is [suggested in the Android docs](https://developer.android.com/kotlin/coroutines/test), but not in a UI test context, and since we have access to `composeTestRule.waitForIdle()` instead, I think it’s a more realistic test if we don’t mess with the dispatchers at all. I’ve banished `runTest` throughout the tests so we never run into this issue again.

The MorePageTests failures don’t make a ton of sense to me, and I was having trouble reproducing them, so I’m just tossing in an extra `waitForIdle` and hoping that’s enough.

iOS

android

### Testing

Ran the problematic tests a few times and never saw failures.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
@@ -1322,6 +1322,8 @@ class StopDetailsViewModelTest {
 
         val expectedTripFilter = TripDetailsFilter(trip1.id, null, 0, false)
 
+        composeTestRule.waitForIdle()
+
         composeTestRule.waitUntil(2_000) { newTripFilter == expectedTripFilter }
         kotlin.test.assertEquals(expectedTripFilter, newTripFilter)
     }
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
@@ -40,9 +40,6 @@ import kotlin.test.assertNull
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.StandardTestDispatcher
-import kotlinx.coroutines.test.advanceUntilIdle
-import kotlinx.coroutines.test.runTest
 import kotlinx.datetime.Clock
 import org.junit.Rule
 import org.junit.Test
@@ -52,7 +49,7 @@ class StopDetailsViewModelTest {
     @get:Rule val composeTestRule = createComposeRule()
 
     @Test
-    fun testLoadStopDetailsLoadsSchedulesAndPredictions() = runTest {
+    fun testLoadStopDetailsLoadsSchedulesAndPredictions() {
         val objects = ObjectCollectionBuilder()
         objects.prediction {}
         objects.trip {}
@@ -72,7 +69,7 @@ class StopDetailsViewModelTest {
     }
 
     @Test
-    fun testRejoinStopPredictionsWhenStopDataSet() = runTest {
+    fun testRejoinStopPredictionsWhenStopDataSet() {
         val objects = ObjectCollectionBuilder()
         objects.prediction {}
         objects.trip {}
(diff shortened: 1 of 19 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
```diff
@@ -902,24 +883,21 @@ class StopDetailsViewModelTest {
...
-            // In resumed state, so joined 1 time for stop, 1 time for resume.
-            // Need to start with lifecycle resumed in order to test pause
-            stopPredictionsConnectCount == 2
-
-            // Trip channels rejoin is no-op because tripData hasn't been set yet
-            && tripPredictionsConnectCount == 1 && vehicleConnectCount == 1
+            // resume rejoin will have been skipped
+            stopPredictionsConnectCount == 1 &&
+                tripPredictionsConnectCount == 1 &&
+                vehicleConnectCount == 1
```
boringcactus, line 897: “tripData hasn’t been set yet” wasn’t actually guaranteed to be true: if the `clearAndLoadTripDetails` call finished before the `LifecycleResumeEffect` kicked in, tripData would in fact have been set, causing the trip channels to be joined a second time as well.

It seemed to me like the easiest approach here was to just not rejoin the channels in the `LifecycleResumeEffect` until we’ve actually been to the background first, so that it doesn’t matter when the `LifecycleResumeEffect` wakes up; this race condition can only hit if `loadTripDetails` is instant, which it won’t be in the real app, so this was only ever a test issue, but the new behavior still feels more correct on principle.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
```diff
@@ -534,15 +537,23 @@ fun stopDetailsManagedVM(
...
+
     LaunchedEffect(stopId) { viewModel.handleStopChange(stopId) }
     LaunchedEffect(filters?.tripFilter) { viewModel.handleTripFilterChange(filters?.tripFilter) }
 
     LifecycleResumeEffect(null) {
-        viewModel.returnFromBackground()
-        viewModel.rejoinStopPredictions()
-        viewModel.rejoinTripChannels()
+        // we don’t want to return and rejoin on first appearance, only on actual resume, because
+        // the handleStopChange/handleTripFilterChange will handle initial joins
```
KaylaBrady, line 547: 👍 thought (non-blocking): I can't remember why I initially created separate `rejoinStopPredictions` and `rejoinTripChannels` functions which don't have an equivalent in the iOS implementation. It seems possible that condensing these three effects into a single `LifecycleResumeEffect` that takes stopId and `filters?.tripFilter` could help to simplify things here and make it less error prone.

KaylaBrady (approved):
Good call that runTest wasn't recommended in a UI testing context where I initially reached for it. Glad this seems to help!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
```diff
@@ -534,15 +537,23 @@ fun stopDetailsManagedVM(
     val groupByDirection by viewModel.groupByDirection.collectAsState()
     val departures by viewModel.stopDepartures.collectAsState()
 
+    var wasInBackground by remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
boringcactus: Good catch, it probably should.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
```diff
@@ -1387,8 +1322,6 @@ class StopDetailsViewModelTest {
 
         val expectedTripFilter = TripDetailsFilter(trip1.id, null, 0, false)
 
-        advanceUntilIdle()
-
         composeTestRule.waitUntil(2_000) { newTripFilter == expectedTripFilter }
```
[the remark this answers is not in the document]
KaylaBrady: 👍 yeah I'd expect so too... I'm somewhat wary of why that waitUntil needed a long hardcoded duration & am hoping we can effectively do away with it by waiting for idle before it.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt
boringcactus: It’d be a little weird to be calling `handleStopChange` if the stop filter changes but the trip filter doesn‘t, but I think having separate `LifecycleResumeEffect`s for the stop and the trip Should™ work.

PR #869, opened 2025-04-04
fix(iOS): update Sentry pod to fix Xcode 16.3 issue

boringcactus:
### Summary

_Ticket:_ none

I let my Xcode install an update (rookie mistake, I know), and Sentry stopped building on iOS due to https://github.com/getsentry/sentry-cocoa/issues/4886. The fix for that is included in sentry-cocoa 8.46+.

iOS

android

### Testing

Verified that the app now builds in Xcode 16.3.

### DIFF
--- a/shared/build.gradle.kts
+++ b/shared/build.gradle.kts
@@ -436,5 +436,9 @@ mokkery {
 
 sentryKmp {
     autoInstall.commonMain.enabled = false
-    autoInstall.cocoapods.sentryCocoaVersion.set("~> 8.46.0")
+    if (autoInstall.cocoapods.sentryCocoaVersion.getOrElse("") == "~> 8.44.0") {
+        autoInstall.cocoapods.sentryCocoaVersion.set("~> 8.46.0")
+    } else {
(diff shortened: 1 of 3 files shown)


shared/build.gradle.kts
```diff
@@ -434,4 +434,11 @@ mokkery {
...
 }
 
-sentryKmp { autoInstall.commonMain.enabled = false }
+sentryKmp {
+    autoInstall.commonMain.enabled = false
+    if (autoInstall.cocoapods.sentryCocoaVersion.getOrElse("") == "~> 8.44.0") {
+        autoInstall.cocoapods.sentryCocoaVersion.set("~> 8.46.0")
+    } else {
+        throw IllegalStateException("sentry-kmp was updated, delete explicit sentry-cocoa version")
+    }
```
boringcactus, line 443: If we never remove `.set("~> 8.46.0")`, we’ll never upgrade past 8.46.0 even once sentry-kmp suggests a higher version. I can’t make the code actually delete itself, but failing all the CI on that eventual dependabot PR should remind us to take a look

BrandonTR (approved):

PR #871, opened 2025-04-07
fix(iOS): widen focus area for stop and route names in nearby transit

boringcactus:
### Summary

_Ticket:_ [Fix: VoiceOver bugs from research session](https://app.asana.com/0/1205732265579288/1209584641001346/f)

Makes the full width available to these text fields tappable to focus them.

iOS

android

### Testing

Manually verified that the full available width is now tappable.

### DIFF
--- a/iosApp/iosApp/ComponentViews/TransitHeader.swift
+++ b/iosApp/iosApp/ComponentViews/TransitHeader.swift
@@ -27,7 +27,7 @@ struct TransitHeader<Content: View>: View {
                 .foregroundStyle(textColor)
                 .textCase(.none)
                 .frame(maxWidth: .infinity, maxHeight: modeIconHeight, alignment: .leading)
-                .contentShape(.rect)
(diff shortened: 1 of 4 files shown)


iosApp/iosApp/ComponentViews/TransitHeader.swift
```diff
@@ -27,6 +27,7 @@ struct TransitHeader<Content: View>: View {
                 .foregroundStyle(textColor)
                 .textCase(.none)
                 .frame(maxWidth: .infinity, maxHeight: modeIconHeight, alignment: .leading)
+                .contentShape(.rect)
```
boringcactus, line 30: This might be one of the least intuitive fixes I’ve ever implemented. I’d expect the content shape to already be a rectangle, but apparently setting it to be a rectangle also makes its size cover the full `.frame()` size.

KaylaBrady (approved):

PR #876, opened 2025-04-07
feat(Android): derive departure tiles from new RouteCardData

boringcactus:
### Summary

_Ticket:_ [[make filtered stop details use new data structures]](https://app.asana.com/0/1205732265579288/1209885620137455/f)

Stacked on #872.

This would be way simpler if we were just moving from the old structure to the new structure, but since we need to keep both working with the feature flag, it’s a bit rough. The tests especially suffer there; if we want to make sure everything works with both configurations, the only straightforward approach is to actually run every test under both configurations, and that takes twice as much code.

iOS

android

### Testing

Checked that tiles look reasonable and have started showing headsigns on the RL southbound right now while one branch is closed and so the other branch has all the trips. Added new tests for new functionality and copied existing tests to prevent regressions under the feature flag.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
```diff
@@ -96,21 +100,147 @@ fun StopDetailsFilteredDeparturesView(
...
+    errorBannerViewModel: ErrorBannerViewModel,
+    updateStopFilter: (StopDetailsFilter?) -> Unit,
+    updateTripFilter: (TripDetailsFilter?) -> Unit,
+    tileScrollState: ScrollState,
+    pinnedRoutes: Set<String>,
+    togglePinnedRoute: (String) -> Unit,
+    onClose: () -> Unit,
+    openModal: (ModalRoutes) -> Unit,
+    openSheetRoute: (SheetRoutes) -> Unit,
+    analytics: Analytics = koinInject()
```
boringcactus, line 230: This is far, far too many parameters for one method, but it’s not that many more parameters than this component already had, and I can’t think of obvious ways to group a bunch of them together at once in a way that’s actually cohesive and not just cosmetic.

boringcactus: I just wasn’t creative enough; most of these parameters are the same in either variant, so putting the ones that actually differ into a `sealed interface` lets me keep the parameter count the same and get rid of the inner helper function completely.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesViewTest.kt
```diff
@@ -136,44 +138,87 @@ class StopDetailsFilteredDeparturesViewTest {
...
-                )
+        val filterState = StopDetailsFilter(routeId = route.id, directionId = 0)
+        val viewModel =
+            StopDetailsViewModel.mocked(
+                settingsRepository =
+                    MockSettingsRepository(mapOf(Settings.GroupByDirection to groupByDirection))
             )
-        val viewModel = StopDetailsViewModel.mocked()
-        viewModel.setDepartures(departures)
+        val data: FilteredDeparturesData
```
EmmaSimon, line 157: I didn't realize that you could initialize vals without immediately setting a value like this, that's good to know!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
```diff
@@ -55,13 +57,23 @@ import kotlinx.coroutines.launch
 import kotlinx.datetime.Instant
 import org.koin.compose.koinInject
 
+sealed interface FilteredDeparturesData {
+    data class PreGroupByDirection(val patternsByStop: PatternsByStop) : FilteredDeparturesData
+
+    data class PostGroupByDirection(
+        val routeCardData: RouteCardData,
+        val routeStopData: RouteCardData.RouteStopData,
+        val leaf: RouteCardData.Leaf
```
EmmaSimon, line 66: I don't think it's worth doing as part of this PR, but it seems like we're often bundling the parent and `RouteCardData`, `StopData`, and `Leaf`s together since they all have different context that the components need (like how `Departures` takes both `RouteCardData` and `StopData`). What do you think about including the `RouteCardData` used to create each `StopData` within the `StopData`, and the `StopData` used to create each `Leaf` in the `Leaf`? I can see that making testing a bit more painful, but it would require a bit less argument slinging overall. We could maybe tie that into the iOS implementation.

EmmaSimon (approved):
Nice, I like this approach, it'll be pretty straightforward (and satisfying) to remove the old codepaths once we switch over.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
boringcactus: I’m concerned that it might be kind of a mess to try to actually construct all those circular references at the same time, but it’s probably worth trying at some point in the future, yeah.

PR #878, opened 2025-04-08
fix: prevent dashes from overlapping

boringcactus:
### Summary

_Ticket:_ [Fix: Overlapping Green Line branch alerts don't consistently appear dashed](https://app.asana.com/0/1205732265579288/1208468599574710/f)

Per @KaylaBrady’s advice, this PR rolls back the changes that were made in https://github.com/mbta/mobile_app/pull/159, creating new layers dynamically to ensure that alerting route backgrounds and alerting route foregrounds are properly layered.

|Platform|Before|After|
|--------|------|-----|
|iOS|ios before|ios after|
|Android|android before|android after|

iOS

android

### Testing

Checked that this works even when selecting and deselecting stops and bus routes. Updated unit tests to expect new behavior.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapLayerManager.kt
```diff
@@ -44,12 +48,56 @@ class MapLayerManager(val map: MapboxMap, context: Context) {
...
+            }
+        val stopLayers = StopLayerGenerator.createStopLayers(colorPalette).map { it.toMapbox() }
+        setLayers(routeLayers, stopLayers)
+    }
+
+    private suspend fun setLayers(routeLayers: List<MapboxLayer>, stopLayers: List<MapboxLayer>) {
         withContext(Dispatchers.Main) {
-            for (layer in layers) {
+            if (!map.styleLayerExists(bufferLayerId)) {
+                val layer = SlotLayer(bufferLayerId)
```
boringcactus, line 83: Inserting an empty layer with a known ID between the route layers and the stop layers prevents routes from being added on top of stops, which was happening before I did that, but it doesn’t technically guarantee that individual routes won’t be drawn in the wrong order if one already existed and another one is being added later. If we decide it’s worth sinking unbounded engineer time into route sorting on the map, we could probably loop back through after we’ve added all the layers to make sure that everything is in the order it’s supposed to be, but I’d just as soon not deal with that.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapLayerManager.kt
```diff
@@ -44,12 +48,56 @@ class MapLayerManager(val map: MapboxMap, context: Context) {
...
+        globalResponse: GlobalResponse,
+        colorPalette: ColorPalette
+    ) {
+        val routeLayers =
+            RouteLayerGenerator.createAllRouteLayers(routes, globalResponse, colorPalette).map {
+                it.toMapbox()
+            }
+        val stopLayers = StopLayerGenerator.createStopLayers(colorPalette).map { it.toMapbox() }
+        setLayers(routeLayers, stopLayers)
+    }
```
EmmaSimon, line 78: I feel a bit weird about the duplication here, why not have a single `RouteLayerGenerator.createAllRouteLayers` function with the below signature and do all the overloading here?

```kotlin
suspend fun createAllRouteLayers(
    routesWithShapes: List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>,
    routesById: Map<String, Route>,
    colorPalette: ColorPalette
): List<LineLayer>
```

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteLayerGenerator.kt
```diff
@@ -4,29 +4,63 @@ import com.mbta.tid.mbta_app.map.style.Exp
...
-            listOf(createRouteLayer()) +
-                // Draw all alerting layers on top so they are not covered by any overlapping route
-                // shape
-                createAlertingRouteLayers(colorPalette)
+            createAllRouteLayers(routesWithShapes, globalResponse.routes, colorPalette)
         }
 
-    fun createRouteLayer(): LineLayer {
-        val layer = baseRouteLayer(routeLayerId)
+    private fun createAllRouteLayers(
```
EmmaSimon, line 41: Why not have this use `withContext(Dispatchers.Default)` in here rather than putting it in the overloading method?

EmmaSimon (commented):
A couple questions, but overall looks solid

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapLayerManager.kt
boringcactus: If there are routes that we draw at the same time that run close enough to overlap, we might draw the wrong one on top if you’re transitioning from a state that only draws the one that should be higher to a state that draws both. Also, at intersections between routes, we might draw the wrong route on top, but that doesn’t matter.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapLayerManager.kt
boringcactus: We can’t do that because the `globalData.routes` is declared as `internal` in GlobalResponse (as of #839), but we can at least narrow down the number of options available at any given step in the process. As part of trying to do that on iOS, I discovered that the iOS version doesn’t actually remove layers properly yet for some reason, so I’ll need to fix that before this PR is ready to merge.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteLayerGenerator.kt
boringcactus: I don’t think it matters since this is `private` anyway, but I can move the `withContext(Dispatchers.Default)` in here if you prefer it that way.

EmmaSimon: Yeah I was thinking of this in concert with the other question, but I still think it makes the most sense to have the `withContext` around the actual work rather than a call to the work.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/MapLayerManager.kt
boringcactus: The iOS issue traced back to a `continue` getting replaced with a `return` in #489.

EmmaSimon (approved):

PR #882, opened 2025-04-10
feat(

KaylaBrady:
### Summary

_Ticket:_ [Update elevator & map settings in onboarding flow](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209875810647997?focus=true)

What is this PR for?

This PR updates the "Hide Maps" Onboarding page + setting to be "Map Display". I played around with inverting the underlying setting at the SettingsRepository level, but that ended up being a way bigger change, so I stuck with just inverting it in the relevant views.

iOS

android

### Testing

What testing have you done?
* Updated unit tests
* Ran locally
| desc | ios | android |
|--------|--------|--------|
| default onboarding | IMG_0270 | Screenshot_20250410_173629 |
| onboarding toggled | IMG_0271 | Screenshot_20250410_173655 |
| settings page | IMG_0272 | Screenshot_20250410_173709 |


androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
```diff
@@ -149,25 +153,31 @@ fun OnboardingScreenView(
                 }
             }
         }
-        OnboardingScreen.HideMaps -> {
+        OnboardingScreen.MapDisplay -> {
+            var localHideMapsSetting by rememberSaveable { mutableStateOf(true) }
```
KaylaBrady, line 157: I don't _love_ this being stored locally here, but it is more awkward on iOS. storing a local value & saving on continue seemed like the best approach to me since we want the toggle defaulted to the off position, which is the opposite of the actual default setting value. Definitely welcome any suggestion for a cleaner approach!

iosApp/iosAppTests/Pages/Onboarding/OnboardingScreenViewTests.swift
```diff
@@ -59,23 +59,29 @@ final class OnboardingScreenViewTests: XCTestCase {
...
 
-    func testHideMapsFlow() throws {
+    @MainActor
+    func testMapDisplayFlow() throws {
         let saveSettingExp = expectation(description: "saves hide maps setting")
-        let settingsRepo = MockSettingsRepository(settings: [.hideMaps: false], onSaveSettings: {
-            XCTAssertEqual($0, [.hideMaps: true])
+        let settingsRepo = MockSettingsRepository(settings: [.hideMaps: false], onSaveSettings: { _ in
+            //    XCTAssertEqual($0, [.hideMaps: false])
+            //    assertion fails, seemingly because of the use of localHideMaps as local state for the view
```
KaylaBrady, line 67: This particularly I wasn't happy about w/ the use of localHideMaps in the view's state.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/OnboardingScreen.kt
```diff
@@ -5,14 +5,14 @@ import com.mbta.tid.mbta_app.repositories.IAccessibilityStatusRepository
 enum class OnboardingScreen {
     Location,
     StationAccessibility,
-    HideMaps,
+    MapDisplay,
```
boringcactus, line 8: This will crash the app on launch for people who have already gone through this screen, because we don’t handle errors in `OnboardingScreen.valueOf` when the `OnboardingRepository` is loading the completed screens.

androidApp/src/main/res/values/strings.xml
```diff
@@ -198,16 +198,14 @@
     <string name="onboarding_feedback_advance">Get started</string>
     <string name="onboarding_feedback_body">MBTA Go is in the early stages! We want your feedback as we continue making improvements and adding new features.</string>
     <string name="onboarding_feedback_header">Help us improve</string>
-    <string name="onboarding_hide_maps_body">When using TalkBack, we can skip reading out maps to keep you focused on transit information.</string>
-    <string name="onboarding_hide_maps_header">Set map preference</string>
-    <string name="onboarding_hide_maps_hide">Hide maps</string>
-    <string name="onboarding_hide_maps_show">Show maps</string>
     <string name="onboarding_location_advance">Continue</string>
```
boringcactus, line 201: If we want all the “Continue”s to be the same, we should get rid of this one, too.

androidApp/src/main/res/values/strings.xml
```diff
@@ -198,16 +198,14 @@
...
     <string name="onboarding_location_body">We use your location to show you nearby transit options.</string>
     <string name="onboarding_location_footer">You can always change location settings later in the Settings app.</string>
     <string name="onboarding_location_header">See transit near you</string>
+    <string name="onboarding_map_display_body">When using TalkBack, we can hide maps to make the app easier to navigate.</string>
+    <string name="onboarding_map_display_header">Set map display preference</string>
     <string name="onboarding_station_accessibility_body">By opting in, we can show you which stations are inaccessible or have elevator closures.</string>
-    <string name="onboarding_station_accessibility_continue">Continue</string>
+    <string name="onboarding_continue">Continue</string>
```
boringcactus, line 208: This is not alphabetized correctly.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
boringcactus: The only other option I can think of would be setting the setting to true when the screen first appears, but if that’s not instant and there’s a perceptible moment where the switch is on before it flips to off, that’s not worth doing.

iosApp/iosAppTests/Pages/Onboarding/OnboardingScreenViewTests.swift
boringcactus: It might work if you host the view and test it via that flow, rather than just asserting on an unmounted view.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/onboarding/OnboardingScreenView.kt
KaylaBrady: I tried that first w/ a LaunchedEffect and there was a flicker, will stick with this then!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/OnboardingScreen.kt
KaylaBrady: Oh Good catch. I may just leave it HideMaps then.

iosApp/iosAppTests/Pages/Onboarding/OnboardingScreenViewTests.swift
KaylaBrady: Yes, thank you!! I had looked at [Approach #2](https://github.com/nalexn/ViewInspector/blob/0.10.2/guide.md#approach-2) in the View Inspector guide but skipped the ever important call to `ViewHosting.host(view: sut)`

boringcactus (approved):

PR #883, opened 2025-04-10
fix: trip display on disrupted branches

BrandonTR:
### Summary

_Ticket:_ [Group by direction | Add in alerts data](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209536961544708?focus=true)

This attempts to solve for [this case in figma](https://www.figma.com/design/eFiLM8xjo9hKprg93BBCtZ/App-UI-Design?node-id=9292-342621&t=SMidgUdFqKMUuc2u-4).

iOS

android

### Testing

Filled in a unit test that tests this scenario.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -190,6 +187,51 @@ data class RouteCardData(
...
+                    tripsToShow.map { (trip, format) ->
+                        val route =
+                            if (needsRoutesInBranching) globalData?.getRoute(trip.trip.routeId)
+                            else null
+                        LeafFormat.Branched.Branch(
+                            route,
+                            trip.headsign,
+                            UpcomingFormat.Some(
+                                format,
+                                UpcomingFormat.SecondaryAlert(StopAlertState.Issue, mapStopRoute)
```
KaylaBrady, line 222: suggestion (non-blocking): I think we don't need to include the alert here after all, based on this [alert case](https://miro.com/app/board/uXjVIDh-CTs=/?moveToWidget=3458764624427852689&cot=14).

KaylaBrady (approved):
Approving as a step forward since this does improve the behavior for branching stops by showing non-disrupted branches! We can continue to iterate on this based on the follow-up items from the miro board discussion as separate tickets.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
BrandonTR: Would we want to just pass in null? I assume that would leave us without any indication of a disruption in the interim

KaylaBrady: Yeah I think null would be OK,  but also totally fine to leave as-is for the interim state since this will be refined in the upcoming [disruption on a branch](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209979069137605?focus=true) task

PR #892, opened 2025-04-15
feat(ios): Templatized alerts in filtered stop details

KaylaBrady:
### Summary

_Ticket:_ [ | Alert summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209965144946667?focus=true)

What is this PR for?
iOS counterpart to using alert summary templates based on android implementation https://github.com/mbta/mobile_app/pull/877

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally for a handful of alerts
IMG_0275IMG_0275IMG_0277


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -140,9 +143,28 @@ struct StopDetailsFilteredDepartureDetails: View {
         .onChange(of: tripFilter) { tripFilter in
             selectedDepartureFocus = tiles.first { $0.id == tripFilter?.tripId }?.id ?? cardFocusId
         }
+        .onChange(of: patternsByStop) { patternsByStop in
+            patternsHere = patternsHere(patternsByStop)
+        }
+        .onAppear { setAlertSummaries() }
+        .onChange(of: stopDetailsVM.global) { _ in setAlertSummaries() }
```
KaylaBrady, line 150: I do not like this string of `onChange` - I tried putting them all in a single `onChange(of: [stopDetailsVM.global, ...])`, but got a typecheck failure.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -140,9 +143,28 @@ struct StopDetailsFilteredDepartureDetails: View {
...
+        .onChange(of: downstreamAlerts) { _ in setAlertSummaries() }
+        .onChange(of: stopId) { _ in setAlertSummaries() }
+        .onChange(of: stopFilter.directionId) { _ in setAlertSummaries() }
+        .onChange(of: patternsHere) { _ in setAlertSummaries() }
+        .onChange(of: now) { _ in setAlertSummaries() }
         .ignoresSafeArea(.all)
     }
 
+    func patternsHere(_ patternsByStop: PatternsByStop) -> [RoutePattern] {
+        // RealtimePatterns.patterns is a List<RoutePattern?> but that gets bridged as [Any] for some reason
```
KaylaBrady, line 161: Pulled comment from another place we are doing this. This will go away once we aren't using PatternsByStop anymore.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
EmmaSimon: I ran into this same thing in #893 and worked around it by [hashing all the requirements together](https://github.com/mbta/mobile_app/pull/893/files#diff-a02e18ebdb7fec9ec2891eb904127b4db46b407ddc37d170333bd776f1448113R39-R51). This isn't a perfect solution, I think we'll run into problems if we have multiple of these in one component and need to use `onChange` in tests, since the resolved value for both with be `Int`. It would be cool if SwiftUI supported this extremely common use case in a way that didn't feel like a hack.

KaylaBrady: Yeah I can do that! I ended up not testing this directly and have instead added more tests to `AlertCard` for the formatted alert strings, but I do like the idea of using a struct to bundle these together rather than hashing

iosApp/iosApp/ComponentViews/AlertCard.swift
```diff
@@ -38,24 +43,11 @@ struct AlertCard: View {
...
-
-    func delayHeader(_ formattedAlert: FormattedAlert) -> String {
-        if let cause = formattedAlert.dueToCause {
-            String(format: NSLocalizedString(
-                "Delays due to %@",
-                comment: "Describes the cause of a delay. Ex: 'Delays due to [traffic]'"
-            ), cause)
-        } else {
-            NSLocalizedString("Delays", comment: "Generic delay alert label when cause is unknown")
+    var headerFont: Typography {
```
KaylaBrady, line 46: TODO: Check this. We probably dont' want calloutSemibold if there is an alert summary to display, only if it is a fallback

KaylaBrady: This is handled in https://github.com/mbta/mobile_app/pull/898

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
boringcactus: The reason is that generics over optionals aren’t possible in Objective-C and therefore get lost in Swift. I wish frequently that Kotlin had actual Swift interop.

iosApp/iosApp/Utils/AttributedStringExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+//  iosApp
+//
+//  Created by Kayla Brady on 4/14/25.
+//  Copyright © 2025 MBTA. All rights reserved.
+//
+
+import SwiftUI
+
+extension AttributedString {
+    static func boldOrDefault(_ stringWithMarkdown: String) -> AttributedString {
```
boringcactus, line 12: I find this function name confusing: it doesn’t make any text bold, it just renders Markdown that may call for some text to be bold. `tryMarkdown` might make more sense.

KaylaBrady: I like that way better, TY!

iosApp/iosApp/ComponentViews/AlertCard.swift
KaylaBrady: Backed out this change

boringcactus (approved):

PR #903, opened 2025-04-18
ci: run android/ios deploy when no platform prefix

KaylaBrady:
### Summary

_Ticket:_ [Update CI to deploy both platforms when we use a certain tag format](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209633379255738?focus=true)

What is this PR for?
Adds new tag matching condition for ios + android to deploy to both platforms at once. 
Referenced this SO post and [GitHub Actions docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

iOS

android

### Testing

What testing have you done?
* Have not tested, but we can give it a try for our next release & I'll come back to edit as needed.

### DIFF
--- a/.github/workflows/android-prod-deploy.yaml
+++ b/.github/workflows/android-prod-deploy.yaml
@@ -4,7 +4,7 @@ on:
   push:
     tags:
       - android-*
-      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+*'
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/android-prod-deploy.yaml
+++ b/.github/workflows/android-prod-deploy.yaml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - android-*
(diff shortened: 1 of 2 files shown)


.github/workflows/android-prod-deploy.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - android-*
+      - '[0-9]+.[0-9]+.[0-9]+*'
```
KaylaBrady, line 7: very rarely have we added a `-X` suffix to the version number to re-deploy a tag, so added the extra `*` just in case.

EmmaSimon (approved):

boringcactus (approved):
The fact that `[]` and `+` and `?` mean regular expression things but `*` and `**` mean shell glob things and `.` has no special meaning is completely horrifying, and would be a strong contender for my least favorite GitHub Actions quirk were it not for [`'false'` maybe being truthy](https://github.com/mbta/mobile_app_backend/pull/302#issuecomment-2762286656).

PR #909, opened 2025-04-23
feat(iOS): show downstream alert cards in trip details

boringcactus:
### Summary

_Ticket:_ [Display downstream shuttles & suspensions in trip details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209527076283977?focus=true)

The iOS version of #900.

iOS

android

### Testing

Added new unit tests for new functionality. Confirmed that existing unit tests aren’t broken by this (none of the relevant ones appear to be broken by Xcode 16.3).

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
@@ -45,6 +45,7 @@ import com.mbta.tid.mbta_app.android.util.Typography
 import com.mbta.tid.mbta_app.android.util.modifiers.haloContainer
 import com.mbta.tid.mbta_app.android.util.typeText
 import com.mbta.tid.mbta_app.model.Alert
+import com.mbta.tid.mbta_app.model.AlertSignificance
 import com.mbta.tid.mbta_app.model.AlertSummary
 import com.mbta.tid.mbta_app.model.ObjectCollectionBuilder
 import com.mbta.tid.mbta_app.model.RouteType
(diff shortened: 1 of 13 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
```diff
@@ -101,7 +102,16 @@ fun TripStops(
...
+                .padding(horizontal = 4.dp)
+                .padding(
+                    bottom =
+                        if (
+                            stops.stops.lastOrNull()?.disruption?.alert?.significance ==
+                                AlertSignificance.Major
+                        )
+                            4.dp
+                        else 0.dp
+                )
```
boringcactus, line 114: Including padding at the bottom for the background makes sense if the list ends with an alert card, but does not make sense if the list ends with a stop with no alert card.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
```diff
@@ -238,7 +248,7 @@ fun TripStops(
             }
             StopList(
                 splitStops.followingStops,
-                lastStopSequence?.plus(if (splitStops.isTruncatedByLastAlert) 1 else 0),
+                lastStopSequence,
```
boringcactus, line 251: During an earlier revision of #900 I was truncating the stops when the TripDetailsStopList was first being constructed, and when I was doing that this was necessary to distinguish a truncated stop from an actual last stop, but since I’m truncating in splitForTarget instead, this isn’t needed.

EmmaSimon (approved):

PR #911, opened 2025-04-23
chore(Android): back out downstream alert cards in trip details

boringcactus:
### Summary

_Ticket:_ [Display downstream shuttles & suspensions in trip details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209527076283977?focus=true) (kinda)

Since we want to get the next release out the door before we can QA #900 and #909, this is the lowest-footprint change that rolls back the effect of #900 in a way that’ll be easy to undo later.

iOS

android

### Testing

Inverted the test that checks for downstream alerts being shown. Verified that downstream alerts are no longer shown in trip details.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
@@ -75,7 +75,7 @@ fun TripStops(
 
     val splitStops: TripDetailsStopList.TargetSplit =
         remember(targetId, stops, stopSequence, global) {
-            stops.splitForTarget(targetId, stopSequence, global, truncateForDisruptions = true)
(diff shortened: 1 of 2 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripStops.kt
```diff
@@ -101,7 +101,7 @@ fun TripStops(
     Box {
         Box(
             Modifier.matchParentSize()
-                .padding(4.dp)
+                .padding(horizontal = 4.dp)
```
boringcactus, line 104: This is a mistake I fixed more thoroughly in #909 - the padding on the bottom is only correct if there’s an alert card at the bottom of the list, but since there can’t be an alert card in the list as of this PR, just never having the padding at the bottom is the right fix for now.

KaylaBrady (approved):
Thanks! Appreciate how small this diff is!

PR #912, opened 2025-04-23
ci: tweak a few things in response to merge queue

boringcactus:
### Summary

_Ticket:_ [Try out GitHub merge queue](https://app.asana.com/1/15492006741476/project/1205991273035889/task/1209644129662966?focus=true)

Merge queue is working pretty nicely, but we don’t need to run the tests again as part of staging deploys, and the concurrency protections were set up before merge queue and contained some assumptions that aren’t valid anymore.

iOS

android

### Testing

Not tested.

### DIFF
--- a/.github/workflows/build-android.yaml
+++ b/.github/workflows/build-android.yaml
@@ -34,6 +34,11 @@ on:
       SENTRY_DSN_ANDROID:
         required: true
 
+concurrency:
+  # non-deploy builds shouldn’t conflict, but builds for deployment can’t overlap or they’ll claim the same version number
+  group: build-android-${{ inputs.flavor }}${{ inputs.deploy && '' || github.run_id }}
+  cancel-in-progress: false
+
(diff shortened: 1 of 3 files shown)


EmmaSimon (approved):

.github/workflows/build-android.yaml
```diff
@@ -34,6 +34,11 @@ on:
       SENTRY_DSN_ANDROID:
         required: true
 
+concurrency:
+  # non-deploy builds shouldn’t conflict, but builds for deployment can’t overlap or they’ll claim the same version number
+  group: build-android-${{ inputs.flavor }}${{ inputs.deploy && '' || github.run_id }}
```
boringcactus, line 39: Oh, crap, this won’t work because `''` is falsy. (There is, I’m fairly certain, no other syntax for conditional expressions in GitHub Actions.)

PR #916, opened 2025-04-24
feat(Android): Reset nearby transit state after app has been backgrounded for over 1 hour

EmmaSimon:
### Summary

_Ticket:_ [Android recenter button bug](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209802459425032?focus=true)

Follow up to #908 

When the app goes to background, remember the timestamp, then the next time it comes back to the foreground, check how long ago it was backgrounded. If it was over an hour ago, clear the navigation stack, and follow the user's current location.

In order to test this, I needed to start injecting `Clock.System` with koin, this is recommended by the [kotlin docs](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-clock/-system/) to make usage of clocks more testable, but it required a bit of refactoring. 

I also switched the `timer` helper to return the state object rather than a single value, because I noticed that the value of `now` was not being updated in the `LifecycleResumeEffect` I added (though I didn't end up actually using `now` there and just call `clock.now()` directly).

android

### Testing

Added a unit test that checks that the state is reset after 1 hour of backgrounding

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitPageTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitPageTest.kt
@@ -16,8 +16,14 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithContentDescription
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performSemanticsAction
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.compose.LocalLifecycleOwner
+import androidx.lifecycle.testing.TestLifecycleOwner
 import androidx.test.rule.GrantPermissionRule
 import com.mapbox.geojson.FeatureCollection
+import com.mapbox.geojson.Point
+import com.mapbox.maps.CameraState
+import com.mapbox.maps.EdgeInsets
 import com.mapbox.maps.extension.compose.animation.viewport.MapViewportState
 import com.mbta.tid.mbta_app.analytics.MockAnalytics
 import com.mbta.tid.mbta_app.android.component.sheet.rememberBottomSheetScaffoldState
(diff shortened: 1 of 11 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/Timer.kt
```diff
@@ -37,15 +38,15 @@ class ClockTickHandler {
...
+        ClockTickHandler.getClockFlow(clock).filter {
             it.toEpochMilliseconds().seconds.inWholeSeconds % tickInterval.inWholeSeconds == 0L
         }
 }
 
 @Composable
-fun timer(updateInterval: Duration = 5.seconds): Instant {
-    val viewModel: TimerViewModel = remember { TimerViewModel(updateInterval) }
-    return viewModel.timerFlow.collectAsState(initial = Clock.System.now()).value
+fun timer(updateInterval: Duration = 5.seconds, clock: Clock): State<Instant> {
```
boringcactus, line 49: It seems like it’d be easier to make this parameter default to `koinInject()` than to add a ton of explicit `clock` parameters to the components that call `timer`.

EmmaSimon: Oh, yeah fair, at first I was thinking that it needed to be a param in the composable so that it could be passed into during the test, but we can just set it in the koin application in the test if we need it.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitPageTest.kt
```diff
@@ -399,4 +411,101 @@ class NearbyTransitPageTest : KoinTest {
 
         composeTestRule.onNodeWithContentDescription("Mapbox Logo").assertDoesNotExist()
     }
+
+    @Test
+    fun testResetAfter1hour() {
```
KaylaBrady, line 416: praise: This seems like a tricky thing to test, nice job writing one!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -284,6 +288,21 @@ fun NearbyTransitPage(
...
 
+    var backgroundTimestamp: Long? by rememberSaveable { mutableStateOf(null) }
+    LifecycleResumeEffect(null) {
+        backgroundTimestamp?.let {
+            val timeSinceBackground = clock.now().minus(Instant.fromEpochMilliseconds(it))
+            if (timeSinceBackground > 1.hours) {
+                navController.navigate(SheetRoutes.NearbyTransit) {
+                    popUpTo(navController.graph.id) { inclusive = true }
+                }
+                nearbyTransit.viewportProvider.follow()
```
KaylaBrady, line 299: question(non-blocking): what happens for users who don't have their location permissions shared?

KaylaBrady (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
EmmaSimon: Good catch! It would get into a broken state and load forever, until you manually panned the map, fixed now.

PR #918, opened 2025-04-25
feat(RouteCardData): Disruption on branch cases

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Group By Direction | Disruption on a branch](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209979069137605?focus=true)

What is this PR for?

With this PR, we now show a separate row for disrupted branches when grouped by headsign to match designs.
Grouping by headsign when we know we are dealing with a branched route seemed like the most straightforward way to accomplish this. TY to @EmmaSimon for feedback on where that grouping logic should live! 

iOS

android

### Testing

What testing have you done?
* Updated stubbed out unit tests 
* Updated android preview data & confirmed displayed as expected
image

* Tested in running app for a few alerts
imageimage

---
  - https://app.asana.com/0/0/1210072519830418

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -143,6 +143,8 @@ data class RouteCardData(
             upcomingTrips,
             alertsHere,
             allDataLoaded,
+            // adding this fakeId will make hasSchedulesToday JustWork for non-branched routes
+            // when routePatterns are not specified in the test
             if (routePatterns.isEmpty()) mapOf("fakeId" to hasSchedulesToday)
             else routePatterns.associate { it.id to hasSchedulesToday },
             alertsDownstream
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -120,11 +122,38 @@ data class RouteCardData(
         override val upcomingTrips: List<UpcomingTrip>,
         val alertsHere: List<Alert>,
         val allDataLoaded: Boolean,
-        override val hasSchedulesToday: Boolean,
+        val hasSchedulesTodayByPattern: Map<String, Boolean>,
```
KaylaBrady, line 125: This is one of my least favorite changes in this PR, but it seemed like the most straightforward path to determining whether we should show "Predictions Unavailable" for a branch with no service vs. if we are dealing with a no scheduled service / service ended scenario.

BrandonTR (approved):

PR #919, opened 2025-04-25
feat(RouteCard): Secondary alert icon at the direction level

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Group By Direction | Secondary alert - show direction-level icon rather than trip level](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209979069137600?focus=true)

What is this PR for?
Instead of putting secondary alert icons on each row, there is now a single secondary alert icon next to the direction label. 

iOS

android

### Testing

What testing have you done?
* Added unit tests 
* Verified looks as expected on running iOS & android apps
IMG_0288
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LeafFormat.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LeafFormat.kt
@@ -34,8 +34,14 @@ sealed class LeafFormat {
         override fun noPredictionsStatus() = (format as? UpcomingFormat.NoTrips)?.noTripsFormat
     }
 
-    /** A [RouteCardData.Leaf] which has multiple destinations within its direction. */
-    data class Branched(val branches: List<Branch>) : LeafFormat() {
+    /**
+     * A [RouteCardData.Leaf] which has multiple destinations within its direction. The secondary
+     * alert will be displayed once for the entire direction
+     */
+    data class Branched(
+        val branches: List<Branch>,
+        val secondaryAlert: UpcomingFormat.SecondaryAlert? = null
+    ) : LeafFormat() {
         data class Branch
         @OptIn(ExperimentalUuidApi::class)
         constructor(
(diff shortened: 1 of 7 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LeafFormat.kt
```diff
@@ -34,8 +34,14 @@ sealed class LeafFormat {
...
     }
 
-    /** A [RouteCardData.Leaf] which has multiple destinations within its direction. */
-    data class Branched(val branches: List<Branch>) : LeafFormat() {
+    /**
+     * A [RouteCardData.Leaf] which has multiple destinations within its direction. The secondary
+     * alert will be displayed once for the entire direction
+     */
+    data class Branched(
+        val branches: List<Branch>,
```
KaylaBrady, line 42: not directly related to this PR, but I was thinking about renaming `Branch` here something like `rows` since it isn't quite the typical meaning of branch. Ex: I think of "Braintree" as one branch, but the list of branches here could include Braintree three times.

boringcactus: Yeah, I guess it’s weird that the list doesn’t contain the distinct branches of the route. I don’t know that just `Row` as the name of the class would make sense, though; maybe `BranchRow`?

boringcactus (approved):
I’m curious how feasible it’d be to check secondary alerts for each branch separately instead of for the leaf as a whole, like we now do with major alerts, but if it’d be even moderately difficult it’s not worth folding in here.

KaylaBrady:
> I’m curious how feasible it’d be to check secondary alerts for each branch separately instead of for the leaf as a whole, like we now do with major alerts, but if it’d be even moderately difficult it’s not worth folding in here.

I think it wouldn't be horrible to we calculate secondary alert by headsign, but that wouldn't address some of the edge cases[ you raised in the miro thread](https://miro.com/app/board/uXjVIDh-CTs=/?openComment=3458764624493932040) about treating the secondary alert at the trip-level, like if the trip's stops differ from the representative trip's stops. I don't _think_ that is possible, but there are still the GL cases you've raised in the past about trips missing a route pattern which could also be misleading if we only show the secondary alert by headsign.

PR #921, opened 2025-04-28
fix(iosApp.NearbyTransitView): Don't reset nearby when stops reordered

KaylaBrady:
### Summary

What is this PR for?
No ticket, addresses[ issue raised in slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1745840910660239) where there have been some reports of nearby transit auto-scrolling to the top of the sheet when there are stops ~equidistant nearby. 

iOS

android

### Testing

What testing have you done?
* added a unit test
* Tested locally with artificial location jitter at a coordinate known to be affected by this issue
* Tested in dev orange

### DIFF
--- a/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
+++ b/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
@@ -45,13 +45,6 @@ struct NearbyTransitView: View {
         let pinnedRoutes: Set<String>
     }
 
-    var setOfNearbyStopIds: Set<String>? {
-        guard let stopIds = nearbyVM.nearbyState.stopIds else {
-            return nil
-        }
-        return Set(stopIds)
-    }
-
     var body: some View {
         VStack(spacing: 0) {
             if let routeCardData = nearbyVM.routeCardData,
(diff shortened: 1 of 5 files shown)


iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
```diff
@@ -70,7 +70,7 @@ struct NearbyTransitView: View {
         .onChange(of: location) { newLocation in
             getNearby(location: newLocation, globalData: globalData)
         }
-        .onChange(of: nearbyVM.nearbyState.stopIds) { nearbyStops in
+        .onChange(of: Set(nearbyVM.nearbyState.stopIds)) { nearbyStops in
```
KaylaBrady, line 73: Oh right this won't Just Work, have to handle nil stopIds & joinPredictions

iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
```diff
@@ -70,11 +70,14 @@ struct NearbyTransitView: View {
...
             getNearby(location: newLocation, globalData: globalData)
         }
-        .onChange(of: nearbyVM.nearbyState.stopIds) { nearbyStops in
-            updateNearbyRoutes()
-            getSchedule()
-            joinPredictions(nearbyStops)
-            scrollToTop()
+        .onChange(of: nearbyVM.nearbyState.stopIds) { [oldValue = nearbyVM.nearbyState.stopIds] newNearbyStops in
+
+            if Set(oldValue ?? []) != Set(newNearbyStops ?? []) {
```
KaylaBrady, line 75: this is a little icky, but it makes it easier to write a test.

EmmaSimon (approved):
Thanks for cleaning up my mess 🙏

KaylaBrady:
Didn't fix the issue

EmmaSimon (commented):
Can we also try removing the `routeCardData = nil` in `NearbyViewModel.getNearbyStops` to see if that prevents the reloading/jumping to the top when location does change?

KaylaBrady:
@EmmaSimon I tried it out walking around the block and it definitely made the difference, great call! I still think 10m is reasonable based on the amount of wobble I see at my desk locally (and is at least better than the 100m now on android), but agree we can revisit this value once it is in staging!

PR #922, opened 2025-04-28
feat(iOS): add helper for getting settings

boringcactus:
### Summary

_Ticket:_ [Explore helpers for reading settings on iOS and Android](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209925176921390?focus=true)

Stores the cache in an `@Environment` property and offers a property wrapper `@GetSetting` for convenience. Less slick than the Android version #920 because SwiftUI state management doesn’t compose as nicely as Compose state management does.

iOS

android

### Testing

Checked that the onboarding flow and the settings page perform as intended. Updated existing unit tests. I’m getting a couple of failures locally but they seem like they should be flukes.


iosApp/iosApp/Pages/More/MoreToggle.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Copyright © 2025 MBTA. All rights reserved.
+//
+
+import Shared
+import SwiftUI
+
+struct MoreToggle: View {
+    let label: String
+    let setting: Settings
+    @ObservedObject var settingsCache: SettingsCache
```
boringcactus, line 15: You can’t combine `@ObservedObject` and `@Environment` (they can’t each wrap the property independently, one of them has to wrap the other wrapper, and neither of them wants to do that), and it’s critical that we rerender the view when the cache content changes, so this has to be passed explicitly, which is unfortunate.

iosApp/iosAppTests/Pages/StopDetails/StopDetailsFilteredDepartureDetailsTests.swift
```diff
@@ -679,7 +683,7 @@ final class StopDetailsFilteredDepartureDetailsTests: XCTestCase {
         )
         let nearbyVM = NearbyViewModel()
         let stopDetailsVM = StopDetailsViewModel()
-        stopDetailsVM.showStationAccessibility = true
+        // TODO: stationAccessibility
```
BrandonTR, line 686: Looks like we're setting this above, did you mean to remove this comment?

BrandonTR (dismissed):
Really like how clean this makes accessing settings. Nice work!

iosApp/iosAppTests/Pages/StopDetails/StopDetailsViewTests.swift
```diff
@@ -105,11 +104,20 @@ final class StopDetailsViewTests: XCTestCase {
...
-        ViewHosting.host(view: sut)
-        let routePills = try sut.inspect().find(StopDetailsFilterPills.self).findAll(RoutePill.self)
-        XCTAssertEqual(2, routePills.count)
-        XCTAssertNotNil(try routePills[0].find(text: "Should be first"))
-        XCTAssertNotNil(try routePills[1].find(text: "Should be second"))
+        DefaultSettings.set([.groupByDirection: true])
+        defer { DefaultSettings.reset() }
+
+        let exp = sut.on(\.didAppear) { view in
+//        let exp = sut.inspection.inspect(after: 0.1) { view in
```
BrandonTR, line 111: Can we remove this as well?

iosApp/iosAppTests/Pages/StopDetails/StopDetailsFilteredDepartureDetailsTests.swift
boringcactus: We’re not setting this in this test. It looks like we don’t need to be, either, but I’d rather leave the test as is. I did mean to go back through and take care of the rest of these, but then a weekend happened and I forgot how close to ready this actually was.

iosApp/iosAppTests/Pages/StopDetails/StopDetailsViewTests.swift
boringcactus: Yes.

iosApp/iosApp/Utils/SettingsCache.swift
```diff
@@ -0,0 +1,102 @@
...
+        content()
+            .task { try? await cache.load() }
+            .environment(\.settingsCache, cache)
+    }
+}
+
+/// Retrieves the value of an individual setting from a `SettingsCache` retrieved from the environment.
+///
+/// May or may not actually update the component when the settings cache finishes loading, but in practice seems to.
+@propertyWrapper struct GetSetting: DynamicProperty {
```
boringcactus, line 84: Realized while writing doc comments per @KaylaBrady’s suggestion in #920 that I don’t actually know why this works. To get the toggles in the settings page to pick up their new state immediately, we need to use `@ObservedObject` and a layer of component indirection, but this doesn’t have that and the components still see the right values of the settings once the settings have loaded.

Two reasons that this might be the case:
1. Some magic in SwiftUI is happening to make this work that isn’t present when accessing the settings cache directly within the view. This would be nice.
2. The components are rerendering on unrelated property changes - `now` or something derived from it, most likely - and just seeing the new value when they read the `@GetSetting`-wrapped property. This would be terrifying, because six months from now when we rearrange the state and the component stops rerendering, we will have forgotten about this, and the component will see the default value of the setting and never update to see the loaded value of the setting, and the best we can hope for is that the miserable debugging session happens *before* we’ve shipped.

Conveniently, I’ve thought of a way to check this, but it might take a minute to execute and I want to get this written down first.

boringcactus: It was 2. If I inject the `SettingsCache` as an `EnvironmentObject` and just read it directly from everything, it looks like `@EnvironmentObject` will correctly rerender the component if a `@Published` property changes, but unfortunately, that leads to ViewInspector tests and failed `environmentObject` inheritance.

KaylaBrady: @boringcactus I thought ViewInspector magic[ worked for environmentObject ](https://mbta.slack.com/archives/C062QNAJZ2M/p1745607781385229?thread_ts=1745605121.630529&cid=C062QNAJZ2M)but not bare Environment?

boringcactus: It’s supposed to, but apparently sometimes it doesn’t.

boringcactus: Specifically, reading the environment object from the view initializer causes all hell to break loose; I think it’s probably a good idea to avoid reading settings from the view initializer anyway, though, so that we won’t get stuck in the wrong state if the settings are still loading when the component appears and they finish loading later.

boringcactus:
Blocked on https://github.com/nalexn/ViewInspector/issues/380.

KaylaBrady (commented):
🎉 🎉 Amazing - love that we won't have to drill settings in a bunch of places anymore! Thanks for doing this work and staying on top of the upstream issue!

Just a few small questions

iosApp/iosApp/Utils/SettingsCache.swift
```diff
@@ -0,0 +1,61 @@
...
+        cache = try await settingsRepo.getSettings().mapValues { $0.boolValue }
+    }
+
+    /// Edits the value of a single `setting` in both the cache and the settings repository.
+    func set(_ setting: Settings, _ value: Bool) {
+        let newSettings = [setting: value]
+        cache = newSettings.merging(cache ?? [:]) { newValue, _ in newValue }
+        Task {
+            try await settingsRepo.setSettings(settings: [setting: KotlinBoolean(bool: value)])
+        }
```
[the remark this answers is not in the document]
boringcactus: I think if you change a setting and then it fails to save, we should reflect the value you intended to set for the rest of the session, instead of resetting the setting immediately; giving you some time with the correct value seems better than giving you no time with the correct value.

iosApp/iosApp/Utils/SettingsCache.swift
```diff
@@ -0,0 +1,61 @@
...
+            .environmentObject(cache)
+    }
+}
+
+extension View {
+    /// Sets the settings within this view to some fixed value.
+    func withFixedSettings(_ settings: [Settings: Bool]) -> some View {
+        environmentObject(SettingsCache(cache: settings))
+    }
+}
```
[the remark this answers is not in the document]
boringcactus: It could, but this is also necessary for previews of views that read settings.

iosApp/iosAppTests/Pages/StopDetails/StopDetailsViewTests.swift
```diff
@@ -60,11 +60,16 @@ final class StopDetailsViewTests: XCTestCase {
             )
         )
 
-        ViewHosting.host(view: sut)
-        let routePills = try sut.inspect().find(StopDetailsFilterPills.self).findAll(RoutePill.self)
-        XCTAssertEqual(2, routePills.count)
-        XCTAssertNotNil(try routePills[0].find(text: "Should be first"))
-        XCTAssertNotNil(try routePills[1].find(text: "Should be second"))
+        let exp = sut.on(\.didAppear) { view in
```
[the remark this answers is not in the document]
boringcactus: Oh, this was because when this PR was first written `StopDetailsUnfilteredView` was reading the `groupByDirection` state in its `init`, which completely breaks with `@EnvironmentObject`, so the `groupByDirection` needed to be checked in `body` and the state wouldn’t be correct until the view had actually appeared. That’s not a problem anymore, though.

iosApp/iosApp/Utils/SettingsCache.swift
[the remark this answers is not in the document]
KaylaBrady: Yeah that's fair. Maybe in an ideal world there'd be some automatic retrying or a toast notification to prompt the user to retry, but failure to save seems like a rare event anyway that it wouldn't be worth optimizing over. We could consider logging to sentry if it does happen to fail so that we understand the frequency of failure.

KaylaBrady (approved):

iosApp/iosApp/Utils/SettingsCache.swift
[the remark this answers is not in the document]
boringcactus: Added that in 1d949ea.

PR #927, opened 2025-04-29
feat(iOS): Filtered stop direction grouping UX

EmmaSimon:
### Summary

_Ticket:_ [Group by direction |  Filtered Stop Details uses new data types](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209914022959070?focus=true)

Switch over to using the `RouteCardData` objects in filtered stop details. This completely breaks the headsign grouped stop details, if you try to use it, you'll get a message telling you to turn on the Group by Direction feature toggle.

iOS

### Testing

Updated all relevant unit tests to pass using direction grouping.

One UI test that we have, `EndToEndOpenStopDetailsTest.testOpenStopDetails` was deactivated, since it can't be tested without the feature flag being enabled, and getting that to work didn't seem worth it (maybe it'll be trivial with the new iOS settings updates?).
I added a TODO and an AC to [this](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210113865246640?focus=true) cleanup ticket to enable it again.

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredView.swift
+++ b/iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredView.swift
@@ -17,12 +17,16 @@ struct StopDetailsFilteredView: View {
     var setStopFilter: (StopDetailsFilter?) -> Void
     var setTripFilter: (TripDetailsFilter?) -> Void
 
-    var departures: StopDetailsDepartures?
+    var routeCardData: [RouteCardData]?
     var now: Date
+
+    var routeData: RouteCardData?
+    var stopData: RouteCardData.RouteStopData?
+    var leafData: RouteCardData.Leaf?
 
     var alerts: [Shared.Alert]
     var downstreamAlerts: [Shared.Alert]
-    var patternsByStop: PatternsByStop?
+
     var servedRoutes: [StopDetailsFilterPills.FilterBy] = []
 
     @ObservedObject var errorBannerVM: ErrorBannerViewModel
@@ -44,7 +48,7 @@ struct StopDetailsFilteredView: View {
         tripFilter: TripDetailsFilter?,
         setStopFilter: @escaping (StopDetailsFilter?) -> Void,
         setTripFilter: @escaping (TripDetailsFilter?) -> Void,
-        departures: StopDetailsDepartures?,
+        routeCardData: [RouteCardData]?,
         now: Date,
         errorBannerVM: ErrorBannerViewModel,
         nearbyVM: NearbyViewModel,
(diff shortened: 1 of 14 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -39,6 +39,12 @@ interface ILeafData {
     val hasSchedulesToday: Boolean
 }
 
+data class DepartureDataBundle(
```
boringcactus, line 42: Are there places in the Android code that are worth refactoring to use this helper class?

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredView.swift
```diff
@@ -17,12 +17,16 @@ struct StopDetailsFilteredView: View {
     var setStopFilter: (StopDetailsFilter?) -> Void
     var setTripFilter: (TripDetailsFilter?) -> Void
 
-    var departures: StopDetailsDepartures?
+    var routeCardData: [RouteCardData]?
     var now: Date
 
+    var routeData: RouteCardData?
+    var stopData: RouteCardData.RouteStopData?
+    var leafData: RouteCardData.Leaf?
```
boringcactus, line 25: If there’s nothing we’d show if, say, the `routeData` and `stopData` exist but the `leafData` is `nil`, would it make sense to just store one `DepartureDataBundle?`?

iosApp/iosAppTests/Pages/StopDetails/StopDetailsFilteredDepartureDetailsTests.swift
```diff
@@ -17,78 +17,75 @@ final class StopDetailsFilteredDepartureDetailsTests: XCTestCase {
...
+    func makeBundle(
+        route: Route,
+        stop: Stop? = nil,
+        patterns: [RoutePattern]? = nil,
+        upcomingTrips: [UpcomingTrip]? = nil,
+        alerts: [Shared.Alert] = [],
+        objects: ObjectCollectionBuilder = ObjectCollectionBuilder()
+    ) -> DepartureDataBundle {
+        makeBundle(
+            lineOrRoute: RouteCardDataLineOrRouteRoute(route: route),
```
boringcactus, line 29: I think if `LineOrRoute` was a `sealed class` rather than a `sealed interface` `.Route(route: route)` would be sufficient here; that may or may not be a change worth making at this stage.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredView.swift
```diff
@@ -121,18 +112,22 @@ struct StopDetailsFilteredView: View {
             }
             .fixedSize(horizontal: false, vertical: true)
 
-            if let patternsByStop {
+            if !nearbyVM.groupByDirection {
+                DebugView { Text(String(
+                    stringLiteral: "Turn on the Group by Direction feature toggle, we're dropping support for headsign grouping"
```
boringcactus, line 117:
```suggestion
                    stringLiteral: "Turn on the Group by Direction feature toggle, we’re dropping support for headsign grouping"
```

I need to be stopped. The macOS keyboard shortcut for [typographically correct apostrophes](https://practicaltypography.com/apostrophes.html) is in my muscle memory, though, and once I saw this straight quote, I couldn’t unsee it.

boringcactus (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
EmmaSimon: I couldn't decide, if we're still considering doing a refactor to include the route and stop data in each leaf, then this should be removed then. But if we think this is good enough then we can adopt it everywhere. I'd maybe prefer to aim for a better solution, but still not certain how involved that would end up being.

iosApp/iosAppTests/Pages/StopDetails/StopDetailsFilteredDepartureDetailsTests.swift
EmmaSimon: Ehhh, yeah if we needed to do this outside of the test code I'd be more eager to change it, but I think this is fine.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
EmmaSimon: Added a [ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210114621155024?focus=true) for refactoring away the need for bundling.

PR #928, opened 2025-04-29
fix(RouteCardData): Omit stops that have redundnant upcoming service

KaylaBrady:
### Summary

_Ticket:_ [Group by direction | Multiple nearby bus stops in the same direction](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210071791285856?focus=true)

What is this PR for?

This PR filters out stops from nearby transit that have upcoming trips that are completely redundant to earlier stops. 

Previously, we'd include a stop if it had any route pattern not seen by an earlier stop. Now, we include a stop if it has any route pattern not seen by an earlier stop that is typical or has upcoming service.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Verified that for the GL case of Fenwood Road & Brigham Circle, Only Fenwood Road is shown. 
Before:
image
After:
image


EmmaSimon (approved):
Amazing, glad this was relatively straightforward 🎉

PR #929, opened 2025-04-29
fix: Discard track change alerts at core CR stops

EmmaSimon:
### Summary

_Ticket:_ [Group by Direction | Discard track changes at core stations](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210082542778227?focus=true)

We had a check in the `NearbyStaticData` to remove track change alerts from core CR stops (South and North St, Ruggles, Back Bay), which hadn't been included in the `RouteCardData` logic. This adds it.

iOS

android

### Testing

Added new unit test verifying this behavior, manually verified with test alerts on dev orange

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -852,9 +852,6 @@ data class RouteCardData(
             }
         }
 
-        private fun List<Alert>.discardTrackChangesAtCRCore(isCRCore: Boolean): List<Alert> =
-            if (isCRCore) this.filterNot { it.effect == Alert.Effect.TrackChange } else this
-
         fun addAlerts(
             alerts: AlertsStreamDataResponse?,
             includeMinorAlerts: Boolean,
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -852,6 +852,9 @@ data class RouteCardData(
             }
         }
 
+        private fun List<Alert>.discardTrackChangesAtCRCore(isCRCore: Boolean): List<Alert> =
+            if (isCRCore) this.filterNot { it.effect == Alert.Effect.TrackChange } else this
```
KaylaBrady, line 856: suggestion (non-blocking): This could be moved to the `Alert` file for consistency with how we've co-located list extension functions for `UpcomingTrip` in that file.

KaylaBrady (approved):
Nice!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
EmmaSimon: I couldn't decide if it made sense for it to be public outside of this, since it's pretty specific to this use case, but I yeah I'll move it out just for clarity.

PR #930, opened 2025-05-01
fix(iOS): Group by direction UI polish

BrandonTR:
### Summary

_Ticket:_ [Group by direction |  Polish UI](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209912113394452?focus=true)

I think this is everything! Let me know if you see anything that should be addressed.

iOS

android

### Testing

Manual


boringcactus:
Moving from filtered to unfiltered stop details causes all the trips to be shown for a surprisingly long time:

https://github.com/user-attachments/assets/ae494556-5b38-4803-8c40-1de13233c232

If that’s annoying to fix, we can spin it out as a separate ticket.

BrandonTR:
> Moving from filtered to unfiltered stop details causes all the trips to be shown for a surprisingly long time:
> 
> If that’s annoying to fix, we can spin it out as a separate ticket.

I went ahead and made a ticket for this https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210162265230760?focus=true

boringcactus (approved):

PR #932, opened 2025-05-01
chore: generate test data from a subset of real data

boringcactus:
### Summary

_Ticket:_ [Make sets of test data for common scenarios for unit tests](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1206787308358329?focus=true)

I don’t know if I love this approach to generating the data, but I do think it would be more difficult to generate this subset from a Gradle task where we’re just tossing around the JSON directly.

I have not exhaustively checked for places where we could be using the new `TestData`.

iOS

android

### Testing

Checked that all tests still pass. My Android Studio previews are always claiming to be out of date, and my Xcode previews are crashing because they can’t find WebKit; I don’t think either of those is related to these changes, but it’s hard to know.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -18,6 +18,7 @@ kotlin = "2.1.10"
 javaphoenixclient = "1.3.1"
 junit = "4.13.2"
 koin-bom = "4.0.4"
+kotlinpoet = "2.1.0"
 kotlinxCoroutinesCore = "1.10.2"
 kotlinxDatetime = "0.6.2"
 ktor = "3.1.2"
@@ -61,12 +62,14 @@ koin-core = { module = "io.insert-koin:koin-core" }
 koin-test = { module = "io.insert-koin:koin-test" }
 kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test", version.ref = "kotlin" }
 koin-junit4 = { module = "io.insert-koin:koin-test-junit4" }
+kotlinpoet = { module = "com.squareup:kotlinpoet", version.ref = "kotlinpoet" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinxCoroutinesCore" }
 kotlinx-datetime = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version.ref = "kotlinxDatetime" }
 ktor-client-content-negotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
 ktor-client-core = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
 ktor-client-darwin = { module = "io.ktor:ktor-client-darwin", version.ref = "ktor" }
 ktor-client-encoding = { module = "io.ktor:ktor-client-encoding", version.ref = "ktor" }
+ktor-client-java = { module = "io.ktor:ktor-client-java", version.ref = "ktor" }
 ktor-client-mock = { module = "io.ktor:ktor-client-mock", version.ref = "ktor" }
 ktor-client-okhttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
 ktor-client-websockets = { module = "io.ktor:ktor-client-websockets", version.ref = "ktor" }
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/utils/TestData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/utils/TestData.kt
@@ -778,6 +778,57 @@ public val TestData: ObjectCollectionBuilder by lazy {
 
     objects.put(
         Stop(
+            id = "121",
+            latitude = 42.320511,
+            longitude = -71.05178,
+            name = "JFK/UMass",
+            locationType = LocationType.STOP,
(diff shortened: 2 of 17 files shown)


gradle/libs.versions.toml
```diff
@@ -61,12 +62,14 @@ koin-core = { module = "io.insert-koin:koin-core" }
 koin-test = { module = "io.insert-koin:koin-test" }
 kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test", version.ref = "kotlin" }
 koin-junit4 = { module = "io.insert-koin:koin-test-junit4" }
+kotlinpoet = { module = "com.squareup:kotlinpoet", version.ref = "kotlinpoet" }
```
boringcactus, line 65: [KotlinPoet](https://square.github.io/kotlinpoet/) is a library for generating Kotlin code; it makes some string escaping stuff a little less confusing. I tried adopting it in `buildSrc/` for our existing code generation, but it wound up being a net increase in lines of code, and the output isn’t nicely formatted so it’d be actually kind of annoying to try to use in `buildSrc/` for the things that run automatically and produce outputs that we keep in-tree.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/utils/TestData.kt
KaylaBrady, line 1: question (non-blocking): lol github struggles when I try to comment within this file since it is so big! My assumption is we should never manually edit this file, is that right? That seems implied by pointing to the project utils as owner, but maybe worth stating explicitly in the comment.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/map/RouteLayerGeneratorTest.kt
```diff
@@ -19,14 +19,14 @@ class RouteLayerGeneratorTest {
 
         assertEquals(
             listOf(
-                "route-layer-Red",
                 "route-layer-Orange",
-                "route-layer-Red-alerting-bg",
-                "route-layer-Red-shuttled",
-                "route-layer-Red-suspended",
+                "route-layer-Red",
                 "route-layer-Orange-alerting-bg",
```
KaylaBrady, line 24: question(non-blocking): Is this just b/c of sort order changing for the test data?

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/ProjectUtils.kt
```diff
@@ -0,0 +1,269 @@
...
+                "87",
+                "CR-Fitchburg",
+                "CR-Haverhill",
+                "CR-Lowell",
+                "CR-Newburyport",
+                "CR-Providence",
+                "Orange",
+                "Red"
+            )
+        val allStopsOnRoutes = setOf("Green-B", "Green-C", "Green-D", "Green-E", "Red")
```
KaylaBrady, line 50: question(non-blocking): Why not for all the included routes?

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/ProjectUtils.kt
```diff
@@ -0,0 +1,269 @@
...
+                get() = Path("build/projectUtils/data").toOkioPath()
+
+            override val cache: Path
+                get() = Path("build/projectUtils/cache").toOkioPath()
+        }
+    }
+}
+
+/**
+ * Utilities that we may need to run from time to time that need access to the real business logic.
```
KaylaBrady, line 30: praise: appreciate you making this easy to re-run in the future if needed!

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/model/Dependency.jvm.kt
```diff
@@ -0,0 +1,3 @@
+package com.mbta.tid.mbta_app.model
```
KaylaBrady, line 1: question(non-blocking): Why is this & FeaturePromo.jvm.kt now needed?

KaylaBrady (commented):
🤩 amazing work! Love seeing how many existing tests this simplifies already and look forward to the ease of writing tests consistently & reviewing tests easily going forward!

KaylaBrady (approved):

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/map/RouteLayerGeneratorTest.kt
boringcactus: Yes, the sort order had previously been undefined and is now defined.

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/model/Dependency.jvm.kt
boringcactus: An `expect` with no matching `actual` is an error at compile time.

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/ProjectUtils.kt
boringcactus: There are a lot of bus stops in particular that we don’t need data for, and including all stops from all included routes doubles the size of TestData.kt.

PR #933, opened 2025-05-01
fix: Don't hide predictions on a trunk where major disruptions are happening only on a subset of branches

EmmaSimon:
### Summary

_Ticket:_ [Group by Direction | Stop Details shouldn't suppress all predictions when only one branch alerting](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210072519830418?focus=true)

Before this change, any major alert that was only relevant for a single branching route going through a stop would suppress all predictions at that stop, even if some branches had normal service.

This change piggybacks on the logic in [Leaf.formatForBranchedService](https://github.com/mbta/mobile_app/blob/main/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt#L327-L340), which converts branches to a `LeafFormat.Single` with a `UpcomingFormat.Disruption` only when the alert affects every branch on the trunk, so the `LeafFormat` can just check that in order to determine if predictions should be suppressed when a major alert exists at a stop.

Simulator Screenshot - Clone 1 of iPhone 16 - 2025-05-01 at 17 13 37

iOS

android

### Testing

Got existing tests passing, added unit tests for the new behavior


iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -32,10 +29,28 @@ struct StopDetailsFilteredDepartureDetails: View {
 
     @EnvironmentObject var viewportProvider: ViewportProvider
 
+    var testTiles: [TileData]? = nil
```
EmmaSimon, line 32: This feels sketchy, but it felt worse to continue passing in the actual tiles separately, since we're now getting more relevant data from the leafFormat because we needed so many values from it that it felt silly to continue passing them all in separately. In order to continue passing the tiles in separately, we'd need to generate the `leafFormat` twice, once in `StopDetailsFilteredView` and once here. I'm hoping we can get rid of this during the [refactor ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210114621155024?focus=true) where we'll have a presentational component which takes the tiles directly and can replace all the tests in `StopDetailsFilteredDepartureDetailsTests`.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsFilteredDeparturesView.kt
```diff
@@ -79,6 +79,7 @@ fun StopDetailsFilteredDeparturesView(
     data: FilteredDeparturesData,
     tileData: List<TileData>,
     noPredictionsStatus: NoTripsFormat?,
+    hidePredictions: Boolean,
```
KaylaBrady, line 82: suggestion (non-blocking): Consider defaulting to false here to avoid having to set it in all unaffected tests

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LeafFormat.kt
```diff
@@ -10,6 +10,14 @@ sealed class LeafFormat {
 
     abstract fun noPredictionsStatus(): UpcomingFormat.NoTripsFormat?
 
+    val hidePredictions: Boolean
```
KaylaBrady, line 13: suggestion: It may help w/ clarity for this variable name to reflect what it represents directly, something like `isAllServiceDisrupted`. `hidePredictions` doesn't capture that this would also hide schedules for say a branched CR route. It also seems more named based on the UI-related effects, which I don't think the data representation should be aware of.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
KaylaBrady: Yeah that seems OK to me as a temporary stopgap! Worth calling out specifically in the refactor ticket once this is merged.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
```diff
@@ -32,10 +29,28 @@ struct StopDetailsFilteredDepartureDetails: View {
...
+
     let inspection = Inspection<Self>()
 
     var analytics: Analytics = AnalyticsProvider.shared
 
+    var leafFormat: LeafFormat {
```
KaylaBrady, line 38: question (non-blocking): Should there be an `onChange` to recalculate this if the global data changes?

KaylaBrady (commented):
Nice, your approach to sharing the logic of branch formatting is much cleaner than I was thinking of!! Just one naming suggestion

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredDepartureDetails.swift
EmmaSimon: Added it to the AC

PR #956, opened 2025-05-05
fix: Update group by direction sort logic

EmmaSimon:
### Summary

_Ticket:_ [Group by direction | Rearrange sort precedence](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210130069642283?focus=true)

Don't sort routes to the bottom of nearby if only a single stop has service ended for the day. Also start sorting multiple stops in a route card based on distance and service ended status, along with service ended leaves.

Screenshot_20250505_123946

iOS

android

### Testing

Updated existing sort test to actually test RouteCardData, and added a new test for the updated behavior

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Stop.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Stop.kt
@@ -2,6 +2,8 @@ package com.mbta.tid.mbta_app.model
 
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import io.github.dellisd.spatialk.geojson.Position
+import io.github.dellisd.spatialk.turf.ExperimentalTurfApi
+import io.github.dellisd.spatialk.turf.distance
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Stop.kt
```diff
@@ -44,6 +46,11 @@ data class Stop(
 
     fun resolveParent(global: GlobalResponse) = resolveParent(global.stops)
 
+    @OptIn(ExperimentalTurfApi::class)
+    fun distanceFrom(position: Position): Double {
+        return distance(position, this.position)
+    }
```
BrandonTR, line 52: Nitpick: I think we can just make this: `fun distanceFrom(position: Position): Double = distance(position, this.position)`

EmmaSimon: Good call, done!

PR #959, opened 2025-05-06
refactor: rearrange some logic before group by direction cleanup

boringcactus:
### Summary

_Ticket:_ [Group by Direction | Cleanup old nearby data codepaths](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210113865246640?focus=true)

There are some helper functions from the pre-group-by-direction classes that we use from group by direction code, so we need to move those functions out before we can get rid of those classes, and we would benefit from keeping those unit tests, too. Those unit tests are huge, though, which has unfortunate consequences for the PR size.

iOS

android

### Testing

Checked that all tests pass.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
@@ -4,15 +4,12 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
 import androidx.lifecycle.ViewModel
-import com.mbta.tid.mbta_app.analytics.Analytics
-import com.mbta.tid.mbta_app.android.util.fetchApi
 import com.mbta.tid.mbta_app.model.NearbyStaticData
 import com.mbta.tid.mbta_app.model.RouteCardData
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.ScheduleResponse
-import com.mbta.tid.mbta_app.repositories.IErrorBannerStateRepository
 import com.mbta.tid.mbta_app.repositories.INearbyRepository
 import io.github.dellisd.spatialk.geojson.Position
 import kotlinx.coroutines.CoroutineScope
@@ -24,8 +21,6 @@ import org.koin.core.component.KoinComponent
 
 class NearbyTransitViewModel(
     private val nearbyRepository: INearbyRepository,
-    private val errorBannerRepository: IErrorBannerStateRepository,
-    private val analytics: Analytics,
 ) : KoinComponent, ViewModel() {
     var loading by mutableStateOf(false)
     var nearby by mutableStateOf<NearbyStaticData?>(null)
@@ -37,7 +32,6 @@ class NearbyTransitViewModel(
     fun getNearby(
         globalResponse: GlobalResponse,
         location: Position,
-        groupByDirection: Boolean,
         setLastLocation: (Position) -> Unit,
         setSelectingLocation: (Boolean) -> Unit
     ) {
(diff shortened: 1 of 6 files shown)


iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -158,15 +162,25 @@ struct StopDetailsPage: View {
...
+                .autoStopFilter(departures: internalDepartures))
         if stopFilter != nextStopFilter {
             nearbyVM.setLastStopDetailsFilter(stopId, nextStopFilter)
         }
         return nextStopFilter
     }
 
     func setTripFilter(filters: StopDetailsPageFilters) {
-        let tripFilter = internalDepartures?.autoTripFilter(
+        let tripFilter = nearbyVM.groupByDirection ? StopDetailsUtils.shared.autoTripFilter(
```
EmmaSimon, line 176: It might make sense to wait for the flag removal PR to get merged before merging this so that you can just not introduce and call all the `StopDetailsDepartures` versions of these functions. Or would that make it harder to get things in a compiling state to have this as an intermediate PR before the one ripping everything out fully?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -626,6 +620,31 @@ data class RouteCardData(
                     .build(sortByDistanceFrom)
                     .sort(sortByDistanceFrom, pinnedRoutes)
             }
+
+        fun getSchedulesTodayByPattern(schedules: ScheduleResponse?): Map<String, Boolean>? =
```
EmmaSimon, line 624: Could this exist directly on `ScheduleResponse` instead?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsUtils.kt
```diff
@@ -0,0 +1,212 @@
...
+    fun autoTripFilter(
+        routeCardData: List<RouteCardData>?,
+        stopFilter: StopDetailsFilter?,
+        currentTripFilter: TripDetailsFilter?,
+        filterAtTime: Instant,
+        globalData: GlobalResponse?
+    ): TripDetailsFilter? {
+        val route = routeCardData?.find { it.lineOrRoute.id == stopFilter?.routeId } ?: return null
+        val leaf =
+            route.stopData.singleOrNull()?.data?.find { it.directionId == stopFilter?.directionId }
```
EmmaSimon, line 90: Is it safe to use `.stopData.singleOrNull()` here? I'd expect to need to do another find to get the matching stopId (which I guess would also need to be passed in), but maybe it's safe to assume that the data here was generated for a specific stop details page? But if that was the case, wouldn't we be guaranteed to also just have a single `RouteCardData` for a single route/line?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
```diff
@@ -215,6 +214,31 @@ constructor(
...
+        fun formatUpcomingTrip(
+            now: Instant,
+            upcomingTrip: UpcomingTrip,
+            routeType: RouteType,
+            context: TripInstantDisplay.Context
+        ) = formatUpcomingTrip(now, upcomingTrip, routeType, context, routeType.isSubway())
+
+        fun formatUpcomingTrip(
+            now: Instant,
+            upcomingTrip: UpcomingTrip,
```
EmmaSimon, line 227: Why not have these be instance methods on individual `UpcomingTrips`?

iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
boringcactus: Making the tests pass in this PR before that PR is merged is kind of annoying, but I guess it’s less annoying than making everything still work regardless of which PR gets merged first.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
boringcactus: Sure.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/StopDetailsUtils.kt
boringcactus: This is only ever reasonable to call from stop details, so there will only ever be a single stop, but filtered and unfiltered stop details have the same list of route card data, so there may be several routes. (They technically don’t even have quite the same list of route card data, since the context passed in is different, but taking further advantage of that by discarding non-selected routes more eagerly is not in scope here.)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
boringcactus: They already weren’t, but I guess they could be.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
```diff
@@ -76,13 +94,14 @@ class NearbyTransitViewModel(
         now: Instant,
         pinnedRoutes: Set<String>?,
     ) {
-        if (global == null || location == null) {
+        val stopIds = nearbyStopIds
```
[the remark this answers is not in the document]
EmmaSimon: Huh, if it's not assigned it should probably be removed. But cool, that sounds reasonable.

PR #960, opened 2025-05-06
fix: Don't fetch schedules & predictions for redundant stops

KaylaBrady:
### Summary

_Ticket:_ [only fetch schedules / predictions for stops that could be shown in nearby transit](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210112159715263?focus=true)

What is this PR for?
* As of group by direction, we fetch schedules and predictions for all nearby stops. This returns to the behavior of only fetching schedules and predictions for the nearest stops that serve at least one unique pattern. This intends to addressed the increase in latency in fetching initial data. It may also resolve some of the [lingering location jitter issue](https://mbta.slack.com/archives/C05QMG9GS9M/p1746478791048789) by reducing the set of stops that are compared when determining whether to reload data / scroll to the top.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran iOS app locally, confirmed data still loaded as expected and that data is fetched for fewer stops. For example, the default location used to fetch predictions for 117 stops, whereas now it fetches for 75. That is still a lot! Note that it includes all child stops for stations.


iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
```diff
@@ -70,11 +70,14 @@ struct NearbyTransitView: View {
         .onChange(of: location) { newLocation in
             getNearby(location: newLocation, globalData: globalData)
         }
-        .onChange(of: nearbyVM.nearbyState.stopIds) { nearbyStops in
-            updateNearbyRoutes()
-            getSchedule()
-            joinPredictions(nearbyStops)
-            scrollToTop()
+        .onChange(of: nearbyVM.nearbyState.stopIds) { [oldValue = nearbyVM.nearbyState.stopIds] newNearbyStops in
```
KaylaBrady, line 73: approach from https://github.com/mbta/mobile_app/pull/921

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RoutePattern.kt
```diff
@@ -30,4 +32,109 @@ data class RoutePattern(
...
+         * earlier stops in the list; each stop must serve at least one route pattern that is not
+         * seen by any earlier stop.
+         */
+        fun filterStopsWithRedundantPatterns(
+            stopIds: List<String>,
+            globalData: GlobalResponse
+        ): List<String> {
+            val parentToAllStops = Stop.resolvedParentToAllStops(stopIds, globalData)
+
+            return patternsGroupedByLineOrRouteAndStop(
```
KaylaBrady, line 54: I don't love that this is building data structures we don't care about, but seemed like the simplest approach. Since we run `patternsGroupedByLineOrRouteAndStop` every time we build route cards, I don't think this should have a negative performance impact.

KaylaBrady:
No rail & CR stops are showing up now -- parent stop issue???

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
```diff
@@ -1,17 +1,34 @@
...
+     * Gets the list of stops within 0.5 miles (or 1 mile for CR). Includes only stops with
+     * [LocationType.STOP].
+     *
+     * @param excludeRedundantService when true, this will omit stops that serves route patterns
+     *   that are all served by closer stops.
+     */
+    fun getStopIdsNearby(
+        global: GlobalResponse,
+        location: Position,
+        excludeRedundantService: Boolean = false
```
boringcactus, line 22: I think giving this parameter no default value would make it clearer when we still want the old behavior.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
```diff
@@ -32,34 +49,68 @@ class NearbyRepository : KoinComponent, INearbyRepository {
...
     }
 }
 
-class MockNearbyRepository(
-    val response: NearbyResponse,
-    val stopIds: List<String> = response.stopIds
-) : INearbyRepository {
-    override fun getStopIdsNearby(global: GlobalResponse, location: Position): List<String> =
-        stopIds
+class MockNearbyRepository(val response: NearbyResponse, val stopIds: List<String> = emptyList()) :
```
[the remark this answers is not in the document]
KaylaBrady: Bad rebase on my part!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
KaylaBrady: Yeah that's a good call. Now that the cleanup of old codepaths is done there might not actually be places we still want the old behavior ... will do a check and remove this parameter entirely if it is possible.

KaylaBrady: Removed!

boringcactus (approved):

PR #961, opened 2025-05-06
fix: Check that the trip id matches, not tile id

EmmaSimon:
### Summary

_Ticket:_ [Selected trip is not highlighted on stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210176112492141?focus=true)

This was introduced in #952, where upcoming trip IDs were made actually unique. This updates the departure tiles to check the UpcomingTrip's Trip id against the trip filter ID.

iOS

android

### Testing

Verified that selected trips are highlighted again


BrandonTR (approved):

boringcactus (commented):
Won’t this mark both instances of the trip as being selected? I think we may need to check `tripFilter.stopSequence` as well.

PR #963, opened 2025-05-07
feat: clean up after group by direction

boringcactus:
### Summary

_Ticket:_ [Group by Direction | Cleanup old nearby data codepaths](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210113865246640?focus=true)

It’s finally time.

iOS

android

### Testing

Checked that all tests pass.


iosApp/iosApp/Localizable.xcstrings
```diff
@@ -11339,6 +11339,7 @@
       }
     },
     "Open for more arrivals" : {
+      "extractionState" : "stale",
```
boringcactus, line 11342: This string was only used in the StopDeparturesSummaryList, although it’s still used in Android, so there’s probably a place it still belongs on iOS.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/PatternSortingTest.kt
```diff
@@ -1,430 +1,12 @@
...
 
-import com.mbta.tid.mbta_app.model.NearbyStaticData.StaticPatterns
-import io.github.dellisd.spatialk.geojson.Position
 import kotlin.test.BeforeTest
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.time.Duration.Companion.minutes
-import kotlinx.datetime.Clock
 
 class PatternSortingTest {
```
boringcactus, line 5: There were no group-by-direction tests here, so this class is now empty; if it would make sense to port some of the tests instead of deleting them, I can take a look at doing that.

iosApp/iosApp/Pages/Map/HomeMapViewHandlerExtension.swift
```diff
@@ -219,15 +219,15 @@ extension HomeMapView {
...
+        let routeCardData = nearbyVM.routeCardData
+        let route = routeCardData?.first(where: { $0.lineOrRoute.containsRoute(routeId: vehicle.routeId) })
+        let stop = route?.stopData.first
+        let allTrips = stop?.data.flatMap(\.upcomingTrips)
+        let trip = allTrips?.first(where: { upcoming in
             upcoming.trip.id == tripId
         })
         let stopSequence = trip?.stopSequence
-        let routeId = trip?.trip.routeId ?? vehicle.routeId ?? patterns?.routeIdentifier
+        let routeId = trip?.trip.routeId ?? vehicle.routeId ?? route?.lineOrRoute.id
```
KaylaBrady, line 230: 👍 I was worried at first that using `lineOrRoute.id` would cause issues on the GL, but see that `patterns?.routeIdentifier` would also return the line id.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
```diff
@@ -227,8 +226,15 @@ object RouteFeaturesBuilder {
         val targetRouteData = stopMapData.routeShapes.filter { filterRoutes.contains(it.routeId) }
 
         if (targetRouteData.isNotEmpty()) {
-            return departures?.let {
-                val targetRoutePatternIds = departures.upcomingPatternIds
+            return routeCardData?.let {
+                val targetRoutePatternIds =
+                    routeCardData
+                        .asSequence()
```
KaylaBrady, line 232: 👍 nice use of sequence!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -558,6 +558,12 @@ data class RouteCardData(
                 is Line -> Direction.getDirectionsForLine(globalData, stop, patterns)
                 is Route -> Direction.getDirections(globalData, stop, this.route, patterns)
             }
+
+        fun containsRoute(routeId: String?) =
```
KaylaBrady, line 552: 👍 nice helper!

iosApp/iosApp/Localizable.xcstrings
KaylaBrady: I hear "Open for more arrivals" on VoiceOver when I hover on a row in the old route card. If not trivial to add here, can you make a follow-up for it?

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/PatternSortingTest.kt
KaylaBrady: That was an oversight on my part when adding the route card pattern sorting functionality! I think it would be nice to have at least the basics covered in a test here, even if just one test for each of the public functions.

KaylaBrady (approved):
Awesome work! Really appreciate how you broke down the commits here.

iosApp/iosApp/Localizable.xcstrings
boringcactus: Added here.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsPageTest.kt
```diff
@@ -25,12 +25,13 @@ class StopDetailsPageTest : KoinTest {
 
     val koinApplication = testKoinApplication()
 
+    @OptIn(MapboxExperimental::class)
```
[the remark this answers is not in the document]
boringcactus: Yes, because StopDetailsPage is marked with `@MapboxExperimental` so not having it has been causing a compiler warning for a long time, although that’s the wrong marker anyway (it should be `@OptIn(MapboxExperimental::class)`) and it’s no longer necessary.

PR #967, opened 2025-05-08
fix(iOS): Get stop details map stop taps working

EmmaSimon:
### Summary

_Ticket:_ [iOS | [Fix: tapping on stops on the filtered stop details map doesn't do anything]](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210177544414522?focus=true)

In a kind of absurd turn of events, the `.simultaneousGesture(TapGesture())` modifier on the `DepartureTile` button was the thing that was, somehow, preventing taps on stops in the map from registering. This isn't actually a problem for the scrolling issue, because for some other arcane SwiftUI reason, scrolls that happen in horizontally scrolling containers don't seem to register as taps in the same way that scrolls in vertically scrolling containers do.

I also added a handful more `.simultaneousGesture(TapGesture())`s to a few other buttons I noticed could be triggered on scroll.

iOS

### Testing

Tested manually on the simulator and on a physical device

### DIFF
--- a/iosApp/iosApp/ComponentViews/AlertCard.swift
+++ b/iosApp/iosApp/ComponentViews/AlertCard.swift
@@ -99,6 +99,7 @@ struct AlertCard: View {
                     .frame(minHeight: 44)
                     .background(color)
                     .clipShape(.rect(cornerRadius: 8.0))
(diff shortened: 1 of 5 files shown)


BrandonTR (approved):
How strange!

PR #969, opened 2025-05-08
refactor: split direction toggle from tiles in filtered stop details

boringcactus:
### Summary

_Ticket:_ [Group by Direction | Refactor RouteStopData and Leaf to include more context](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210114621155024?focus=true)

This just barely fit into the one point time box. I don’t know if I actually think it’s worth it, though.

iOS

android

### Testing

Checked that all tests still pass and the app isn’t obviously broken.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -105,12 +105,15 @@ data class RouteCardData(
                 data
                     .flatMap { it.alertsHere }
                     .filter { alert -> alert.effect == Alert.Effect.ElevatorClosure }
+                    .distinct()
 
         val hasElevatorAlerts: Boolean
             get() = elevatorAlerts.isNotEmpty()
     }
 
     data class Leaf(
+        val lineOrRoute: LineOrRoute,
+        val stop: Stop,
         val directionId: Int,
         val routePatterns: List<RoutePattern>,
         val stopIds: Set<String>,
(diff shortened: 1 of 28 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -105,12 +100,15 @@ data class RouteCardData(
                 data
                     .flatMap { it.alertsHere }
                     .filter { alert -> alert.effect == Alert.Effect.ElevatorClosure }
+                    .distinct()
```
boringcactus, line 103: This fixes a bug where elevator closures get two cards because they exist in both leaves’ `alertsHere`.

EmmaSimon (approved):
Thanks for getting this into a good place!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
EmmaSimon: Just to double check, does this also get rid of the first "Elevator closure" card [here](https://mbta.slack.com/archives/C082U4THN5A/p1747054800162369), which I assume is coming from the `alertsHere` list?

boringcactus: No, it does not.

EmmaSimon: https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210226994170849?focus=true

boringcactus: Now it does.

PR #971, opened 2025-05-09
chore: Manual dependency updates (part 2)

KaylaBrady:
### Summary

Ticket: [Manual dependency updates](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209367900968825?focus=true)

What is this PR for?
* Android Gradle Plugin is updated to latest [supported](https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#kotlin-2-0-0-and-later) by current Kotlin version
* Gradle itself is updated to meet Android Gradle Plugin requirements
* buildSrc serialization plugin is updated to match Gradle version's embedded Kotlin

I opted not to do the sentry upgrade yet because [the next sentry-kotlin-multiplatform release ](https://github.com/getsentry/sentry-kotlin-multiplatform/blob/main/CHANGELOG.md#unreleased) will upgrade the Java SDK to v8.11.1 which will match[ sentry-android-gradle-plugin 5.0.0](https://github.com/getsentry/sentry-android-gradle-plugin/releases/tag/5.5.0), at which point we can upgrade without the mismatched version crash that led to reverting the upgrade previously in https://github.com/mbta/mobile_app/pull/950. 

Note:
There are now a few deprecation warnings related to build task functionality that will removed in Gradle 9. Leaving those upgrades out of scope for now since It would take me a bit to dig into these tasks, and they do still work.
* Task "bomAndroid" - `'getter for dependencyProject: Project' is deprecated.`
* Task "bomCycloneDxCliDownload" and "bomIosSwiftPM" - `exec(Action<in ExecSpec>): ExecResult' is deprecated. `

iOS

android

### Testing

What testing have you done?
* Ran iOS & android locally & confirmed both worked as expected.


buildSrc/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
     `kotlin-dsl`
-    kotlin("plugin.serialization") version "1.9.22"
+    kotlin("plugin.serialization") version "2.0.20"
```
KaylaBrady, line 3: Just used the version that matched in the warning message when I set this to `2.1.20`

boringcactus (approved):

PR #972, opened 2025-05-09
feat(Android): expose route search behind feature flag

boringcactus:
### Summary

_Ticket:_ [🤖 | Favorites | Expose route search](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816289542473?focus=true)

We’ll need to go in later and make sure the search actually looks as intended, but that’s explicitly out of scope for this ticket.

iOS

android

### Testing

Added unit tests to ensure that routes aren’t shown unless the flag is enabled and that route pills are shown to make bus routes easy to select.


KaylaBrady (approved):
Sweet! As a non-android daily user it took me a minute to realize the route results were blocked by the keyboard and that I needed to collapse it. Maybe not a problem for regular android users, but in any case something that can be considered far later than now!

PR #979, opened 2025-05-13
feat(Android): Add empty favorites tab

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Favorites | Add empty favorites tab](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816289542468?focus=true)

Add a new tab behind the favorites feature flag on Android. I managed to get the map not to reload by keeping the initial sheet entrypoint as a separate piece of state within `ContentView`, which is then passed down into `MapAndSheetPage` and used as the initial route in the sheet's `NavHost`.

Screenshot_20250512_204643

iOS

android

### Testing

Updated existing tests to pass


androidApp/src/main/java/com/mbta/tid/mbta_app/android/SheetRoutes.kt
```diff
@@ -7,11 +7,27 @@ import androidx.navigation.toRoute
...
 import com.mbta.tid.mbta_app.model.TripDetailsFilter
+import kotlin.reflect.typeOf
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.encodeToString
 
 @Serializable
 sealed class SheetRoutes {
+
+    // An enum is used to represent the entrypoint so that it can be used in rememberSaveable,
+    // and to make clear that there is a specific subset of root nav routes that the sheet can use.
```
boringcactus, line 17: This might not work, but I wonder if having a nested `sealed interface Entrypoint` and then making `Favorites : SheetRoutes(), Entrypoint` would make both the `rememberSaveable` and the subsetting behave correctly. without requiring a separate value to exist.

boringcactus (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/SheetRoutes.kt
EmmaSimon: Huh, yeah good idea, I'll give it a shot!

EmmaSimon: In order to get the `rememberSaveable` working I needed to create a `Saver`, but I like this better.

PR #980, opened 2025-05-13
fix(ios.StopDetailsPage): Fix thrashing when stopFilter changes

KaylaBrady:
### Summary

_Ticket:_ [iOS | Group By Direction | All trips are shown when navigating from filtered stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210162265230760?focus=true)

What is this PR for?
This PR includes two fixes: 
1. When navigating from unfiltered stop details => filtered stop details, don't flicker between 3 and all upcoming trips ([slack thread](https://mbta.slack.com/archives/C05QMG9GS9M/p1747133184066279))
2. When navigating from filtered stop details => unfiltered stop details, don't show the full list of all upcoming trips on initial load

iOS

android

### Testing

What testing have you done?
* Ran locally & confirmed issue fixed in both directions
* Added a unit test which isn't great, but will at least create a failure if we change this behavior in the future. 

https://github.com/user-attachments/assets/66e95bc8-cbab-47dc-979b-a3a2651def78

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
+++ b/iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
@@ -10,12 +10,23 @@ import Shared
 import SwiftPhoenixClient
 import SwiftUI
 
+struct RouteCardParams: Equatable {
+    let alerts: AlertsStreamDataResponse?
+    let global: GlobalResponse?
+    let now: Date
+    let pinnedRoutes: Set<String>
+    let stopData: StopData?
+    let stopFilter: StopDetailsFilter?
+    let stopId: String
+}
+
 struct StopDetailsPage: View {
     var filters: StopDetailsPageFilters
 
     // StopDetailsPage maintains its own internal state of the departures presented.
     // This way, when transitioning between one StopDetailsPage and another, each separate page shows
     // their respective departures rather than both showing the departures for the newly presented stop.
+
     @State var internalRouteCardData: [RouteCardData]?
     @State var now = Date.now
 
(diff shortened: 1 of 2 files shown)


iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -80,22 +80,27 @@ struct StopDetailsPage: View {
...
-                updateDepartures()
+                updateDepartures(stopFilter: stopFilter)
             }
             .onChange(of: filters) { nextFilters in setTripFilter(filters: nextFilters) }
             .onChange(of: internalRouteCardData) { _ in
                 let nextStopFilter = setStopFilter()
                 setTripFilter(filters: .init(stopId: stopId, stopFilter: nextStopFilter, tripFilter: tripFilter))
             }
-            .task(id: stopId) {
+            .task(id: "\(stopId) \(stopFilter) \(tripFilter)") {
```
KaylaBrady, line 100: planning to make this more like NearbyTransitView and have the task only set now and an onChange that takes all the necessary params for updateDepartures instead of this approach

iosApp/iosApp/Pages/StopDetails/StopDetailsPage.swift
```diff
@@ -178,19 +198,32 @@ struct StopDetailsPage: View {
...
             Task { @MainActor in
-                nearbyVM.setRouteCardData(stopId, nextRouteCardData)
+                nearbyVM.setRouteCardData(routeCardParams.stopId, nextRouteCardData)
                 internalRouteCardData = nextRouteCardData
             }
         }
     }
+
+    // Testing convenience
+    func updateDepartures() {
```
KaylaBrady, line 220: I don't love this, but improving the existing tests seems hairy since they currently use `sleep`

EmmaSimon (approved):

PR #983, opened 2025-05-13
feat: show GL pills even if not branching

boringcactus:
### Summary

_Ticket:_ [Group by Direction | Show GL route pills in Nearby Transit card on branches](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210114621155036?focus=true)

This was indeed easy after the refactor to move the lineOrRoute into the leaf.

iOS

android

### Testing

Added unit tests in shared code and on each platform. Manually verified that pills are shown correctly on each platform.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -167,26 +167,28 @@ data class RouteCardData(
                     it.significance >= AlertSignificance.Secondary
             } ?: alertsDownstream.firstOrNull()
 
+        private data class PotentialService(val routeId: String, val headsign: String)
+
         /**
-         * Get all headsigns that might be shown for this leaf. For bus, the only headsigns that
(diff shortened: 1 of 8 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -167,26 +167,28 @@ data class RouteCardData(
...
          */
-        fun potentialHeadsigns(
+        private fun potentialService(
             now: Instant,
             representativeRoute: Route,
             globalData: GlobalResponse?,
-        ): Set<String> {
-            val potentialHeadsigns = mutableSetOf<String>()
+        ): Set<PotentialService> {
+            val potentialService = mutableSetOf<PotentialService>()
```
boringcactus, line 184: Including the routes here should fix a bug that we may hit in Silver Line grouping: if the headsigns are all the same but the routes aren’t, we still need to treat the leaf as branched so that we can label the routes separately. It also makes picking the route in the no-trips case substantially easier.

KaylaBrady (approved):

PR #986, opened 2025-05-14
refactor: move more logic into TripDetailsStopList.Entry.format

boringcactus:
### Summary

_Ticket:_ [Fix: display arrival time at shuttle boundary if possible to exit](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210239404411208?focus=true)

Whatever approach we wind up taking will involve refining the logic of how we decide whether to use the disruption or the formatted time, so it’ll be helpful to have one home for that logic rather than two.

iOS

android

### Testing

Checked that all tests still pass.

### DIFF
--- a/iosApp/iosApp/Pages/StopDetails/TripHeaderCard.swift
+++ b/iosApp/iosApp/Pages/StopDetails/TripHeaderCard.swift
@@ -20,7 +20,7 @@ enum TripHeaderSpec {
 // swiftlint:disable:next type_body_length
 struct TripHeaderCard: View {
     let spec: TripHeaderSpec
-    let tripId: String
+    let trip: Trip
     let targetId: String
     let routeAccents: TripRouteAccents
     let onTap: (() -> Void)?
@@ -126,7 +126,7 @@ struct TripHeaderCard: View {
     @ViewBuilder private func vehicleDescription(
         _ vehicle: Vehicle, _ stop: Stop, _ entry: TripDetailsStopList.Entry?, _ atTerminal: Bool
     ) -> some View {
-        if vehicle.tripId == tripId {
+        if vehicle.tripId == trip.id {
             VStack(alignment: .leading, spacing: 2) {
                 VStack(alignment: .leading, spacing: 2) {
                     vehicleStatusDescription(vehicle.currentStatus, atTerminal)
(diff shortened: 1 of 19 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -23,7 +23,7 @@ constructor(val tripId: String, val stops: List<Entry>, val startTerminalEntry:
         val prediction: Prediction?,
         // The prediction stop can be the same as `stop`, but it can also be a child stop which
         // contains more specific boarding information for a prediction, like the track number
-        val predictionStop: Stop?,
+        val predictionStop: Stop? = stop.takeIf { prediction != null },
```
boringcactus, line 26: A prior draft involved creating a new `scheduleStop`, which would be completely unwieldy with no default, so I gave both `scheduleStop` and `predictionStop` a default, and then I noticed that the `Entry.stop` is not actually the parent station and in fact will always be the schedule stop if the prediction stop is different, so I removed `scheduleStop`, but I think giving the prediction stop a default value makes the tests simpler, so I’m leaving that in.

iosApp/iosApp/Pages/StopDetails/TripHeaderCard.swift
```diff
@@ -304,14 +304,15 @@ struct TripHeaderCard: View {
...
-            default: .some(formatted)
-            }
+        guard let formatted = entry.format(trip: trip, now: now.toKotlinInstant(), routeType: routeAccents.type)
+        else { return nil }
+        switch onEnum(of: formatted) {
+        case let .disruption(disruption): return .disruption(
+                .init(alert: disruption.alert),
+                iconName: disruption.iconName
+            )
+        case let .some(some): return .some(some.trips[0].format)
```
BrandonTR, line 314: Just checking, we should never have an empty trips collection here right?

iosApp/iosApp/Pages/StopDetails/TripStopRow.swift
```diff
@@ -251,17 +254,22 @@ struct TripStopRow: View {
...
-        } else {
-            .some(stop.format(now: now, routeType: routeAccents.type))
+    var upcomingTripViewState: UpcomingTripView.State? {
+        guard let formatted = stop.format(trip: trip, now: now, routeType: routeAccents.type) else { return nil }
+        switch onEnum(of: formatted) {
+        case let .disruption(disruption): return .disruption(
+                .init(alert: disruption.alert),
+                iconName: disruption.iconName
+            )
+        case let .some(some): return .some(some.trips[0].format)
```
BrandonTR, line 264: Same q

BrandonTR (approved):
Just one q, otherwise lgtm!

iosApp/iosApp/Pages/StopDetails/TripHeaderCard.swift
boringcactus: Yes, if `format` returns an `UpcomingFormat.Some` then it will always have exactly one `FormattedTrip` in `trips`.

PR #999, opened 2025-05-16
fix: Don't show redundant stops if atypical trips aren't visible

EmmaSimon:
### Summary

_Ticket:_ [Only display additional stop if next 3 trips are different](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210239404411211?focus=true)

This actually was about as straightforward as we expected, I just wasn't finding any good examples in prod to test with early on. I followed the lead of `potentialService` and used `isBus` to determine if it was safe to check the next 2 or 3 upcoming trips. Also moved the 2 and 3 row values into constants, since we were starting to have lots of duplicate magic numbers floating around.

This does not cover the "Service ended" case mentioned in the QA, I think that will need some other special handling, but since that's reliant on the formatting to decide whether service is over, I'm less sure how we'd deal with it.

| Before | After |
| --- | --- |
| Screenshot_20250516_181448 | Screenshot_20250516_181346 |

### Testing

Added a new unit test to check the behavior


KaylaBrady (approved):
Appreciate the constants!

PR #1005, opened 2025-05-20
fix: Prevent extra trips from appearing in unfiltered stop details when navigating to filtered

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Group By Direction | All trips are shown when navigating from unfiltered stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210318500071981?focus=true)

Moving the route card state into the filtered view seems to have fixed the problem, I was initially trying to do something more complicated with filtering the flow based on context, but that was causing the loading state to flicker on every filter change.

### Testing

Manually verified that extra trips no longer flash into unfiltered stop cards


KaylaBrady (approved):
Nice!

PR #1007, opened 2025-05-20
fix(ios.StopDetailsFilteredDepartureDetails): Preserve accessibility focus on tile

KaylaBrady:
### Summary

_Ticket:_ [Fix VO focus when navigating to filtered stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210322015597203?focus=true)

What is this PR for?
* Ensure we always focus on the first selected tile. This was a regression introduced by https://github.com/mbta/mobile_app/pull/952 when the id for an `UpcomingTrip` changed.
* Also includes android fixed discussed at dev check-in for resetting scroll position when changing direction

iOS

android

### Testing

What testing have you done?
* Ran locally on iOS and confirmed that VO correctly focused on the first tile when coming from any view. 
    * Thought about writing a ViewInspector test, but [accessibilityFocused](https://github.com/nalexn/ViewInspector/blob/0.10.2/readiness.md) is not yet implemented 
* Ran on android and confirm tile scroll position resets to the start of the list when the direction filter changes

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
@@ -180,7 +180,7 @@ fun MapAndSheetPage(
     val tileScrollState = rememberScrollState()
     val scope = rememberCoroutineScope()
 
-    LaunchedEffect(previousNavEntry) {
+    LaunchedEffect(currentNavEntry) {
         if (
             previousNavEntry is SheetRoutes.StopDetails &&
                 currentNavEntry is SheetRoutes.StopDetails &&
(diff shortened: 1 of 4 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -180,7 +180,7 @@ fun MapAndSheetPage(
     val tileScrollState = rememberScrollState()
     val scope = rememberCoroutineScope()
 
-    LaunchedEffect(previousNavEntry) {
+    LaunchedEffect(currentNavEntry) {
```
EmmaSimon, line 183: This doesn't fix a similar issue where if you tap a tile that's halfway off the screen, on iOS it will scroll that whole tile into view. That might be more complicated on Android though.

BrandonTR (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
KaylaBrady: @EmmaSimon it looks to me like we do scroll the whole tile into view for android too. That may be fixed by the changes to fix the use of `tileData.id` in StopDetailsFilteredDeparturesView

EmmaSimon: Oh, amazing!

PR #1008, opened 2025-05-20
fix(Android): Prevent global data from being reset to null on every filter change

EmmaSimon:
### Summary

_Ticket:_ [🤖 | GL pill flicker in stop details departure cards](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210329839893517?focus=true)

The GL pills on filtered stop details departure tiles were disappearing every time you changed the selected trip or direction toggle. This was happening because the global data was somehow being reset to null in `Leaf.format`, which was causing the route value to be null, even though a route ID existed and the route was meant to be displayed.

This PR moves the global data into a flow in the stop details VM. We store global in the map VM, so there was some precedence for this, and it ensures that we're always using the exact same global data in `stopDetailsManagedVM` that we are in the stop details pages.

android

### Testing

Updated tests to pass


boringcactus (approved):

boringcactus:
Huh, this correctly detected that nothing on iOS changed and then ran the iOS tests anyway.

PR #1009, opened 2025-05-20
ci: really fix filtering for skipping platform specific tests

boringcactus:
### Summary

_Ticket:_ none

As can be seen in [this run](https://github.com/mbta/mobile_app/actions/runs/15145766015) of #1008, even now that the path filtering has been fixed in #1006, the conditions are broken. I am not sure if this will actually work, but if it does, then I will cry.

iOS

android

### Testing

Can’t really be tested.


KaylaBrady (approved):
Seems promising

PR #1013, opened 2025-05-21
chore: add changelog links where dependabot doesn’t pull changelogs automatically

boringcactus:
### Summary

_Ticket:_ none

Sometimes, Dependabot knows how to track down the changelog for dependencies it’s updating, and that’s nice. Other times, we need to hunt down the release notes ourselves, mostly in the Android standard library but sometimes on dependencies that just don’t have that working for whatever reason; for those, it seems like it’d be nice to have a link right there in the diff context, like we already do for the Jetpack Compose BOM.

iOS

android

### Testing

Not testable.


KaylaBrady (approved):
Nice!

PR #1014, opened 2025-05-21
ci: run empty job instead of skipping if platform not changed

boringcactus:
### Summary

_Ticket:_ none

#1009 went slightly too far, because you can’t skip required jobs, and if they aren’t required then the merge queue doesn’t care if they fail. Hopefully, skipping all the steps within the job will get us everything we want. Have I mentioned lately that I don’t think GitHub Actions is very good?

### Testing

`actionlint` doesn’t seem to spot any issues with this, for whatever that’s worth.

### DIFF
--- a/.github/workflows/build-android.yaml
+++ b/.github/workflows/build-android.yaml
@@ -18,6 +18,11 @@ on:
       sentry-environment:
         required: true
         type: string
+      skip-everything:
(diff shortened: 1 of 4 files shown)


EmmaSimon (approved):
Oof, yeah this is fun 😓

PR #1028, opened 2025-05-29
feat(Android): Show favorites in favorites tab

BrandonTR:
### Summary

_Ticket:_ [🤖 Favorites | Show favorites in tab](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816289542486?focus=true)

Displays favorites in favorites tab
Screenshot_20250529_093250

android

### Testing

Added tests for filtering and changes to include redundant stops


androidApp/src/main/res/values/strings.xml
```diff
@@ -118,6 +118,7 @@
     <string name="facility_issue">Facility Issue</string>
     <string name="facility_issue_sentence_case">Facility issue</string>
     <string name="favorites_link">Favorites</string>
+    <string name="no_stops_added" translatable="false">No stops added</string>
```
boringcactus, line 121: I’d rather use `tools:ignore="MissingTranslation"` than `translatable="false"` so it’s clearer that there need to be translations eventually and that once they exist they should be used.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/FavoritesPage.kt
```diff
@@ -1,13 +1,69 @@
...
+        }
+    }
+    LaunchedEffect(nearbyTransit.viewportProvider.isManuallyCentering) {
+        if (nearbyTransit.viewportProvider.isManuallyCentering) {
+            targetLocation = null
+        }
+    }
+    LaunchedEffect(nearbyTransit.viewportProvider.isFollowingPuck) {
+        if (nearbyTransit.viewportProvider.isFollowingPuck) {
+            targetLocation =
```
boringcactus, line 54: Should we abstract out all the `targetLocation` management into a single function so it can be shared between nearby transit and favorites?

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/FavoritesViewModelTest.kt
```diff
@@ -0,0 +1,166 @@
...
+                stop1.id to listOf(routePatternOne.id, routePatternTwo.id),
+                stop2.id to listOf(routePatternOne.id, routePatternTwo.id),
+            ),
+        )
+
+    val koinApplication = testKoinApplication(builder)
+
+    @Test
+    fun testFiltering() = runBlocking {
+        val favoritesVM = FavoritesViewModel()
```
boringcactus, line 121: It looks like this test doesn’t require Android to be running at all, since the view model logic doesn’t rely on being run inside of Compose, which means this could be moved from `androidTest` to `test` to run on a cheaper instance in CI.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/FavoritesViewModel.kt
```diff
@@ -0,0 +1,89 @@
...
+            )
+        routeCardData = filterRouteAndDirection(loadedRouteCardData, global)
+    }
+
+    fun filterRouteAndDirection(
+        routeCardData: List<RouteCardData>?,
+        global: GlobalResponse,
+    ): List<RouteCardData>? {
+        return routeCardData?.filter { data ->
+            data.stopData =
```
boringcactus, line 66: If you’re returning a new list anyway, it seems like this wouldn’t have to be an in-place mutation, and you could `.copy()` instead to avoid having to make `stopData` a `var`.

PR #1032, opened 2025-05-30
fix(RouteCardData.potentialService): Evaluate all upcoming trips for bus in unfiltered stop details

KaylaBrady:
### Summary

_Ticket:_ No ticket, based on [report in slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1748627972953529)

What is this PR for?

We are sometimes unexpectedly treating the 15 as unbranched and always showing a headsign of "St Peter's Square" even for "Fields Corner" trips. This seems to only occur if the next two trips are both heading to "St Peter's Square". 

We want to only look at the first two bus trips to determine whether to use branching rules in nearby transit, but in filtered stop details, we should look at all trips.

iOS

android

### Testing

What testing have you done?
* Added a unit test
* Ran locally and confirmed I see other headsigns for the 15, even when the next 2 trips are both to "St Peter's Square"
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -195,6 +195,7 @@ data class RouteCardData(
             now: Instant,
             representativeRoute: Route,
             globalData: GlobalResponse?,
+            context: Context,
         ): Set<PotentialService> {
             val potentialService: MutableMap<Pair<String, String>, MutableSet<String>> =
                 mutableMapOf()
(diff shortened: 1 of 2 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -195,14 +195,17 @@ data class RouteCardData(
...
         ): Set<PotentialService> {
             val potentialService: MutableMap<Pair<String, String>, MutableSet<String>> =
                 mutableMapOf()
             val cutoffTime = now + 120.minutes
             val tripsUpcoming = upcomingTrips.filter { it.isUpcomingWithin(now, cutoffTime) }
             val isBus = representativeRoute.type == RouteType.BUS
             val tripsToConsider =
-                if (isBus) tripsUpcoming.take(TYPICAL_LEAF_ROWS) else tripsUpcoming
+                if (isBus && context == Context.StopDetailsFiltered) tripsUpcoming
+                else tripsUpcoming.take(TYPICAL_LEAF_ROWS)
```
boringcactus, line 207: Actually, looking at this a little closer, is this behavior change on non-bus routes something that we’re sure we want?

KaylaBrady: Ooh great catch

boringcactus (approved):

PR #1048, opened 2025-06-04
feat: hide irrelevant bus stops in layer

boringcactus:
### Summary

_Ticket:_ [Omit bus stops on unselected routes below zoom 15](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210277504456915?focus=true)

An alternative approach to #1031 that feels more responsive.

iOS

android

### Testing

Checked that the bus stops disappear and reappear as called for. Added new unit tests.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
@@ -100,6 +100,7 @@ fun HomeMapView(
     val coroutineScope = rememberCoroutineScope()
     val layerManager = remember { LazyObjectQueue<MapLayerManager>() }
     val selectedStop by viewModel.selectedStop.collectAsState(null)
+    val stopFilter by viewModel.stopFilter.collectAsState(null)
 
     val configLoadAttempted by viewModel.configLoadAttempted.collectAsState(initial = false)
     val railRouteShapes by viewModel.railRouteShapes.collectAsState(initial = null)
@@ -126,6 +127,11 @@ fun HomeMapView(
     val context = LocalContext.current
     val density = LocalDensity.current
     val layoutDirection = LocalLayoutDirection.current
+
+    val stopLayerGeneratorState =
+        remember(selectedStop, stopFilter) {
+            StopLayerGenerator.State(selectedStopId = selectedStop?.id, stopFilter = stopFilter)
+        }
 
     fun handleStopClick(map: MapView, point: Point): Boolean {
         val pixel = map.mapboxMap.pixelForCoordinate(point)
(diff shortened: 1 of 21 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -100,6 +100,7 @@ fun HomeMapView(
     val coroutineScope = rememberCoroutineScope()
     val layerManager = remember { LazyObjectQueue<MapLayerManager>() }
     val selectedStop by viewModel.selectedStop.collectAsState(null)
+    val stopFilter by viewModel.stopFilter.collectAsState(null)
```
EmmaSimon, line 103: Would it make sense to update the VM flow to be of `StopLayerGenerator.State` (I think it would need to be updated to contain the whole stop object instead of just the id) closer to how it works on iOS? I'm unsure if it would make a noticeable difference or if compose is smart enough to not do this, but I worry that every time `handleNavChange` is run, it results in two distinct changes to the generator state, and the layer manager is unnecessarily run twice. If that could also include the list of route shapes to filter too that would be great, but that's definitely an out of scope refactor.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/map/style/ExpEval.kt
```diff
@@ -180,6 +336,12 @@ private data class EvaluationContext(val featureProperties: FeatureProperties, v
     }
 }
 
+@JvmName("evaluateString")
```
EmmaSimon, line 339:
```suggestion
@JvmName("evaluateBoolean")
```

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/map/style/ExpEval.kt
```diff
@@ -22,11 +27,43 @@ private fun JSONValue.toKotlin(): JsonElement =
...
-private data class EvaluationContext(val featureProperties: FeatureProperties, val zoom: Double) {
+private data class EvaluationContext(
+    val featureProperties: FeatureProperties,
+    val zoom: Double,
+    val letBindings: Map<String, JsonElement> = emptyMap(),
+) {
     fun evaluate(json: JsonElement): JsonElement {
         return if (json is JsonArray) {
             when (val operator = (json.firstOrNull() as? JsonPrimitive)?.contentOrNull) {
+                "array" -> {
```
EmmaSimon, line 38: Thanks for adding the rest of these 🙏 

Should the default state throw an exception so it's required to write a new case if you start using a new one, or are there valid cases for it to pass through the json?

EmmaSimon (approved):
Nice work. Nothing blocking, I think you're pretty done with this ticket, which is fair! Thanks for getting close to an optimal approach.

PR #1049, opened 2025-06-04
build: upgrade Xcode and SwiftPM dependencies

boringcactus:
### Summary

_Ticket:_ none

Since the ViewInspector fix for Xcode 16.3 has been shipped, we can now upgrade to that version, which might be enough to unblock #922.

### Testing

Checked that the tests pass.


BrandonTR (approved):
🎉

PR #1050, opened 2025-06-04
feat(android.RouteStopListView): Collpase non-typical stops

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Favorites | Collapse non-typical bus stops](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210389717064826?focus=true)

What is this PR for?
* Collapses non-typical stops in the route list. Leaves some placeholders for improving the accessibility VoiceOver behavior after further discussion with design - will make a ticket to follow up on that.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally 
image
image

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
@@ -2,6 +2,7 @@ package com.mbta.tid.mbta_app.android.component
 
 import android.content.Context
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Arrangement
@@ -134,6 +135,8 @@ fun StopListRow(
                 }
                 if (stopRowStyle != StopRowStyle.StandaloneStop) {
                     RouteLine(routeAccents, stateBefore, stateAfter, targeted)
+                } else {
+                    Box(modifier = Modifier.width(20.dp).background(colorResource(R.color.key))) {}
                 }
                 Column(
                     Modifier.padding(vertical = 12.dp).padding(start = 16.dp).semantics {
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
@@ -35,6 +35,7 @@ import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.heading
 import androidx.compose.ui.semantics.isTraversalGroup
 import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.testTag
 import androidx.compose.ui.semantics.traversalIndex
 import androidx.compose.ui.unit.dp
 import com.mbta.tid.mbta_app.android.R
@@ -146,11 +147,13 @@ fun StopListRow(
                                 painter = painterResource(R.drawable.mbta_logo),
                                 contentDescription = null,
                                 tint = Color.Unspecified,
+                                modifier = Modifier.semantics { testTag = "mbta_logo" },
                             )
                         } else if (stop.vehicleType == RouteType.BUS) {
                             Icon(
                                 painter = painterResource(R.drawable.stop_bus),
                                 contentDescription = null,
+                                modifier = Modifier.semantics { testTag = "stop_bus" },
                                 tint = Color.Unspecified,
                             )
                         } else {
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
@@ -57,12 +57,11 @@ import com.mbta.tid.mbta_app.model.RouteType
 import com.mbta.tid.mbta_app.model.Stop
 import com.mbta.tid.mbta_app.model.UpcomingFormat
 
-enum class StopRowStyle {
-    FirstLineStop,
-    MidLineStop,
-    LastLineStop,
-    StandaloneStop,
-}
+class StopPlacement(
+    val isFirst: Boolean = false,
+    val isLast: Boolean = false,
+    val includeLineDiagram: Boolean = true,
+)
 
 @Composable
 fun StopListRow(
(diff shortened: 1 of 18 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
```diff
@@ -70,12 +82,34 @@ data class RouteDetailsStopList(val stops: List<Entry>) {
...
 
-                RouteDetailsStopList(stops)
+                val segments = splitIntoSegments(stops)
+
+                RouteDetailsStopList(segments)
             }
+
+        fun splitIntoSegments(entries: List<Entry>): List<Segment> {
+            val (accSegments, lastWipSegment) =
+                entries.fold(Pair<List<Segment>, Segment?>(listOf(), null)) {
```
boringcactus, line 98: I think just using a `MutableList<Segment>` and a `for` loop might be more legible here than `fold` with a `Pair` accumulator.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopListTest.kt
```diff
@@ -184,17 +184,27 @@ class RouteDetailsStopListTest {
...
         assertEquals(
             RouteDetailsStopList(
                 listOf(
-                    RouteDetailsStopList.Entry(mainStop, connectingRoutes = listOf(connectingRoute))
+                    RouteDetailsStopList.Segment(
+                        listOf(
+                            RouteDetailsStopList.Entry(
+                                mainStop,
+                                connectingRoutes = listOf(connectingRoute),
+                                patterns = listOf(),
```
boringcactus, line 203: I would expect `patterns = listOf(pattern)` here; if it shouldn’t be there, there should probably be a comment explaining why.

androidApp/src/main/res/values/strings.xml
```diff
@@ -160,6 +160,8 @@
     <string name="inbound">Inbound</string>
     <string name="is_stops_away_from">%1$s is %2$d stops away from %3$s</string>
     <string name="later_today">later today</string>
+    <string name="less_common_stop">Less common stop</string>
+    <string name="less_common_stops">&lt;b&gt;%1$d&lt;/b&gt;  less common stops</string>
```
boringcactus, line 164:
```suggestion
    <string name="less_common_stops">&lt;b&gt;%1$d&lt;/b&gt; less common stops</string>
```

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
```diff
@@ -62,26 +74,26 @@ fun StopListRow(
...
+    descriptor: @Composable () -> Unit = {},
     rightSideContent: @Composable RowScope.(Modifier) -> Unit = {},
 ) {
     val context = LocalContext.current
 
     val stateBefore =
-        when {
-            firstStop -> RouteLineState.Empty
+        when (stopRowStyle) {
+            StopRowStyle.FirstLineStop -> RouteLineState.Empty
```
boringcactus, line 91: I would expect `StandaloneStop` to give `stateBefore` `Empty` and `stateAfter` `Empty`, although the new `RouteLine` conditional probably can‘t be pushed down into the `RouteLine` component itself so it may not actually make a difference.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListView.kt
```diff
@@ -114,17 +116,45 @@ fun RouteStopListView(
...
-                            stop.stop,
-                            onClick = { onClick(stop) },
-                            routeAccents = TripRouteAccents(lineOrRoute.sortRoute),
-                            modifier = Modifier.minimumInteractiveComponentSize(),
-                            connectingRoutes = stop.connectingRoutes,
-                            firstStop = index == 0,
-                            lastStop = index == stopList.stops.lastIndex,
-                            rightSideContent = { modifier -> rightSideContent(stop, modifier) },
-                        )
+                    stopList.segments.withIndex().map { (segmentIndex, segment) ->
```
boringcactus, line 119: I’m not sure that a `map` returning nothing is better than just a `for` loop.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/CollapsableStopList.kt
```diff
@@ -0,0 +1,157 @@
...
+                            fadeIn(animationSpec = tween(500)) togetherWith
+                                fadeOut(animationSpec = tween(500))
+                        },
+                    ) {
+                        Row(
+                            verticalAlignment = Alignment.CenterVertically,
+                            horizontalArrangement = Arrangement.Center,
+                            modifier = Modifier.padding(start = 22.dp).width(20.dp),
+                        ) {
+                            if (it) {
```
boringcactus, line 96: I find it confusing that this `it` is the lambda argument from the `AnimatedContent` rather than the `Row`; naming that argument would make it clearer.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/CollapsableStopList.kt
```diff
@@ -0,0 +1,157 @@
...
+                    }
+                }
+
+                if (!isLastSegment) {
+                    HaloSeparator()
+                }
+            }
+
+            if (stopsExpanded) {
+                segment.stops.map { stop ->
```
boringcactus, line 141: `.map` with a lambda returning `Unit` just builds a list of `Unit`s for no reason. I think just a `for` loop would make more sense.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
```diff
@@ -93,7 +105,7 @@ fun StopListRow(
                 .height(IntrinsicSize.Min)
                 .defaultMinSize(minHeight = 48.dp)
         ) {
-            if (!lastStop && !targeted && disruption == null) {
+            if (stopRowStyle != StopRowStyle.LastLineStop && !targeted && disruption == null) {
```
boringcactus, line 108: What if the last stop in the entire list is a standalone stop?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListView.kt
EmmaSimon: There's also `.forEach`

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
KaylaBrady: I could almost push it down into the RouteLine, except the afterState is separately used when there is a disruption and we don't use the RouteLine component. Opted to add a case for standalone stops here via the new `includeLineDiagram` prop.

boringcactus (approved):

PR #1053, opened 2025-06-05
ci: retry Android instrumented tests

boringcactus:
### Summary

_Ticket:_ none

https://github.com/mbta/mobile_app/pull/1050 had four consecutive flaky Android instrumented test failures in the merge queue.

### Testing

shellcheck doesn’t complain, and this is the same function that worked fine in Xcode Cloud.

---
  - https://app.asana.com/0/0/1210425090924778

### DIFF
--- a/bin/android-instrumented-test-ci.sh
+++ b/bin/android-instrumented-test-ci.sh
@@ -4,7 +4,30 @@
 # https://github.com/ReactiveCircus/android-emulator-runner/issues/316#issuecomment-1665866511
 
 set +e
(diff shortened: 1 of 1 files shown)


KaylaBrady (approved):
TY! Added a reference to this PR in the [systematic flaky test rethink ticket](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210425090924778?focus=true) so that we can further consider the approach, but definitely seems like an improvement worth having in the meantime

PR #1057, opened 2025-06-09
feat(Android): Favorites subway route picker flow

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Favorites | Add stops flow - subway](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210229155855227?focus=true)

Add the new route picker page, and make it possible to select favorite stops on the route page.

 I added two new sealed classes for route picker and details navigation, a `RouteDetailsContext` which changes the behavior of route details and route picker depending on whether you got there through the favorites flow or the search results (and in the future, the standard route details flow). And also `RoutePickerPath`, which determines the options listed under the route picker, right now the only one partially implemented is the `Root` path, which displays subway, and in the future, links to the other modes.

I also added some helper extension functions to `NavHostController` for easier access of fields in the current sheet route.

The ticket mentions LTR transitions if trivial, but I didn't and I think it may end up being impossible, since the background color of the sheet (and crucially, the very top including the drag handle) is controlled by the sheet rather than the page content, I don't think there's a straightforward way to animate those transitions. Unless we can figure out some way to do the transitions entirely in the sheet.

Screenshot_20250609_114118Screenshot_20250609_114139

iOS

android

### Testing

Added tests for new components, and for new filtering method on global response.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/SheetRoutes.kt
```diff
@@ -27,7 +29,12 @@ sealed class SheetRoutes {
         val tripFilter: TripDetailsFilter?,
     ) : SheetRoutes()
 
-    @Serializable data class RouteDetails(val routeId: String) : SheetRoutes()
+    @Serializable
+    data class RoutePicker(val path: RoutePickerPath, val context: RouteDetailsContext) :
```
KaylaBrady, line 33: question: Does the RoutePickerPath only matter if RouteDetailsContext is favorites? Wondering if it can live within `RouteDetailsContext` or if there is a benefit to keeping it separate

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ActionButton.kt
```diff
@@ -26,6 +26,7 @@ enum class ActionButtonKind(
 ) {
     Back(14.dp, R.string.back_button_label, R.drawable.fa_chevron_left),
     Close(10.dp, R.string.close_button_label, R.drawable.fa_xmark),
+    Plus(24.dp, R.string.add_stops, R.drawable.plus),
```
KaylaBrady, line 29: nitpick(non-blocking): The add_stops text feels more specific than the generic plus button here, but this seems fine since there aren't any other uses of it yet.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/NavTextButton.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.unit.dp
+import com.mbta.tid.mbta_app.android.R
+import com.mbta.tid.mbta_app.android.util.Typography
+
+@Composable
+fun NavTextButton(string: String, onTap: () -> Unit) {
```
KaylaBrady, line 16: 👍 appreciate that this is separated out!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -240,16 +247,34 @@ fun MapAndSheetPage(
     }
 
     fun handleStopNavigation(stopId: String) {
+        if (navController.selectedStopId == stopId) return
```
KaylaBrady, line 250: question (non-blocking): Is this related or has been causing separate problems?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/NavHostControllerExtension.kt
```diff
@@ -0,0 +1,34 @@
...
+inline fun <reified T : SheetRoutes> NavHostController.currentRouteMatches(): Boolean =
+    this.fromCurrentRoute { it?.let { it::class == T::class } } == true
+
+/**
+ * Convert the current nav stack entry into a SheetRoute, then run the provided lambda on it and
+ * return the result.
+ */
+fun <T> NavHostController.fromCurrentRoute(take: (SheetRoutes?) -> T): T =
+    this.currentBackStackEntry?.let { take(SheetRoutes.fromNavBackStackEntry(it)) } ?: take(null)
+
```
KaylaBrady, line 25: nitpick(non-blocking): This taking a lambda to transform kind of makes it seem like this is doing two separate things where it might not need to. I think it might be clearer to chain the `let` / transformation after the call to convert into a SheetRoute outside this function.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/routePicker/RoutePickerView.kt
```diff
@@ -0,0 +1,84 @@
...
+    errorBannerViewModel: ErrorBannerViewModel,
+) {
+    val globalData = getGlobalData("RoutePickerView.globalData")
+
+    if (globalData == null) {
+        CircularProgressIndicator(Modifier.semantics { contentDescription = "Loading" })
+        return
+    }
+
+    val routes = remember(globalData, path) { globalData.getRoutesForPicker(path) }
```
KaylaBrady, line 45: question(non-blocking): Should this be `rememberSaveable`?

KaylaBrady (approved):

PR #1064, opened 2025-06-11
test(android): break apart some flaky tests

KaylaBrady:
### Summary

_Ticket:_ [🤖 systematic flaky test rethink](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210425090924778?focus=true)

What is this PR for?
This PR breaks apart flaky UI Tests by mocking the ViewModel and other inputs that do complicated logic. Also includes some improvements to the logging of RetryableComposeTestRule that I found useful along the way. Some light documentation of strategies for resolving flaky tests have been added to the [Test Troubleshooting doc](https://www.notion.so/mbta-downtown-crossing/Test-Troubleshooting-551faaa2f7b14fbdafa9214f63e5f816?source=copy_link#20af5d8d11ea80f5b98cd3cc43c4d8b6), but I think the muscle memory of mocking View Models as we add new UI tests will be most helpful.

iOS

android

### Testing

What testing have you done?
* Ran each of these tests with the RetryableComposeTestRule for 100 iterations and confirmed they did not flake.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
@@ -2,10 +2,9 @@ package com.mbta.tid.mbta_app.android.map
 
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.runtime.remember
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsNotDisplayed
-import androidx.compose.ui.test.hasContentDescription
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithContentDescription
 import androidx.compose.ui.test.onNodeWithTag
@@ -20,7 +19,6 @@ import com.mbta.tid.mbta_app.android.SheetRoutes
 import com.mbta.tid.mbta_app.android.location.MockLocationDataManager
 import com.mbta.tid.mbta_app.android.location.ViewportProvider
 import com.mbta.tid.mbta_app.android.state.SearchResultsViewModel
-import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.repositories.MockConfigRepository
 import com.mbta.tid.mbta_app.repositories.MockSearchResultRepository
 import com.mbta.tid.mbta_app.repositories.MockSentryRepository
@@ -28,8 +26,18 @@ import com.mbta.tid.mbta_app.repositories.MockVisitHistoryRepository
 import com.mbta.tid.mbta_app.usecases.ConfigUseCase
 import com.mbta.tid.mbta_app.usecases.VisitHistoryUsecase
 import com.mbta.tid.mbta_app.utils.TestData
+import dev.mokkery.MockMode
+import dev.mokkery.answering.autofill.AutofillProvider
+import dev.mokkery.answering.calls
+import dev.mokkery.answering.returns
+import dev.mokkery.every
+import dev.mokkery.matcher.any
+import dev.mokkery.mock
+import kotlin.test.assertTrue
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.runBlocking
-import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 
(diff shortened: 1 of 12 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModelTest.kt
```diff
@@ -1154,32 +1154,30 @@ class StopDetailsViewModelTest {
...
+
         composeTestRule.waitForIdle()
         composeTestRule.waitUntil(2000) { newStopFilter == expectedFilter }
         kotlin.test.assertEquals(expectedFilter, newStopFilter)
     }
 
     @Test
-    fun testManagerAppliesTripFilterAutomaticallyOnDepartureChange() {
-
+    fun testManagerAppliesTripFilterAutomaticallyOnDepartureChange() = runBlocking {
```
KaylaBrady, line 1180: This test was failing on between 13 to 25 out of 100 test runs. I think it comes down to a race condition with this [LaunchedEffect](https://github.com/mbta/mobile_app/blob/b0dc0118e8e784121dbde1b2e129e3b967605909/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/StopDetailsViewModel.kt#L584) that uses `withContext(Dispatchers.Default)` -- that was sometimes finishing after the `LaunchedEffect` in this test to set new routeCardData. 

The simplest thing seemed to be to move setting the newRouteCardData out of a LaunchedEffect for this test, which immediately resulted in 100 / 100 successes, so I applied it to the similar tests in this file. 

Potentially we'd want to refactor things to avoid that race condition altogether, though I'm not sure that is worth doing at this point with the future ViewModel changes w/ Molecule.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
```diff
@@ -294,44 +302,58 @@ class HomeMapViewTests {
...
+
+        AutofillProvider.forMockMode.types.register(StateFlow::class) { MutableStateFlow(null) }
+
+        AutofillProvider.forMockMode.types.register(Flow::class) { MutableStateFlow(null) }
+
+        val mapVM = mock<IMapViewModel>(MockMode.autofill)
+        every { mapVM.selectedStop } returns MutableStateFlow(TestData.getStop("121"))
+        every { mapVM.configLoadAttempted } returns MutableStateFlow(true)
+        every { mapVM.showRecenterButton } returns MutableStateFlow(false)
+        every { mapVM.showTripCenterButton } returns MutableStateFlow(false)
```
KaylaBrady, line 330: This may have been sufficiently easy to do without mokkery, but I wanted to work with it as a tool that exists already in our test code but is maybe underutilized. Would be interested in feedback on whether we'd want to adopt mokkery more widely as a tool for mocking view models in UI tests, and the extent to which that is still relevant in a world where all VMs are using molecule. (I _think_ it is still relevant, but need to look closer)

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/MapAndSheetPageTest.kt
```diff
@@ -260,7 +263,7 @@ class MapAndSheetPageTest : KoinTest {
 
         composeTestRule.waitUntilExactlyOneExists(hasContentDescription("Mapbox Attribution"))
         composeTestRule.onNodeWithContentDescription("Mapbox Attribution").assertIsDisplayed()
-        composeTestRule.waitUntilDoesNotExist(hasContentDescription("Loading..."))
+        composeTestRule.waitUntilDoesNotExistDefaultTimeout(hasContentDescription("Loading..."))
```
KaylaBrady, line 266: The only thing that seemed to help for this test was increasing the timeout. I was playing with ways to make an a higher timeout more consistent across tests and was annoyed that you can't just override the existing timeout default for `waitUntilDoesNotExist`. Don't love this, could just pass around the constant DEFAULT_TIMEOUT in places instead.

boringcactus: If we wanted to, we could replace every `waitUntilDoesNotExist(` with a `waitUntilDoesNotExistDefaultTimeout(` in a custom Spotless formatter, vaguely like how we’re handling `Map.getValue` in shared code. That might be more trouble than it’s worth, though.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/MapAndSheetPageTest.kt
```diff
@@ -451,6 +455,34 @@ class MapAndSheetPageTest : KoinTest {
...
+                        CameraState(
+                            // Specifically setting zoom so that we don't fall back to using the
+                            // default
+                            // center location. If default center is used, then since `hasPermission
+                            // = true`,
+                            // A MapEffect will also call follow and throw off the count of calls to
+                            // follow.
+                            Point.fromLngLat(sampleStop.longitude, sampleStop.latitude),
+                            EdgeInsets(0.0, 0.0, 0.0, 0.0),
+                            /* zoom = */ ViewportProvider.Companion.Defaults.zoom,
```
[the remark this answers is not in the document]
KaylaBrady: This was just a copy / paste from mapbox code, will cleanup

bin/android-instrumented-test-ci.sh
```diff
@@ -27,7 +27,7 @@ function retry() {
   done
 }
 
-RETRIES=3 retry ./gradlew :androidApp:connectedStagingDebugAndroidTest --no-daemon
+./gradlew :androidApp:connectedStagingDebugAndroidTest --no-daemon
```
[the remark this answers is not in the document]
EmmaSimon: No that sounds good to me! It did seem a bit wasteful to retry everything rather than individual tests, I like this better, just wanted to be certain it was intentional.

PR #1067, opened 2025-06-11
feat(iOS): write enhanced favorites from filtered stop details

boringcactus:
### Summary

_Ticket:_ [ | Favorites | Store new favorites from stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210440278891496?focus=true)

This just about matches the current Android implementation; there are some pieces that live in different layers of the view hierarchy on iOS and Android, though, so it’s not 100% identical.

iOS

android

### Testing

Added a unit test for both the new behavior and the old behavior based on the feature flag. Manually verified that the behavior looks right.


iosApp/iosApp/ViewModels/StopDetailsViewModel.swift
```diff
@@ -65,10 +67,12 @@ class StopDetailsViewModel: ObservableObject {
     private let tripRepository: ITripRepository
     private let vehicleRepository: IVehicleRepository
 
-    private let togglePinnedUsecase = UsecaseDI().toggledPinnedRouteUsecase
+    private let favoritesUsecases: FavoritesUsecases
+    private let togglePinnedUsecase: TogglePinnedRouteUsecase
```
boringcactus, line 71: It turns out that if you allow overriding the `Repository` you need to correctly use that repository in the `UseCase` or else you will be completely miserable. I was printing the repository itself as a string (since that’s derived from its memory address) in like five different places trying to figure out where it became the wrong repository before I finally tracked this down.

BrandonTR (approved):

PR #1068, opened 2025-06-12
feat: Make search bar more closely match spec

EmmaSimon:
### Summary

_Ticket:_ Prep for [🤖 | Favorites | Add stops flow - search / filter](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210229155855234?focus=true)

We've had a search bar on Android that didn't match spec, because Android made it impossible to update the default styling on their built in `SearchBarDefaults.InputField`. This copies the code out of that and updates it to better match our styling. It's still not 100% there, but it's much closer, and this will fit more cleanly into the route picker filter page.

Before:
Screenshot_20250612_153442

After:
Screenshot_20250612_165507

Spec:
Screenshot 2025-06-12 at 4 52 09 PM

android

### Testing

Verified that the system string resources are read out correctly by TalkBack.


boringcactus (approved):
It’s cool that we can do stuff like this with the Android standard library components, but it’s bad that we need to.

PR #1069, opened 2025-06-16
feat: create Molecule FavoritesViewModel

boringcactus:
### Summary

_Ticket:_ [ | Favorites | Show favorites in tab](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210440278891499?focus=true)

I’m splitting out the shared changes from the iOS changes so that this PR isn’t absurdly large.

This creates shared versions of some of the state helpers from Android: some of them are structured differently, so I haven’t removed the Android versions yet, but they’ll hopefully be useful across several Molecule ViewModels as we move more of the logic to the shared code.

iOS

android

### Testing

Added unit tests for the new Molecule `FavoritesViewModel`. Checked manually that this will work with the iOS changes that still need their own tests and will be in a forthcoming PR.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/PredictionsByStopJoinResponse.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/PredictionsByStopJoinResponse.kt
@@ -82,3 +82,5 @@ data class PredictionsByStopJoinResponse(
         val empty = PredictionsByStopJoinResponse(emptyMap(), emptyMap(), emptyMap())
     }
 }
+
+fun PredictionsByStopJoinResponse?.orEmpty() = this ?: PredictionsByStopJoinResponse.empty
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
@@ -1,0 +1,186 @@
+package com.mbta.tid.mbta_app.viewModel
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import co.touchlab.skie.configuration.annotations.DefaultArgumentInterop
+import com.mbta.tid.mbta_app.model.RouteCardData
+import com.mbta.tid.mbta_app.model.RouteStopDirection
+import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
+import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.usecases.FavoritesUsecases
+import io.github.dellisd.spatialk.geojson.Position
+import kotlinx.coroutines.flow.Flow
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
+
+class FavoritesViewModel(private val favoritesUsecases: FavoritesUsecases) :
+    MoleculeViewModel<FavoritesViewModel.Event, FavoritesViewModel.State>() {
+    sealed interface Event {
+        data object ReloadFavorites : Event
+
+        data class SetActive(val active: Boolean, val isReturningFromBackground: Boolean?) : Event
+
+        data class SetAlerts(val alerts: AlertsStreamDataResponse?) : Event
+
+        data class SetLocation(val location: Position?) : Event
+
+        data class SetNow(val now: Instant) : Event
+    }
+
+    data class State(
+        val favorites: Set<RouteStopDirection>?,
+        val isReturningFromBackground: Boolean,
+        val routeCardData: List<RouteCardData>?,
+        val staticRouteCardData: List<RouteCardData>?,
+    ) {
+        constructor() : this(null, false, null, null)
+    }
+
+    @Composable
+    override fun runLogic(events: Flow<Event>): State {
+        var favorites: Set<RouteStopDirection>? by remember { mutableStateOf(null) }
+        var isReturningFromBackground: Boolean by remember { mutableStateOf(false) }
+        var routeCardData: List<RouteCardData>? by remember { mutableStateOf(null) }
+        var staticRouteCardData: List<RouteCardData>? by remember { mutableStateOf(null) }
+
+        var active: Boolean by remember { mutableStateOf(true) }
+        var alerts: AlertsStreamDataResponse? by remember { mutableStateOf(null) }
+        var location: Position? by remember { mutableStateOf(null) }
+        var now: Instant by remember { mutableStateOf(Clock.System.now()) }
+
+        val globalData = getGlobalData("FavoritesViewModel.getGlobalData")
+        val stopIds =
+            remember(favorites, globalData) {
+                val stops = favorites?.mapNotNull { globalData?.getStop(it.stop) }
+                stops?.flatMap { stop ->
+                    stop.childStopIds.filter { globalData?.stops?.containsKey(it) ?: false } +
+                        stop.id
+                }
+            }
+        val schedules = getSchedules(stopIds, "FavoritesViewModel.getSchedules")
+        val predictions = subscribeToPredictions(stopIds, active)
+
+        LaunchedEffect(Unit) { favorites = favoritesUsecases.getRouteStopDirectionFavorites() }
+
+        LaunchedEffect(Unit) {
+            events.collect { event ->
+                when (event) {
+                    Event.ReloadFavorites ->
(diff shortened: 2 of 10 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+                staticRouteCardData = emptyList()
+            } else {
+                val loadedRouteCardData =
+                    RouteCardData.routeCardsForStaticStopList(
+                        stopIds,
+                        globalData,
+                        RouteCardData.Context.Favorites,
+                    )
+                staticRouteCardData =
+                    filterRouteAndDirection(loadedRouteCardData, globalData, favorites)
```
boringcactus, line 135: Strictly speaking, the static route card data isn’t required until we implement favorites deletion, but that’s basically the next ticket, and it’ll be a little bit annoying to add in later instead (since the static data loads before the realtime data in the test).

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/fetchApi.kt
```diff
@@ -0,0 +1,34 @@
+package com.mbta.tid.mbta_app.viewModel
```
KaylaBrady, line 1: suggestion (non-blocking): this, getGlobalData, getScheduls, and subscribeToPredictions don't totally feel like they fit into the viewModel package to me. Maybe they could go into a separate package along the lines of `apiRequests`, `getData`, etc.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/response/PredictionsByStopJoinResponse.kt
```diff
@@ -82,3 +82,5 @@ data class PredictionsByStopJoinResponse(
         val empty = PredictionsByStopJoinResponse(emptyMap(), emptyMap(), emptyMap())
     }
 }
+
+fun PredictionsByStopJoinResponse?.orEmpty() = this ?: PredictionsByStopJoinResponse.empty
```
KaylaBrady, line 86: 👍 nice!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+import com.mbta.tid.mbta_app.model.RouteStopDirection
+import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
+import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.usecases.FavoritesUsecases
+import io.github.dellisd.spatialk.geojson.Position
+import kotlinx.coroutines.flow.Flow
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
+
+class FavoritesViewModel(private val favoritesUsecases: FavoritesUsecases) :
```
KaylaBrady, line 20: question (non-blocking): Thinking about the intersection of adopting Molecule and some of the learnings from https://github.com/mbta/mobile_app/pull/1064. What do you think of making an interface for FavoritesViewModel so that it is easier to mock the fireEvent wrappers functions for UI tests?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+                        globalData,
+                        location,
+                        schedules,
+                        predictions,
+                        alerts,
+                        now,
+                        emptySet(),
+                        RouteCardData.Context.Favorites,
+                    )
+                routeCardData = filterRouteAndDirection(loadedRouteCardData, globalData, favorites)
```
KaylaBrady, line 118: thought(non-blocking): This would be a further refactor so understandably not in scope for this PR, but wondering if here and for the staticRouteCardData it would be worth passing the favorites as a parameter so we don't have to filter after the fact.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
```diff
@@ -0,0 +1,315 @@
...
+                    now,
+                ),
+            )
+        fun FavoritesViewModel.State.tweakStaticDataTime() =
+            this.copy(
+                staticRouteCardData =
+                    this.staticRouteCardData?.map { routeCardData ->
+                        routeCardData.copy(at = staticDataTime)
+                    }
+            )
```
KaylaBrady, line 274: question: I'm having trouble parsing at a glance - why is this necessary?

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
```diff
@@ -0,0 +1,315 @@
...
+                FavoritesViewModel.State(
+                    favorites = favorites.routeStopDirection,
+                    isReturningFromBackground = false,
+                    routeCardData = expectedRealtimeData,
+                    staticRouteCardData = expectedStaticData,
+                ),
+                awaitItem().tweakStaticDataTime(),
+            )
+        }
+    }
```
KaylaBrady, line 314: suggestion: 
I think it would be worthwhile to add some basic tests for the fireEvent wrappers too.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+                    is Event.SetNow -> now = event.now
+                }
+            }
+        }
+
+        LaunchedEffect(predictions) {
+            if (predictions != null) {
+                isReturningFromBackground = false
+            }
+        }
```
KaylaBrady, line 89: question: why is this false? Maybe something about the naming of `active` and `isReturningFromBackground` could be clarified through modified naming or comments.

KaylaBrady (commented):
Love to see this split out and more moving into shared code! Have a few questions/comments as I get up to speed on Molecule.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/fetchApi.kt
boringcactus: Since they’re Composables, the only shared code that can use them is the ViewModels, but a subpackage would be reasonable.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
boringcactus: The static route cards are created with [the actual current time](https://github.com/mbta/mobile_app/blob/c72f28a4139515de528a7e56e0450b17c5d1f6f2/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt#L704), rather than with a parameter we can control.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
boringcactus: This is false because once the predictions have loaded, the favorites tab has finished returning from the background. On iOS, the `NearbyTransitView` sets `isReturningFromBackground = false` (which is bound to the `ErrorBannerViewModel`’s `loadingWhenPredictionsStale`) whenever a new predictions message is received, and on Android, the `PredictionsViewModel` directly sets the `ErrorBannerViewModel`’s `loadingWhenPredictionsStale` when a new `PredictionsByStopJoinResponse` is received.

Renamed `isReturningFromBackground` to `awaitingPredictionsAfterBackground` and tweaked the `SetActive` event.

KaylaBrady: That renaming is great, thank you!

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
KaylaBrady: Ah, got it. Any reason not to add now defaulting to the current time as a parameter instead? I think that would help generally for testability and for the readability of this particular test.

boringcactus: That’s cleaner, yeah.

PR #1078, opened 2025-06-17
fix(test): limit the parameter space of parametric tests

boringcactus:
### Summary

_Ticket:_ none

Noticed that our shared tests are slower than I expect them to be, and I think I found the issue.

iOS

android

### Testing

Checked that the tests all still pass locally.

### DIFF
--- a/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/parametric/ParametricTest.kt
+++ b/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/parametric/ParametricTest.kt
@@ -2,6 +2,7 @@ package com.mbta.tid.mbta_app.parametric
 
 import kotlin.enums.enumEntries
 import kotlin.test.assertEquals
+import kotlin.test.fail
 import kotlinx.coroutines.runBlocking
 
 /**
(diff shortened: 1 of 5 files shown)


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/parametric/ParametricTest.kt
```diff
@@ -52,6 +53,13 @@ class ParametricTest(val block: suspend ParametricTest.() -> Unit) {
     }
 
     internal suspend fun executeAll() {
+        val parameterSetCount = parameterSpace.map { it.size }.fold(1, Int::times)
+        if (parameterSetCount == 1) {
+            fail("parametricTest has no parameters, don’t use it")
+        }
+        if (parameterSetCount >= 10_000) {
+            fail("parametricTest has $parameterSetCount iterations, that’s too many")
+        }
```
boringcactus, line 62: I checked at a limit of 100, and there was one test with 4,000 and one with 200 but rewriting them didn’t make the test suite any faster locally, so I think 10,000 is a reasonable cutoff; that’s still more parameters than we should ever actually need, but it’d’ve caught the terrible case even before we added a fourth context (since at $3^{10}=59\thinspace049$ that was still over this limit).

PR #1098, opened 2025-06-26
feat(Android): First pass for route UI

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Favorites Route Details layout UI](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210485399858975?focus=true)

We really should stop assuming every UI ticket will be trivial. This only includes direction picker, route list, and loading state styling, it does _not_ include the GL route picker styling, header styling, or background route color fill, those will happen in a follow up PR.

I did some minor refactoring to better handle the stop list data in the UI, since the stop list response from the backend had no context on what route and direction it was loaded for, `RouteDetailsStopList.fromPieces` was happily consuming stale data and returning stop lists based on the data from the previously loaded opposite direction. This was especially noticeable on bus routes, where it would collapse every non-station stop on the route as "uncommon", since they weren't actually served by any pattern in that direction. Now the response keeps the context of the route and direction it was loaded for, so `RouteDetailsStopList.fromPieces` can check that it's using the correct stop list, and return null if not.

I also did some probably too cute randomization of the stop name and route pill lengths of the loading placeholder data. It makes it so that there's not jarring uniformity in the placeholder data, but also keeps that data consistent across loads by seeding the RNG with the route ID and direction ID combo.

android

### Testing

A lot of the UI changes aren't testable, but added a new unit test for the mismatched route/direction stop list loading, and existing tests pass.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListViewTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListViewTest.kt
@@ -75,7 +75,11 @@ class RouteStopListViewTest {
 
         val koin =
             testKoinApplication(objects) {
-                routeStops = MockRouteStopsRepository(listOf(stop1.id, stop2.id, stop3.id))
+                routeStops =
+                    MockRouteStopsRepository(
+                        listOf(stop1.id, stop2.id, stop3.id),
+                        routeId = mainRoute.id,
+                    )
             }
         val errorBannerVM = ErrorBannerViewModel(errorRepository = MockErrorBannerStateRepository())
 
@@ -86,6 +90,7 @@ class RouteStopListViewTest {
                     RouteDetailsContext.Details,
                     GlobalResponse(objects),
                     onClick = clicks::add,
+                    onBack = {},
                     onClose = { closed = true },
                     errorBannerViewModel = errorBannerVM,
                     rightSideContent = { context, _ ->
@@ -176,6 +181,7 @@ class RouteStopListViewTest {
                     RouteDetailsContext.Details,
                     GlobalResponse(objects),
                     onClick = {},
+                    onBack = {},
                     onClose = {},
                     errorBannerViewModel = errorBannerVM,
                     defaultSelectedRouteId = route2.id,
@@ -232,7 +238,8 @@ class RouteStopListViewTest {
             testKoinApplication(objects) {
                 routeStops =
                     MockRouteStopsRepository(
-                        listOf(stop1.id, stop2.id, stop3NonTypical.id, stop4NonTypical.id)
+                        listOf(stop1.id, stop2.id, stop3NonTypical.id, stop4NonTypical.id),
+                        routeId = mainRoute.id,
                     )
             }
         val errorBannerVM = ErrorBannerViewModel(errorRepository = MockErrorBannerStateRepository())
@@ -244,6 +251,7 @@ class RouteStopListViewTest {
                     RouteDetailsContext.Details,
                     GlobalResponse(objects),
                     onClick = {},
+                    onBack = {},
                     onClose = {},
                     errorBannerViewModel = errorBannerVM,
                     rightSideContent = { _, _ -> },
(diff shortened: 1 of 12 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListViewTest.kt
```diff
@@ -265,21 +273,26 @@ class RouteStopListViewTest {
...
 
         composeTestRule.setContent {
             KoinContext(koin.koin) {
                 RouteStopListView(
-                    RouteCardData.LineOrRoute.Route(objects.getRoute("Red")),
+                    route,
                     RouteDetailsContext.Favorites,
                     GlobalResponse(objects),
                     onClick = {},
+                    onBack = {},
```
KaylaBrady, line 295: suggestion (non-blocking): Maybe worth a quick test of calling onBack

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -522,7 +523,15 @@ fun MapAndSheetPage(
...
-                    onClose = { navController.popBackStackFrom<SheetRoutes.RouteDetails>() },
+                    onBack = { navController.popBackStackFrom<SheetRoutes.RouteDetails>() },
+                    onClose = {
+                        while (
+                            navController.currentRouteMatches<SheetRoutes.RoutePicker>() ||
+                                navController.currentRouteMatches<SheetRoutes.RouteDetails>()
+                        ) {
+                            navController.popBackStack()
+                        }
+                    },
```
KaylaBrady, line 534: question(non-blocking): Why not just clear the stack?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
```diff
@@ -74,11 +74,16 @@ data class RouteDetailsStopList(val segments: List<Segment>) {
...
-            routeStops: RouteStopsResponse?,
+            routeStops: RouteStopsResult?,
             globalData: GlobalResponse,
         ): RouteDetailsStopList? =
             withContext(Dispatchers.Default) {
-                if (routeStops == null) return@withContext null
+                if (
+                    routeStops == null ||
+                        routeStops.routeId != routeId ||
+                        routeStops.directionId != directionId
```
KaylaBrady, line 84: question(non-blocking): this seems reasonable to me, but I don't quite understand why an out of date result was being used (and am having trouble replicating myself). Is this a race condition when switching back and forth really quickly? If there is a mismatch, is there a possibility that this will indefinitely return null and the correct data will never be shown?

KaylaBrady (approved):
This looks good! Appreciate you breaking this chunk out

> We really should stop assuming every UI ticket will be trivial.

Agreed! Let's take this to our next refinement, and we can talk through whether any of our existing tickets are under pointed as part of the next sprint planning.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
EmmaSimon: Okay, I added a new `navigateToEntrypointFrom`, which uses the same approach of specifying the origin route to ensure that the nav can't be triggered again while animating to the new route. This surfaced a few issues with using reified types (you can't have a nested function which has a reified type), which got me to figure out a way to stop using reified types entirely for the `NavHostController` route extension helpers, which makes things way easier to work with overall.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
KaylaBrady: > In theory the LaunchedEffect should be cancelled every time the state changes and it should always reload from the endpoint for the current state, regardless of the current state of the request.

Ah yep good call, that makes me feel better!

PR #1099, opened 2025-06-27
fix(shared.FavoritesViewModelTest): Address flakiness

KaylaBrady:
### Summary

_Ticket:_ [favorites flaky molecule tests](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210621071746794?focus=true)

What is this PR for?

<details><summary>Pre-Review</summary>
This PR takes an approach for addressing flakiness in Molecule VM tests. There may be other approaches we want to consider instead, but this is is a first step towards solving the problem. 

This PR combines approach #2 and #4 as outlined [in slack here](https://mbta.slack.com/archives/C062QNAJZ2M/p1750874718166159)

> 2. Use [runCurrent() + expectMostRecent()](https://github.com/cashapp/turbine/issues/246#issuecomment-1674281572) instead of awaitItemSatisfying to get the last item that should match the desired state rather than the first (except I used advanceUntilIdle() instead of runCurrent())

> 4. Pass the dispatcher context from the test to routeCardsForStopList

@boringcactus had raised some concerns about how using `awaitItemSatisfying` could ignore inconsistent states that we wouldn't necessarily want to ignore. I think that remains true in this switch to `expectMostRecent`, but I like the explicitness that what is being tested is expected to be the last available item.

**Note:** this does go a bit against the philosophy of Turbine as outlined [in the readme](https://github.com/cashapp/turbine/?tab=readme-ov-file#asynchronicity-and-turbine)

> instead of using tools like runCurrent() to "push" an asynchronous flow along, Turbine's awaitItem(), awaitComplete(), and awaitError() "pull" them along by parking until a new event is ready.

Potentially we could continue to use `awaitItemSatisfying` here in combination with only the change to inject the test dispatcher. Still, I prefer the explicitness of `expectMostRecent`, but am open to other opinions on that! 
</details> 

Based on review, removing the use of `expectMostRecent` in favor of continuing to use `awaitItemSatisfying`, though in some cases with some more specific matchers. 

This keeps the change to use the test's coroutine context in `routeCardsForStopList` as recommended in the [Testing Coroutines doc ](https://developer.android.com/kotlin/coroutines/test#injecting-test-dispatchers).

iOS

android
- [X] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible (ideally in shared code)

### Testing

What testing have you done?
* Updated unit tests. Ran a few times locally without flakiness, but the real test will be seeing how this plays out over time in CI.
While testing, I periodically changed the injected dispatcher to Dispatchers.Default and confirmed that the tests would fail, which confirmed for me that doing the dispatcher injection was a worthwhile change to make.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -13,8 +13,8 @@ import kotlin.math.max
 import kotlin.time.Clock
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlin.time.Instant
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -1,5 +1,6 @@
 package com.mbta.tid.mbta_app.model
 
+import co.touchlab.skie.configuration.annotations.DefaultArgumentInterop
 import com.mbta.tid.mbta_app.map.style.Color
 import com.mbta.tid.mbta_app.model.UpcomingFormat.NoTripsFormat
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
@@ -643,6 +644,7 @@ data class RouteCardData(
          * [filterAtTime] and [filterAtTime] + [hideNonTypicalPatternsBeyondNext] are omitted.
          * Cancelled trips are also omitted when [context] = NearbyTransit.
          */
+        @DefaultArgumentInterop.Enabled
         suspend fun routeCardsForStopList(
             stopIds: List<String>,
             globalData: GlobalResponse?,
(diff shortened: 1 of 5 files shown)


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
```diff
@@ -478,17 +504,21 @@ class FavoritesViewModelTest : KoinTest {
...
 
         testViewModelFlow(viewModel).test {
+            println("before advance")
+            advanceUntilIdle()
+            println("after advance")
             assertEquals(
                 emptyList(),
-                awaitItemSatisfying { it.routeCardData != null }
+                expectMostRecentItem()
                     .routeCardData!!
```
KaylaBrady, line 522: This is still flaking 🤔 I'm not sure how `expectMostRecentItem` here could *not* have routeCardData set.

boringcactus: Maybe the `advanceUntilIdle` sets the stop IDs but a second `advanceUntilIdle` would be necessary to fetch the predictions?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -650,7 +651,32 @@ data class RouteCardData(
...
+            stopIds: List<String>,
+            globalData: GlobalResponse?,
+            sortByDistanceFrom: Position?,
+            schedules: ScheduleResponse?,
+            predictions: PredictionsStreamDataResponse?,
+            alerts: AlertsStreamDataResponse?,
+            now: Instant,
+            pinnedRoutes: Set<String>,
+            context: Context,
+            coroutineDispatcher: CoroutineDispatcher,
```
boringcactus, line 677: Why declare this as a separate function rather than just a new parameter with a default argument in the original function?

PR #1100, opened 2025-06-30
refactor: supporting work for map viewmodel

BrandonTR:
### Summary

_Ticket:_ [🤖 | Handle viewport state more reliably](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210454674414103?focus=true)

This is the supporting work for #1097 split into it's own PR. At this point all logic should pretty much be the same but some definitions were changed to be less reliant on MapBox and Compose objects. This also includes the change to move MapBox config management out of the view model.

### Testing

Adjusted existing tests to accommodate changes.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/utils/MapLayerManager.kt
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+    suspend fun addLayers(
+        routes: List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>,
+        state: StopLayerGenerator.State,
+        globalResponse: GlobalResponse,
+        colorPalette: ColorPalette,
+    )
+
+    fun resetPuckPosition()
```
BrandonTR, line 25: Going to add some documentation to these per other PR's feedback

PR #1111, opened 2025-07-02
refactor: create settings view model with Molecule

boringcactus:
### Summary

_Ticket:_ none

This is the code I threw out when I was working on my initial Molecule experimentation.

iOS

android

### Testing

What testing have you done?


boringcactus:
I only opened this PR in the first place to link to as an example of why this change wasn’t worth making.

PR #1113, opened 2025-07-02
refactor(ios.SaveFavoritesFlow): Switch popup library

KaylaBrady:
### Summary

_Ticket:_ Follow up to [ | Favorites | Stop details confirmation pop-up](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210440278891509?focus=true)

What is this PR for?

As discussed at dev-check-in on 7/1, switches to use https://github.com/Mijick/Popups since that will also support the upcoming toast notification work. 

iOS

android

### Testing

What testing have you done?
* Ran locally
* Confirmed with design that this is acceptable (just needed to decrease the horizontal modal padding from 50)
image


KaylaBrady:
issue: With VoiceOver on, it is possible to navigate beyond the contents of the modal. 

When using CustomAlert, it is typically not possible to do so, though when a vehicle departs and an interrupting VoiceOver announcement is sent, then you can navigate through all the sheet contents and lose the modal which isn't ideal.

KaylaBrady:
https://github.com/Mijick/Popups/issues/186

PR #1114, opened 2025-07-03
chore: use Android NDK 27 builds of Mapbox

boringcactus:
### Summary

_Ticket:_ none

I just got an Android Studio warning about [16 KB page size support](https://developer.android.com/guide/practices/page-sizes), which Google Play will start requiring in November and which Mapbox doesn’t provide by default. [Evidently](https://github.com/mapbox/mapbox-maps-android/issues/2448), Mapbox had been using the Android NDK 23, which doesn’t support 16 KB page sizes and would’ve been a breaking change to just upgrade from, so they’re publishing `-ndk27` versions of most of their packages, and if we move to those now then we should be ready for the requirement coming in November.

iOS

android

### Testing

Checked that the Android Studio APK analyzer says the Mapbox shared libraries are aligned to 16 KB and that the map still works.


KaylaBrady (approved):
TY for catching!

PR #1118, opened 2025-07-03
fix: align search loading/error/empty between iOS and Android

boringcactus:
### Summary

_Ticket:_ [🤖 Search | Implement empty/error search result states](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210561875928847?focus=true)

I went ahead and rearranged the Android component structure to be slightly more structurally similar to the iOS component structure while I was in there.

iOS

android

### Testing

Checked that empty state and error text displays correctly and is translated on both iOS and Android.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/results/StopResultsView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/results/StopResultsView.kt
@@ -12,6 +12,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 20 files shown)


iosApp/iosApp/Localizable.xcstrings
```diff
@@ -14478,7 +14572,7 @@
         },
         "fr" : {
           "stringUnit" : {
-            "state" : "translated",
+            "state" : "needs_review",
             "value" : "Envoyer des commentaires sur l’application"
```
boringcactus, line 14576: Xcode was telling me French was 93% translated but the other languages were all only 92%, and I knew that couldn’t be right. It looks like when this string was changed from `"Send app feedback"` to `"Send App Feedback"` in #895, the French translation was not reset to needs review like the rest of the translations were.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/search/results/StopResultsView.kt
```diff
@@ -29,8 +30,35 @@ import com.mbta.tid.mbta_app.android.R
...
+            val shape =
+                if (stops.size == 1) {
+                    RoundedCornerShape(10.dp)
+                } else if (index == 0) {
+                    RoundedCornerShape(topStart = 10.dp, topEnd = 10.dp)
+                } else if (index == stops.lastIndex) {
+                    RoundedCornerShape(bottomStart = 10.dp, bottomEnd = 10.dp)
+                } else {
+                    RoundedCornerShape(0.dp)
+                }
```
EmmaSimon, line 53: nitpick(non-blocking): This is probably overkill, but could we split this out into a helper function so it's not repeated here and in `RouteResultsView`? Maybe an extension on `List`, which takes the current index and returns the proper shape? Long term it might be nice to have a generic component that does all of this for lazy columns and applies halo borders to the partial shapes. I ran into the same issue for the route picker mode lists, the spec called for a rounded border and halo, but that was impossible with a LazyColumn so I needed to just use a regular Column.

boringcactus: We could even split the whole structure out into a shared component, and then later on we can adjust that component to use a LazyColumn and add a halo.

PR #1122, opened 2025-07-07
build(deps): upgrade to kotlinx.datetime 0.7.1 with stdlib Clock/Instant

boringcactus:
### Summary

_Ticket:_ none

Includes #1102 and #1106.

iOS

android

### Testing

Checked that the Android and iOS tests still pass.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,7 +30,8 @@ junit = "4.13.2"
 koin-bom = "4.0.4"
 kotlinpoet = "2.2.0"
 kotlinxCoroutinesCore = "1.10.2"
-kotlinxDatetime = "0.6.2"
+kotlinxDatetime = "0.7.1"
+kotlinxSerialization = "1.9.0"
 ktor = "3.1.3"
 # changelog: https://github.com/mapbox/mapbox-maps-android/blob/main/CHANGELOG.md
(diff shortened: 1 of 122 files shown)


gradle/libs.versions.toml
```diff
@@ -82,6 +83,7 @@ kotlinpoet = { module = "com.squareup:kotlinpoet", version.ref = "kotlinpoet" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinxCoroutinesCore" }
 kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "kotlinxCoroutinesCore" }
 kotlinx-datetime = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version.ref = "kotlinxDatetime" }
+kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinxSerialization" }
```
boringcactus, line 86: This had apparently not been a direct dependency until now; I guess it was being dragged in by `ktor-serialization-kotlinx-json`. It needed to be upgraded beyond whatever version we had before, since the serializers for the new stdlib `Instant` can’t live in the stdlib itself.

PR #1126, opened 2025-07-09
fix(shared.MapViewModel): Show only selected route shapes in stop details

KaylaBrady:
### Summary

_Ticket:_ [🤖 Show only selected route shapes in stop details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210621071746803?focus=true)

What is this PR for?

This PR reworks some of the logic of the MapViewModel to ensure that the appropriate map sources & layers are reloaded when the state changes. Will add some specific comments inline about the approach.

iOS

android

### Testing

What testing have you done?
* Ran locally, confirmed the correct stop routes load & are cleared when navigating back to nearby
* Wrote a few unit tests. Having some challenges with async behavior causing one test to flake - would appreciate any feedback on how to improve this!

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
@@ -1,5 +1,6 @@
 package com.mbta.tid.mbta_app.map
 
+import co.touchlab.skie.configuration.annotations.DefaultArgumentInterop
 import com.mbta.tid.mbta_app.map.style.Feature
 import com.mbta.tid.mbta_app.map.style.FeatureCollection
 import com.mbta.tid.mbta_app.map.style.FeatureProperty
@@ -24,6 +25,7 @@ import com.mbta.tid.mbta_app.utils.resolveParentId
 import io.github.dellisd.spatialk.geojson.LineString
 import io.github.dellisd.spatialk.turf.ExperimentalTurfApi
 import io.github.dellisd.spatialk.turf.lineSlice
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
@@ -48,18 +50,28 @@ object RouteFeaturesBuilder {
 
     val propAlertStateKey = FeatureProperty<String>("alertState")
 
+    @DefaultArgumentInterop.Enabled
     suspend fun generateRouteSources(
         routeData: List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>,
         globalData: GlobalResponse,
         globalMapData: GlobalMapData?,
-    ) = generateRouteSources(routeData, globalData.stops, globalMapData?.alertsByStop.orEmpty())
+        coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default,
+    ) =
+        generateRouteSources(
+            routeData,
+            globalData.stops,
+            globalMapData?.alertsByStop.orEmpty(),
+            coroutineDispatcher,
+        )
 
+    @DefaultArgumentInterop.Enabled
     suspend fun generateRouteSources(
         routeData: List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>,
         stopsById: Map<String, Stop>,
         alertsByStop: Map<String, AlertAssociatedStop>,
+        coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default,
     ): List<RouteSourceData> =
-        withContext(Dispatchers.Default) {
+        withContext(coroutineDispatcher) {
             routeData.map {
                 generateRouteSource(
                     routeId = it.routeId,
@@ -150,6 +162,9 @@ object RouteFeaturesBuilder {
         val fullLineString = LineString(coordinates)
         val alertAwareSegments =
             routePatternShape.routeSegments.flatMap { segment ->
+                if (segment.sourceRouteId == "Green-B") {
+                    println("Green-B segment found")
+                }
                 segment.splitAlertingSegments(alertsByStop = alertsByStop ?: emptyMap())
             }
         return alertAwareSegments.mapNotNull { segment ->
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
@@ -162,9 +162,6 @@ object RouteFeaturesBuilder {
         val fullLineString = LineString(coordinates)
         val alertAwareSegments =
             routePatternShape.routeSegments.flatMap { segment ->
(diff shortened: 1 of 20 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/map/RouteFeaturesBuilder.kt
```diff
@@ -150,6 +162,9 @@ object RouteFeaturesBuilder {
         val fullLineString = LineString(coordinates)
         val alertAwareSegments =
             routePatternShape.routeSegments.flatMap { segment ->
+                if (segment.sourceRouteId == "Green-B") {
+                    println("Green-B segment found")
```
KaylaBrady, line 166: TODO: Remove a bunch of logging

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/RailRouteShapeRepository.kt
```diff
@@ -44,6 +48,50 @@ constructor(
     val response: MapFriendlyRouteResponse = MapFriendlyRouteResponse(emptyList()),
     val onGet: () -> Unit = {},
 ) : IRailRouteShapeRepository {
+
+    constructor(
+        objects: ObjectCollectionBuilder
+    ) : this(
```
KaylaBrady, line 54: Needed to mock this in order to test that the route data was changing as expected

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
```diff
@@ -122,47 +124,81 @@ class MapViewModel(
...
+        var allStopSourceData by remember { mutableStateOf<FeatureCollection?>(null) }
+
+        // The current route source & stop layer state to display
+        var routeSourceData by remember { mutableStateOf<List<RouteSourceData>?>(null) }
+        var routeShapes by remember {
+            mutableStateOf<List<MapFriendlyRouteResponse.RouteWithSegmentedShapes>?>(null)
+        }
+        var stopLayerGeneratorState by remember {
+            mutableStateOf(StopLayerGenerator.State(null, null))
+        }
```
KaylaBrady, line 140: Biggest / key change: Introducing these additional state variables to track set what should be displayed on the map based on the state. 

I opted for this approach since the root of the bug was that we were making calls to the LayerManager in different functions (like handleNavChange), but not everywhere that it needed to. Setting this state in the appropriate places, then calling LayerManager when they changed seemed to me like it will be easier to manage as we add new states. 

But potentially it is making writing tests harder...or introducing real race conditions that I haven't contended with appropriately.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModelTests.kt
```diff
@@ -139,4 +168,154 @@ class MapViewModelTests : KoinTest {
...
+            viewModel.layerManagerInitialized(layerManger)
+            viewModel.densityChanged(1f)
+            assertEquals(MapViewModel.State.Overview, awaitItem())
+            advanceUntilIdle()
+            viewModel.navChanged(SheetRoutes.StopDetails(chestnutHill.id, null, null))
+            advanceUntilIdle()
+
+            assertEquals(MapViewModel.State.StopSelected(chestnutHill, null), awaitItem())
+            advanceUntilIdle()
+            viewModel.navChanged(SheetRoutes.NearbyTransit)
```
KaylaBrady, line 304: the effects of `navChanged(SheetRoutes.NearbyTransit)` happen before the effects of `navChanged(SheetRoutes.StopDetails` are run, even with the calls to advanceUntilIdle() here. If I remove the final call to `navChanged(SheetRoutes.NearbyTransit)`, then the check for `layerManger.updateRouteSourceData(matching { it.size == 1 })` works as expected.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModelTests.kt
```diff
@@ -139,4 +169,154 @@ class MapViewModelTests : KoinTest {
...
+            assertEquals(MapViewModel.State.Overview, awaitItem())
+            advanceUntilIdle()
+            viewModel.navChanged(SheetRoutes.StopDetails(chestnutHill.id, null, null))
+            assertEquals(MapViewModel.State.StopSelected(chestnutHill, null), awaitItem())
+            advanceUntilIdle()
+        }
+
+        advanceUntilIdle()
+        viewModel.navChanged(SheetRoutes.NearbyTransit)
+        advanceUntilIdle()
```
KaylaBrady, line 308: This test passes when run on its own if this `viewModel.navChange` call is moved outside the `testViewModelFlow`, but still fails when run in a batch. Not clear to me why calling this within `testViewModelFlow` would *never* pass - there maybe some fundamental of molecule or async testing that I'm missing here.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModelTests.kt
```diff
@@ -139,4 +169,154 @@ class MapViewModelTests : KoinTest {
             assertEquals(MapViewModel.State.StopSelected(stop, null), awaitItem())
         }
     }
+
+    @Test
+    fun whenInStopDetailsNotResetToAllRailOnAlertChange() = runTest {
```
KaylaBrady, line 174: Ugh, these other tests are flaking in CI. Will try to debug a little, but may just ignore them too 😞

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
EmmaSimon: I think this makes more sense, I like the move towards having all the relevant map data and calls in a single place. Testability of this is probably going to be a pain no matter what.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModelTests.kt
KaylaBrady: Made an attempt at that in https://github.com/mbta/mobile_app/pull/1126/commits/31ef166246a3d2851dc46c09df53f7c7bbdaf39b, I think the tricky part is trying to reset the mock midway in order to verify the effect of only the single call we want to test for each. I opted to ignore them all as of that commit but left in the re-write. 

LMK if that isn't quite what you were envisioning! @boringcactus - also would be interested if you have any thoughts on how to make these molecule tests more stable.

KaylaBrady: Oh maybe because what I care about here is more the side effects from LaunchedEffects I can test using composeTestRule instead and `waitUntil`... I'll give that a very small timebox and otherwise just merge as-is with the tests ignored.

PR #1129, opened 2025-07-11
feat(ios.FavoritesPage): Edit favorites entrypoint

KaylaBrady:
### Summary

_Ticket:_ [ | Favorites | UX | Delete favorites](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210440278891503?focus=true)

What is this PR for?

This PR is the first chunk of adding navigation into the edit favorites flow. I opted to do a little refactoring to add the action buttons into the existing SheetHeader component and adopt that on android as well, so breaking this off here for review.

iOS

android

### Testing

What testing have you done?
* Ran locally on iOS & android
* Added some simple unit tests to build on in next PR
IMG_0331
IMG_0332

### DIFF
--- a/iosApp/iosApp/Localizable.xcstrings
+++ b/iosApp/iosApp/Localizable.xcstrings
@@ -7312,6 +7312,9 @@
         }
       }
     },
+    "Done" : {
+      "comment" : "Button text for closing flow"
+    },
     "Drag handle" : {
       "comment" : "Screen reader text for the sheet drag handle to move the textual content over the map",
       "extractionState" : "manual",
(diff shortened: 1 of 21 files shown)


iosApp/iosApp/Localizable.xcstrings
```diff
@@ -7312,6 +7312,9 @@
         }
       }
     },
+    "Done" : {
```
EmmaSimon, line 7315: I assume the translations will be added in the follow up PR? Kinda surprised they aren't already here from the Android implementation.

KaylaBrady: Ooh good catch, also surprised they weren't already here and was fully assuming they would be. Will add. Also surprised the CI didn't catch this, I thought there was a new check for missing iOS translations

PR #1130, opened 2025-07-14
feat(ios.EditFavoritesPage):  Delete favorites functionality

KaylaBrady:
### Summary

_Ticket:_ [ | Favorites | UX | Delete favorites](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210440278891503?focus=true)

What is this PR for?
* Adds delete favorites flow to iOS
* Updates android implementation to match based on [discussion with design](https://www.figma.com/design/eFiLM8xjo9hKprg93BBCtZ?node-id=16509-139151#1338423338)
    * Only trash icon is tappable
    * tapping trash immediately removes favorite
    * Note: Android favorites uses a different VM - I did the minimal amount of work to get this functionality in the old VM knowing that it will be going away soon. 

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran on both iOS & android locally, confirmed I could remove favorites as expected
IMG_0335


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -157,6 +165,7 @@ class FavoritesViewModel(
                         RouteCardData.Context.Favorites,
                         // not depending on now because it only matters for testing
                         now,
+                        location,
```
KaylaBrady, line 168: Since setLocation is only called from the favorites page (and not while edit favorites page is open), I think it is safe to pass the latest value of location here so that the sort order on the edit page matches the sort order from nearby.

PR #1142, opened 2025-07-15
feat: shared logic for route details branching

boringcactus:
### Summary

_Ticket:_ [🤖 | Favorites | Route Details Branching implementation](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210764630199366?focus=true)

Without the TestData changes, this PR is +1341 -8, which is still unreasonably large, sorry.

iOS

android

### Testing

Checked that every single diagram drawn by the visualizer looks defensible. Added unit tests.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranching.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranching.kt
@@ -1,0 +1,691 @@
+package com.mbta.tid.mbta_app.model.routeDetailsPage
+
+import com.mbta.tid.mbta_app.model.Route
+import com.mbta.tid.mbta_app.model.RoutePattern
+import com.mbta.tid.mbta_app.model.RoutePattern.Typicality
+import com.mbta.tid.mbta_app.model.RouteType
+import com.mbta.tid.mbta_app.model.Stop
+import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.model.routeDetailsPage.RouteBranching.DisambiguatedStopId
+import com.mbta.tid.mbta_app.repositories.RouteStopsResult
+import kotlin.collections.component1
+import kotlin.collections.component2
+import kotlin.collections.iterator
+
+private object Workarounds {
+    fun rewriteRouteStopsResult(routeStopsResult: RouteStopsResult): RouteStopsResult {
+        return when {
+            routeStopsResult.routeId == "64" && routeStopsResult.directionId == 0 -> {
+                // as of 2025-07-11, the 64 outbound has an A->C->E B->C->E D->E, which would be
+                // manageable if the RouteStopsResult said ABCDE, but it says DABCE, causing three
+                // parallel segments
+                val a = listOf("730", "2755")
+                val b = listOf("1060", "72")
+                val c = listOf("1123")
+                val d = listOf("2231", "12232", "24486", "24487", "24488", "24489", "2442", "2443")
+                val e = listOf("2444")
+                val bad = d + a + b + c + e
+                val good = a + b + c + d + e
+                if (routeStopsResult.stopIds.take(bad.size) == d + a + b + c + e) {
+                    routeStopsResult.copy(stopIds = good + routeStopsResult.stopIds.drop(good.size))
+                } else {
+                    routeStopsResult
+                }
+            }
+            else -> routeStopsResult
+        }
+    }
+
+    fun rewriteStopGraph(
+        routeStopsResult: RouteStopsResult,
+        stopGraph:
+            RouteBranching.MutableGraph<DisambiguatedStopId, Pair<Stop, MutableSet<Typicality>>>,
+    ) {
+        when (routeStopsResult.routeId to routeStopsResult.directionId) {
+            Pair("33", 0) -> {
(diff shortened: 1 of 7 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranching.kt
```diff
@@ -0,0 +1,691 @@
...
+        operator fun unaryMinus() =
+            when (this) {
+                Left -> Right
+                Right -> Left
+            }
+    }
+
+    data class StickSideState(
+        val before: Boolean,
+        val currentStop: Boolean,
```
KaylaBrady, line 202: suggestion(non-blocking): It took me until I read where this is set in code to understand `currentStop` - some docs here might help.

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/ProjectUtils.kt
```diff
@@ -62,7 +63,7 @@ object ProjectUtils {
     }
 
     fun fetchTestData(): Unit = runBlocking {
-        startKoin { modules(appModule(AppVariant.Prod), platformModule) }
+        startKoin { modules(appModule(AppVariant.Staging), platformModule) }
```
KaylaBrady, line 66: question(non-blocking): Why not continue to use Prod?

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranchingTest.kt
```diff
@@ -0,0 +1,354 @@
...
+                    listOf(
+                        BranchStop(pawtucket, StickState(left = forward, right = empty)),
+                        BranchStop(providence, StickState(left = forward, right = empty)),
+                        BranchStop(tfGreenAirport, StickState(left = forward, right = empty)),
+                        BranchStop(
+                            wickfordJunction,
+                            StickState(left = forward.copy(after = false), right = empty),
+                        ),
+                    ),
+                    name = "Providence",
```
KaylaBrady, line 324: suggestion: I think it would be worth some separate unit tests for the branch naming candidates to reflect each of the possible cases

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/RouteBranchingVisualizer.kt
```diff
@@ -0,0 +1,283 @@
...
+import kotlin.io.path.extension
+import kotlin.io.path.nameWithoutExtension
+import kotlin.io.path.writeLines
+import kotlin.io.path.writeText
+import kotlin.system.exitProcess
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.measureTimedValue
+import org.koin.core.context.startKoin
+
+object RouteBranchingVisualizer {
```
KaylaBrady, line 27: suggestion(non-blocking): Maybe worth a quick doc.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranching.kt
```diff
@@ -0,0 +1,691 @@
...
+                    )
+                else null
+            return RouteBranching(stopGraph, segmentGraph, segments)
+        }
+
+        private fun buildStopGraph(
+            routeStopsResult: RouteStopsResult,
+            patterns: List<RoutePattern>,
+            globalData: GlobalResponse,
+        ): Graph<DisambiguatedStopId, Pair<Stop, Set<Typicality>>> {
```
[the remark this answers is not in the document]
boringcactus: This feels like something that should be done in the frontend, since all of the data is already present, but the topological sort of the list of stops also felt like something that should be done in the frontend, and we put it in the backend instead so we could benefit from the V3 API’s mountain of special cases without having to redefine them all ourselves. Given that route details already requires a response from the backend, it seems like it would make sense to move all of this logic into the backend, so that we can adjust special cases without a new release and eventually consider caching or eagerly computing these graphs for all users.

shared/src/jvmMain/kotlin/com/mbta/tid/mbta_app/ProjectUtils.kt
boringcactus: Prod doesn’t have https://github.com/mbta/mobile_app_backend/pull/323 so the global data can’t load.

KaylaBrady: Ah! Good reminder that we need to kick off a backend deploy anyway before the next release, I can kick that off.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/routeDetailsPage/RouteBranching.kt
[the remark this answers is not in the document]
KaylaBrady: If it ends up being a lot of work to shift, we can consider that for a future enhancement / discuss prioritization with product. I do think it could be nice to have long-term, but don't see it as a hard blocker for releasing favorites.

boringcactus:
Subsumed by https://github.com/mbta/mobile_app_backend/pull/341.

PR #1145, opened 2025-07-15
fix(shared.MapViewModel): Center on stop when tapping from search result

KaylaBrady:
### Summary

_Ticket:_ [🤖 | v1.2.6 QA | Tapping on a search result doesn’t pan map](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210793281349695?focus=true)

What is this PR for?
Previously, `viewportManager.stopCenter` was only called for the `SelectedStop` event. Now it is also called during the `NavChanged` event. Based on [feedback in slack](https://mbta.slack.com/archives/C062QNAJZ2M/p1752594812853529), I opted to do this with a helper function called within event handling rather than introduce a new `LaunchedEffect`. 

This also updates other minor inconsistencies, like the trip center button missing when navigating from search.

iOS

android

### Testing

What testing have you done?
* Ran locally, tapping on various map elements to confirm centering behaved as expected. When tapping on a row for a far away vehicle, the map stays centered on the stop, even though clicking on that tile within StopDetails does center on that vehicle (video below - SL example). On prod today it doesn't center immediately on the selected vehicle, but it also doesn't center when navigating to that trip via a tile. I didn't try too hard to change that behavior

https://github.com/user-attachments/assets/97d29177-6ae7-48d9-9e88-190328ed2ea8

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
@@ -204,9 +204,6 @@ fun HomeMapView(
                 MapEffect(true) { map ->
                     map.mapboxMap.addOnMapClickListener { point ->
                         map.getStopIdAt(point) {
-                            globalData?.getStop(it)?.let { stop ->
-                                viewModel.selectedStop(stop, null)
-                            }
                             handleStopNavigation(it)
                             analytics.tappedOnStop(it)
                         }
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
@@ -64,8 +64,6 @@ import com.mbta.tid.mbta_app.dependencyInjection.RepositoryDI
 import com.mbta.tid.mbta_app.map.StopLayerGenerator
 import com.mbta.tid.mbta_app.model.RouteCardData
 import com.mbta.tid.mbta_app.model.SheetRoutes
-import com.mbta.tid.mbta_app.model.Stop
-import com.mbta.tid.mbta_app.model.StopDetailsFilter
 import com.mbta.tid.mbta_app.model.Vehicle
 import com.mbta.tid.mbta_app.repositories.IGlobalRepository
 import com.mbta.tid.mbta_app.viewModel.IMapViewModel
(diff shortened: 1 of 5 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -113,25 +110,11 @@ fun HomeMapView(
...
-                val state = (state as State.VehicleSelected)
-                state.stop to null
-            }
-        }
-
     val showTripRecenterButton =
         when (state) {
             is State.StopSelected,
             is State.Overview -> false
-            is State.VehicleSelected -> !viewportProvider.isVehicleOverview && !isSearchExpanded
```
KaylaBrady, line 134: I think this was resolving an old issue related to the trip button appearing over the search results, but that is no longer a problem (and there isn't equivalent logic for the puck recenter button)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
```diff
@@ -316,7 +296,11 @@ fun HomeMapView(
                             onClick =
                                 if (vehicle.tripId != null) {
                                     {
-                                        viewModel.selectedVehicle(vehicle, stop, stopFilter)
+                                        viewModel.selectedVehicle(
```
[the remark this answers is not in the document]
EmmaSimon: I'm seeing the same thing when tapping departure tiles, I'm taking a closer look

androidApp/src/main/java/com/mbta/tid/mbta_app/android/map/HomeMapView.kt
KaylaBrady: Oh good call, I'll test that out. It would be easy enough to add to the regular recenter button if that is a problem, I think we just need a check equivalent to `isSearchExpanded && isSearchVisible`

KaylaBrady: Tested it out, and search results + center on a vehicle can't co-exist, so I think this is safe to remove.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
```diff
@@ -361,7 +355,7 @@ class MapViewModel(
         val newState =
             if (routePickerOrDetails && currentState is State.VehicleSelected) {
                 if (currentState.stop != null) {
-                    State.StopSelected(currentState.stop, currentState.stopFilter)
+                    State.StopSelected(currentState.stop!!, currentState.stopFilter)
```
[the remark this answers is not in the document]
KaylaBrady: Ah lost track of it, thanks for the reminder! In now.

EmmaSimon (approved):

PR #1152, opened 2025-07-17
fix(Android): Recenter map when location changes

EmmaSimon:
### Summary

_Ticket:_ No ticket

The map was not updating the viewport when location updates were sent. This was happening because the `following` boolean was set to a stale value in the `locationDataManager.currentLocation.collect` callback, and was never being updated when the following state changed. There was also an interconnected issue where `following` was always set to false when the VM state was `TripSelected`, which shouldn't be the case because you can be following current location when a trip is selected, and it was causing the recenter button to never disappear on stop details, even when following. Changing every use of following back to `viewportProvider.isFollowingPuck` solved both issues.


KaylaBrady (commented):
Appreciate the consolidation of following state! Walked down the block and confirmed this recenters as expected.

KaylaBrady (approved):

PR #1156, opened 2025-07-21
fix: misc. route-specific edge case favorites QA 

KaylaBrady:
### Summary

_Ticket:_ [🤖 | Favorites | Fix: route-specific findings](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210755319204273?focus=true)

What is this PR for?
This PR addresses 3 small QA findings for edge case routes
1. Disclaimer when trying to add a stop with no service
2. Android only - trash icon missing in edit favorites when long headsign
3. All stops collapsed as less common when no typical pattern

iOS

android
- [X] All user-facing strings added to strings resource in alphabetical order

### Testing

What testing have you done?

| Case | Android | iOS |
|--------|--------|--------|
| 1. Drop-of only (also unit tested) Forced case for iOS at park st since add favorites not implemented yet for SL | image | IMG_0337 |
| 2. Android trash icon | image| N/A |
| 3.  no typical patterns (also unit tested) | image  | IMG_0338 |

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListView.kt
@@ -364,8 +364,10 @@ private fun RouteStops(
         horizontalAlignment = Alignment.Start,
         scrollEnabled = !loading,
     ) {
+        val hasTypicalSegment = stopList.segments.any { it.isTypical }
+
         stopList.segments.forEachIndexed { segmentIndex, segment ->
-            if (segment.isTypical) {
+            if (segment.isTypical || !hasTypicalSegment) {
                 segment.stops.forEachIndexed { stopIndex, stop ->
                     val stopPlacement =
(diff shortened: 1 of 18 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/routeDetails/RouteStopListView.kt
```diff
@@ -364,8 +364,10 @@ private fun RouteStops(
         horizontalAlignment = Alignment.Start,
         scrollEnabled = !loading,
     ) {
+        val hasTypicalSegment = stopList.segments.any { it.isTypical }
```
KaylaBrady, line 367: As [discussed in slack](https://mbta.slack.com/archives/C062QNAJZ2M/p1752863038564189?thread_ts=1752848976.319869&cid=C062QNAJZ2M), this is a simplified approach to solving this problem which requires the assumption that when there is no pattern of typicality 1, then all patterns are of the same typicality.

EmmaSimon (approved):

PR #1164, opened 2025-07-23
feat(Analytics): Add tracking for favorites

KaylaBrady:
### Summary

_Ticket:_ [Favorites | Analytics](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210660818493260?focus=true)

What is this PR for?

This PR adds analytics tracking in iOS & android related to favorites
* \# of favorites on startup
* When favorites added & removed
* When user taps departure
* When page changes

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Deploying to dev orange to trigger some events, will check with product that all looks as expected


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
```diff
@@ -161,6 +195,7 @@ abstract class Analytics {
     }
 
     fun track(screen: AnalyticsScreen) {
+        lastTrackedScreen = screen
```
KaylaBrady, line 198: This seemed like the most straightforward way to avoid having to drill the context through to the lower-level route card component where the tracking actually happens.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -107,7 +118,11 @@ class FavoritesViewModel(
                 onAnyMessageReceived = { awaitingPredictionsAfterBackground = false },
             )
 
-        LaunchedEffect(Unit) { favorites = favoritesUsecases.getRouteStopDirectionFavorites() }
+        LaunchedEffect(Unit) {
+            val fetchedFavorites = favoritesUsecases.getRouteStopDirectionFavorites()
+            favorites = fetchedFavorites
+            analytics.recordSession(fetchedFavorites.count())
```
EmmaSimon, line 124: question(non-blocking): Did you consider putting this with all the other `recordSession` calls in `ContentView.kt`? I guess it's a choice between pulling the `favoritesUsecase` in there, or the `analytics` object here, and both are slightly awkward.

EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
EmmaSimon: Ah, right, that sounds good then. I'm not particularly attached to colocating them, and that's a good enough reason to leave it here.

PR #1165, opened 2025-07-23
fix(Android): MapVM RouteCardData buffer overflow

EmmaSimon:
### Summary

_Ticket:_ [Event buffer overflow MapViewModel](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210857402669010?focus=true)

Use a separate flow for route card updates in the MapViewModel. It's a `StateFlow` rather than a `SharedFlow` because we only ever care about the latest value and have no reason to buffer a queue of updates.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
@@ -37,7 +37,9 @@ import kotlin.time.Duration.Companion.seconds
 import kotlin.time.Instant
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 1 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
```diff
@@ -193,13 +194,18 @@ class MapViewModel(
             }
         }
 
+        LaunchedEffect(Unit) {
+            // Route card data is sent through a separate StateFlow rather than the event flow
+            // because frequent updates were causing buffer overflows.
+            routeCardDataUpdates.collectLatest { routeCardData = it }
```
boringcactus, line 200: It might even work to `val routeCardData = routeCardDataUpdates.collectAsState()` and skip this `LaunchedEffect` entirely.

boringcactus (approved):

PR #1170, opened 2025-07-24
feat: shared code for final route branching approach

boringcactus:
### Summary

_Ticket:_ [🤖 | Favorites | Route Details Branching implementation](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210764630199366?focus=true)

Consumes https://github.com/mbta/mobile_app_backend/pull/341 / https://github.com/mbta/mobile_app_backend/pull/342 in a parallel code path that nothing calls yet. I’ve already done the Android work, so I know this is the set of functionality that’s needed for the Android work, but I rearranged a couple of things as part of splitting out this PR separately, so it may be another bit before the Android stacked PR appears.

Also, I know I said this wasn’t going to be a 2,000 line PR, but after splitting out the shared work I added back in the old path (and adding is more lines than replacing) and I wrote tests for `NewSegment.twistedConnections`, plus the branching segment list is verbose enough that the repo test with the hardcoded JSON literal is larger than any other file diff in this PR.

iOS

android

### Testing

Added unit tests. Checked that all tests pass.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
@@ -1,18 +1,74 @@
 package com.mbta.tid.mbta_app.model
 
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
-import com.mbta.tid.mbta_app.repositories.RouteStopsResult
+import com.mbta.tid.mbta_app.repositories.NewRouteStopsResult
+import com.mbta.tid.mbta_app.repositories.OldRouteStopsResult
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
-data class RouteDetailsStopList(val directionId: Int, val segments: List<Segment>) {
+data class RouteDetailsStopList(
+    val directionId: Int,
+    val oldSegments: List<OldSegment>?,
+    val newSegments: List<NewSegment>?,
+) {
 
     /** A subset of consecutive stops that are all typical or all non-typical. */
-    data class Segment(val stops: List<Entry>, val hasRouteLine: Boolean) {
+    data class OldSegment(val stops: List<OldEntry>, val hasRouteLine: Boolean) {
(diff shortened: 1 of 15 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
```diff
@@ -1,18 +1,74 @@
...
-data class RouteDetailsStopList(val directionId: Int, val segments: List<Segment>) {
+data class RouteDetailsStopList(
+    val directionId: Int,
+    val oldSegments: List<OldSegment>?,
+    val newSegments: List<NewSegment>?,
+) {
 
     /** A subset of consecutive stops that are all typical or all non-typical. */
-    data class Segment(val stops: List<Entry>, val hasRouteLine: Boolean) {
+    data class OldSegment(val stops: List<OldEntry>, val hasRouteLine: Boolean) {
```
EmmaSimon, line 16: question/nitpick: For this rename, are we planning to remove `OldSegment` etc and rename everything again once we implement this on iOS, or will the old endpoint and data structures stick around to fall back to if the branching calculation fails? 

It seems ideal for the fallback to happen on the backend, and for it to return the disconnected stops in the new format. But if we do the fallback on the frontend by making a separate request to the old endpoint, and plan to keep both old and new structures around indefinitely, then naming them "old" and "new" feels wrong.

EmmaSimon (approved):
This was much less intimidating than I expected based on line count! I just have one clarifying question.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteDetailsStopList.kt
boringcactus: I was planning to remove `OldSegment` and rename `NewSegment` back to `Segment`, yeah. (Originally, I had just replaced the old `Segment` with the new `Segment`, but since we aren’t doing shared and Android and iOS all in the same PR, that wasn’t going to cut it.)

PR #1179, opened 2025-07-28
fix: Order nearby stops after filtering

EmmaSimon:
### Summary

_Ticket:_ [Investigate: nearby stop sort order](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210845316775461?focus=true)

The list of stop IDs going into `filterStopsWithRedundantPatterns` was sorted by distance, but the list coming out gets pieced together from a set and map, which lost the distance context. In most cases this was fine, because another distance sort happens during route card data creation, but in cases like the GL at Hynes, there are multiple bus routes which have a nearest stop at Kenmore. Since the GL Kenmore platforms were also within the radius and the bus routes patterns were being processed first, the GL Kenmore platforms were being returned above Hynes, then the route card processing would assume that was the closest GL stop and filter out Hynes.

This change simply sorts the filtered stop IDs based on the original stop ID list.

iOS

android

### Testing

Verified that Hynes now shows up properly when it's the nearest GL stop

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/NearbyRepository.kt
@@ -40,7 +40,7 @@ class NearbyRepository : KoinComponent, INearbyRepository {
             nearbyLeafStops = findLeafStops()
         }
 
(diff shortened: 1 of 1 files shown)


boringcactus (approved):
Glad this fix turned out simple.

PR #1180, opened 2025-07-28
feat(android): Default to favorites first time only

KaylaBrady:
### Summary

_Ticket:_ [ 🤖 | Favorites | Default to favorites first time only](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210767320164830?focus=true)

What is this PR for?

This PR includes the android implementation for defaulting to favorites the first time only. As [discussed in slack](https://mbta.slack.com/archives/C062QNAJZ2M/p1753384488495949), using a new repo to determine whether the user has viewed favorites or not rather than tying it to the launch promo. 

There was a bit of UI jankiness to work through for this new loading state on android, so breaking it off here for review before addressing possible UI jank on the iOS side. 

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally, simulating version 2.0.0 with enhancedFavorites setting forced to true. Confirmed that favorites is shown by default the first time, but not the second time.
https://github.com/user-attachments/assets/b5be686c-bb44-437e-a2ae-e83cb6b1a5c9

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
@@ -349,6 +349,7 @@ fun MapAndSheetPage(
 
     fun navigateToEntrypoint(entrypoint: SheetRoutes.Entrypoint) {
         try {
+            Log.i("KB", "navigateToEntrypoint ${entrypoint}")
             navController.navigate(entrypoint, popUp)
         } catch (e: IllegalStateException) {
             // This should only happen in tests when the navigation graph hasn't been initialized
@@ -376,6 +377,7 @@ fun MapAndSheetPage(
         backgroundTimestamp?.let {
             val timeSinceBackground = clock.now().minus(Instant.fromEpochMilliseconds(it))
             if (timeSinceBackground > 1.hours && sheetNavEntrypoint != null) {
+                Log.i("KB", "lifeCycleResume navigateToEntrypoint")
                 navigateToEntrypoint(sheetNavEntrypoint)
                 if (nearbyTransit.locationDataManager.hasPermission) {
                     coroutineScope.launch { nearbyTransit.viewportProvider.follow() }
@@ -394,7 +396,9 @@ fun MapAndSheetPage(
     LaunchedEffect(nearbyTransit.alertData) { mapViewModel.alertsChanged(nearbyTransit.alertData) }
 
     LaunchedEffect(sheetNavEntrypoint) {
-        if (sheetNavEntrypoint != null) {
+        if (sheetNavEntrypoint != null && previousNavEntry != null) {
+            Log.i("KB", "sheetNavEntrypoint changed navigateToEntrypoint")
+
             navigateToEntrypoint(sheetNavEntrypoint)
         }
     }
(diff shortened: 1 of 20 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -423,189 +437,235 @@ fun MapAndSheetPage(
...
+                            initialStopId == targetStopId &&
+                            initialStopFilter?.routeId == targetStopFilter?.routeId
+                    ) {
+                        EnterTransition.None
                     }
-                    if (filters.stopFilter != null) {
-                        analytics.track(AnalyticsScreen.StopDetailsFiltered)
+
+                    if (initialRoute == targetRoute) {
+                        EnterTransition.None
```
KaylaBrady, line 502: Added to reduce jitter. Otherwise, nearby page was sliding in after the loading tab state, which was jarring.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -423,189 +437,235 @@ fun MapAndSheetPage(
...
+            }
+        }
+    }
 
-                fun navigate(route: SheetRoutes) {
-                    navController.navigateFrom(SheetRoutes.Favorites::class, route)
-                }
+    @Composable
+    fun SheetContent(modifier: Modifier = Modifier) {
+        if (sheetNavEntrypoint == null) {
```
EmmaSimon, line 465: nitpick(non-blocking): Could you refactor this to not add the extra indent to the `NavHost`? It makes it a bit difficult to review, but this composable already has too much going on and gets pretty cozy with the line limit even without the extra indent. Maybe `SheetContent` can continue conditionally showing loading state, but the `NavHost` is split out into a separate composable? Or ideally you could probably even split the `NavGraphBuilder.() -> Unit` out into a separate function.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -423,189 +437,235 @@ fun MapAndSheetPage(
...
+                Column(
+                    verticalArrangement = Arrangement.spacedBy(8.dp),
+                    modifier = Modifier.loadingShimmer(),
                 ) {
-                    EnterTransition.None
-                } else {
-                    slideIntoContainer(
-                        AnimatedContentTransitionScope.SlideDirection.Up,
-                        animationSpec = tween(easing = EaseInOut),
+                    SheetHeader(title = stringResource(R.string.got_it))
```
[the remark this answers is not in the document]
KaylaBrady: Ah that was from testing to make sure the title was properly hidden - will switch back to Nearby Transit

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/TabPreferencesRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+enum class DefaultTab {
+    Favorites,
+    Nearby;
+
+    fun toEntrypoint(): SheetRoutes.Entrypoint {
+        return when (this) {
+            Favorites -> SheetRoutes.Favorites
+            Nearby -> SheetRoutes.NearbyTransit
+        }
+    }
```
EmmaSimon, line 22: nitpick: Does this need to be a function or could you do a `val` with `get() = { when ... }`? I've been trying to avoid functions when there are no arguments and the value never changes like this.

PR #1182, opened 2025-07-29
feat: eastern time all the time

boringcactus:
### Summary

_Ticket:_ [Eastern time, all the time](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816091144019?focus=true)

I’m sorry I keep making these massive PRs.

I probably could’ve just centralized the formatting to use Eastern Time without also introducing the custom `EasternTimeInstant`, but I think the custom `EasternTimeInstant` is clearer and also simpler to create with a specific value.

iOS

android

### Testing

Checked on both iOS and Android that things that are about to happen show two hours later. Updated unit tests to assert that the correct time is shown.


EmmaSimon (approved):
Yeah this is a lot, thanks for tracking all of the usages down 😵 I appreciate that stuff like initializers and addition/subtraction are a bit more streamlined and consistent across platforms now.

PR #1183, opened 2025-07-30
feat(ios): Default to favorites if shown promo

KaylaBrady:
### Summary

_Ticket:_ [ 🤖 | Favorites | Default to favorites first time only](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210767320164830?focus=true)

What is this PR for?
* iOS version of https://github.com/mbta/mobile_app/pull/1180. 
* The iOS sheet component structure & animation is a bit different than android. I spent a little time on getting the loading state to line up with the nearby transit loading state, but it isn't perfect (ex: no tab bar). I never saw this loading state unless I purposefully added delays when loading the default tab preference or loading promo data, so I left those pieces out, but lmk if you have concerns / I should spend more time on that!

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally, changing the version to 2.0.0 to confirm default to favorites after showing promo. 

https://github.com/user-attachments/assets/4b79dbbe-9214-4f8a-ba76-26c6a8bba298


iosApp/iosApp/ContentView.swift
```diff
@@ -194,7 +206,7 @@ struct ContentView: View {
     }
 
     @ViewBuilder
-    var navSheetContents: some View {
+    var sheetContents: some View {
```
KaylaBrady, line 209: sheetContents uses `lastSafe` which defaults to `.nearby`. The easiest change seemed to be only use sheetContents if we know the default tab is loaded, and therefore `lastSafe` will accurately reflect whether nearby or favorites is last.

boringcactus (approved):

PR #1190, opened 2025-07-30
feat(iOS): Add favorites to launch screen

EmmaSimon:
### Summary

_Ticket:_ [ Favorites | UI Polish](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210710490021005?focus=true)

This just includes the added tab in the storyboard, no other pieces. 

Storyboards are annoying, I'm glad we don't have to edit these more. As part of this, I made it so that the selected tint color is Deemphasized, so that none of the tabs appears selected on the launch screen.

Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 16 05 29

iOS

### Testing

Confirmed that it shows up on launch, and matches the layout when the actual app loads


KaylaBrady (approved):
Good call on not styling any tab as selected!

PR #1192, opened 2025-07-31
feat: animate star button when state changes

boringcactus:
### Summary

_Ticket:_ [Favorites | animated star press](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210789765319254?focus=true)

I’m happy with how this looks.

Screen recording [in Slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1753998950677019?thread_ts=1753993811.415389&cid=C05QMG9GS9M).

iOS

android

### Testing

Checked that the animation looks reasonable on both Android and iOS and doesn’t play unnecessarily in the filtered stop details header when loading finishes or when switching between directions.


BrandonTR:
I noticed it on iOS. It seems to happen under specific circumstances. If you add a new favorite from nearby list then navigate into stop details, it animates. Here's a video of it happening on the Orange Line

https://github.com/user-attachments/assets/d58e47a2-ec71-4b49-85bd-edfcf4475264

PR #1195, opened 2025-08-01
fix: show status for CR even for times in the past

boringcactus:
### Summary

_Ticket:_ [Display scheduled Commuter Rail times with status when prediction has null times but non-null boarding_status](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210826290372849?focus=true)

It seems like, if we want to show schedules in the past that still have a status, it’d also be correct to show predictions in the past that still have a status.

iOS

android

### Testing

Added unit tests for new logic. Verified that old CR schedules with a current status are correctly shown in the filtered stop details tile list.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
@@ -42,6 +42,12 @@ sealed class TripInstantDisplay {
     data class ScheduleTime(val scheduledTime: EasternTimeInstant, val headline: Boolean = false) :
         TripInstantDisplay()
 
+    data class ScheduleTimeWithStatus(
+        val scheduledTime: EasternTimeInstant,
+        val status: String,
+        val headline: Boolean = false,
+    ) : TripInstantDisplay()
+
     data class ScheduleMinutes(val minutes: Int) : TripInstantDisplay()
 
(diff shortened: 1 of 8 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
```diff
@@ -115,6 +121,19 @@ sealed class TripInstantDisplay {
                 }
             }
             val predictionTime = prediction.stopTimeAfter(now)
+            if (allowTimeWithStatus && prediction.status != null) {
```
KaylaBrady, line 124: question: I see `allowTimeWithStatus` only is for the stop details view - is there any reason we wouldn't want to show a delayed scheduled CR trip in nearby transit too?

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataLeafTest.kt
```diff
@@ -164,6 +164,116 @@ class RouteCardDataLeafTest {
         )
     }
 
+    @Test
+    fun `formats CR as prediction in past with status`() {
```
KaylaBrady, line 168: question: I don't know enough about CR statuses - does this happen frequently? Wonder if it would result in most-all CR departures remaining in the tiles for a while after the train has actually departed

KaylaBrady (commented):
Appreciate the testing! A few questions, maybe more for product follow-up

PR #1218, opened 2025-08-05
fix(iOS): Improve route picker transition

BrandonTR:
### Summary

_Ticket:_ [ | Favorites | Mode picker => route picker transition](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210964905642161?focus=true)

This makes the animation more consistent with other navigation behavior.

### Testing

N/A


KaylaBrady:
Looks like the failing iOS test was the one that will be fixed in https://github.com/mbta/mobile_app/pull/1216... the outage delayed merging that but it should be in soon!

PR #1221, opened 2025-08-05
fix: follow-up work on commuter rail status

boringcactus:
### Summary

_Ticket:_ [Display scheduled Commuter Rail times with status when prediction has null times but non-null boarding_status](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210826290372849?focus=true)

Adds the formatting in nearby transit called for in [Adam’s mockup](https://mbta.slack.com/archives/C05QMG9GS9M/p1754346288816529?thread_ts=1754073004.733619&cid=C05QMG9GS9M) that was deferred in #1195, plus adds some previews and tests so we can be more confident that this actually works even in the absence of real issues to hit this case.

iOS

android

### Testing

Checked that the previews look correct and the tests pass.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
@@ -42,10 +42,15 @@ sealed class TripInstantDisplay {
     data class ScheduleTime(val scheduledTime: EasternTimeInstant, val headline: Boolean = false) :
         TripInstantDisplay()
 
-    data class ScheduleTimeWithStatus(
+    data class ScheduleTimeWithStatusColumn(
         val scheduledTime: EasternTimeInstant,
         val status: String,
         val headline: Boolean = false,
+    ) : TripInstantDisplay()
+
+    data class ScheduleTimeWithStatusRow(
+        val scheduledTime: EasternTimeInstant,
+        val status: String,
     ) : TripInstantDisplay()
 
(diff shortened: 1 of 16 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
```diff
@@ -78,7 +83,31 @@ sealed class TripInstantDisplay {
...
+                if (routeType == RouteType.COMMUTER_RAIL) {
+                    when {
+                        predictionTime != null && allowTimeWithStatus ->
+                            return TimeWithStatus(
+                                predictionTime,
+                                prediction.status,
+                                showTimeAsHeadline,
+                            )
+                        predictionTime != null -> return Time(predictionTime, showTimeAsHeadline)
+                        scheduleTime != null && allowTimeWithStatus ->
```
KaylaBrady, line 99: suggestion(non-blocking): `allowTimeWithStatus` seems like a misnomer now if there is a case where a ScheduleTimeWithStatus can be returned even if `allowTimeWithStatus` is false. 

Since it is only used in the CR case, maybe this variable can be removed entirely and the checks here could be `predictionTime != null && context = Context.StopDetailsFiltered`

KaylaBrady (approved):
Appreciate the testing & previews!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripInstantDisplay.kt
boringcactus: Makes sense.

PR #1242, opened 2025-08-13
feat: Consolidate route + mode a11y labels

EmmaSimon:
### Summary

_Ticket:_ [ | Favorites | Address VoiceOver QA](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211025820891516?focus=true)

This fixes the issue raised in QA where ferry pills are always ready out as "X ferry ferry" by the screen reader. This was happening because ferry is the only mode which includes the vehicle type in the name. 

This was also an issue in a handful of other places, this change finds all the spots where we're joining route names with the vehicle type for screen readers and makes them all follow the same shared code path. The one special case that's still doing a redundant ferry check is in StopDetailsFilteredHeader.swift, because that's using a distinct translation string which also includes the "at [stop]" at the end. Android can use the helper though, because it simply joins the pill semantics with a separate "at [stop]" string.

This does also lose the ability to pluralize "ferries" on a route, since the type is removed and the "ferry" is included in the route name. I also considered stripping off "Ferry" from the route name, but that seemed more fiddly and not worth it.

iOS

android

### Testing

Added unit tests for shared and platform specific code


KaylaBrady (approved):
Really appreciate your consideration for the effects in other places pills are used and your consolidation of this logic!

PR #1246, opened 2025-08-14
chore: cleanup unused pinned params

KaylaBrady:
### Summary

_Ticket:_ [Name](URL)

What is this PR for?

iOS

android

### Testing

What testing have you done?


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/PatternSortingTest.kt
```diff
@@ -209,98 +207,31 @@ class PatternSortingTest {
...
-                sortOrder = 1,
-            )
-        val routeCard7 =
-            routeCard(
-                pinned = false,
-                service = Service.Yes,
-                subway = true,
-                near = true,
-                sortOrder = 1,
-            )
```
KaylaBrady, line 267: routeCard7 was redundant to routeCard1 once pinned removed

PR #1247, opened 2025-08-14
fix: remove twist from top and bottom less common stops dropdowns

BrandonTR:
### Summary

_Ticket:_ [ | Favorites | Address UI QA - less common stops](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211025684993059?focus=true)

Removes the twists from the top and bottom stops in the route list on iOS and Android.
Screenshot_20250814_114844
Simulator Screenshot - iPhone 16 Pro - 2025-08-14 at 11 48 10

### Testing

What testing have you done?

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListGroupToggle.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListGroupToggle.kt
@@ -57,7 +57,7 @@ fun StopListGroupToggle(
         ),
     label: @Composable RowScope.() -> Unit,
 ) {
-    val showTwist =
+    val hideTwist =
         stickConnections.any { (connection, twisted) ->
             twisted &&
                 (connection.fromVPos != RouteBranchSegment.VPos.Top ||
@@ -91,7 +91,7 @@ fun StopListGroupToggle(
             ) {
                 if (it) 90f else 0f
             }
-        if (showTwist) {
+        if (hideTwist) {
             Box(modifier = modifier.size(34.dp), contentAlignment = Alignment.Center) {
                 if (stopsExpanded) {
                     Box(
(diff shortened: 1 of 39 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListGroupToggle.kt
```diff
@@ -50,8 +55,10 @@ fun StopListGroupToggle(
                 true,
             )
         ),
+    stopPlacement: StopPlacement = StopPlacement(),
     label: @Composable RowScope.() -> Unit,
 ) {
+    val isFirstOrLast = stopPlacement.isFirst || stopPlacement.isLast
```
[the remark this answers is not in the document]
BrandonTR: I had forgotten about the 33 case, thanks for catching that Emma! And thanks for the solution Melody!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListGroupToggle.kt
```diff
@@ -52,6 +57,12 @@ fun StopListGroupToggle(
         ),
     label: @Composable RowScope.() -> Unit,
 ) {
+    val hideTwist =
+        stickConnections.any { (connection, twisted) ->
+            twisted &&
+                (connection.fromVPos != RouteBranchSegment.VPos.Top ||
+                    connection.toVPos != RouteBranchSegment.VPos.Bottom)
+        }
```
EmmaSimon, line 65: nitpick: If you define a val on the list of pairs like this in `RouteBranchSegment.kt`, you might be able to do this entirely in shared and not have to repeat it (at least I'm pretty sure the interop for this works in Swift)

```kotlin
val List<Pair<RouteBranchSegment.StickConnection, Boolean>>.terminatedTwist: Boolean
    get() =
        this.any { (connection, twisted) ->
            twisted &&
                (connection.fromVPos != RouteBranchSegment.VPos.Top ||
                    connection.toVPos != RouteBranchSegment.VPos.Bottom)
        }
```

boringcactus: Might actually be easier to check for this in `RouteDetailsStopList.Segment.twistedConnections()` and return `null` from that if we hit this case, since the outcome is that we hide the connections.

EmmaSimon: Oh, yeah I like that idea. It might be cool if we could still show a bypassing parallel route line over the toggle if one exists, but I don't think that's necessary if it makes things at all more complicated.

EmmaSimon (approved):

PR #1250, opened 2025-08-15
fix?(shared.MapViewModel): Move map side effects into background

KaylaBrady:
### Summary

_Ticket:_ [Event buffer overflow MapViewModel](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211010043650124?focus=true)

What is this PR for?
This PR *potentially* fixes the event buffer overflow issue we've been seeing in prod. I haven't been able to reproduce this locally, but my best bet is that some events were taking a long time to process and filling up the buffer, contributing to the overflow. 

I added some logging to the duration of processing events (in ms), and found events that affected the map position took > 1s. 
image

Potentially if a user is tapping around to a bunch of trips and navigating back and forth a lot, this could back up the event queue enough to cause an overflow. 

That is still somewhat speculative since I haven't been able to repro this issue, and the breadcrumbs in Sentry are somewhat varied, but this seems plausible. 

iOS

android

### Testing

What testing have you done?
* Ran locally and confirmed I could still smoothly navigate between stops / trips
* Confirmed in logging that the duration of processing each event was < 1ms
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
@@ -34,11 +34,15 @@ import com.mbta.tid.mbta_app.utils.ViewportManager
 import com.mbta.tid.mbta_app.utils.timer
 import com.mbta.tid.mbta_app.viewModel.MapViewModel.Event
 import com.mbta.tid.mbta_app.viewModel.MapViewModel.Event.RecenterType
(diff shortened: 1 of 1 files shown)


EmmaSimon (approved):
This seems promising, should we include it in 2.0.0? I noticed today that the Play Store is complaining about our crash rate and may be deprioritizing us in search because of it. 

Image

KaylaBrady:
Ooh good observation on the crash rate, I'll merge & we can flag to product on Monday as to whether to include with the favorites release or as a fast follow

PR #1259, opened 2025-08-19
fix: clear error banner when navigating between pages

KaylaBrady:
### Summary

_Ticket:_ [Clear error banner when navigating between pages](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209118361132271?focus=true)

What is this PR for?
This PR clears the error banner on each page change to address the confusing state where a new page loads completely normally, but there is still an error banner from the old page. 

### Note
Writing a unit test of this is hard on iOS. I'm thinking of just moving this VM into shared code to simplify testing - then I can make a `pageChanged` event, and clearing the state can be done entirely in shared code.

iOS

android

### Testing

What testing have you done?
* Added android unit test
* Forced an error on stop details page, confirmed locally that it clears when navigating out of stop details


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/TripRepository.kt
```diff
@@ -40,17 +40,19 @@ internal class TripRepository : ITripRepository, KoinComponent {
...
 
     override suspend fun getTrip(tripId: String): ApiResult<TripResponse> =
-        ApiResult.runCatching {
-            val response =
-                mobileBackendClient.get {
-                    url {
-                        path("api/trip")
-                        parameter("trip_id", tripId)
-                    }
+        ApiResult.Error(500, "oops")
```
KaylaBrady, line 43: TODO: undo this before review

PR #1264, opened 2025-08-20
chore: manual dependency updates

boringcactus:
### Summary

_Ticket:_ [Manual dependency updates](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1210210024736992?focus=true)

I’ve also permanently solved our Sentry version mismatch woes, after a silly amount of Gradle wrestling.

iOS

android

### Testing

Checked that the build still works and that the Sentry version stuff actually works.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -39,7 +39,6 @@ sentry {
     org = "mbtace"
     projectName = "mobile_app_android"
     authToken = System.getenv("SENTRY_AUTH_TOKEN")
-    autoInstallation { enabled = false }
 }
 
 android {
@@ -114,7 +113,6 @@ dependencies {
     implementation(libs.mapbox.turf)
     implementation(libs.okhttp)
     implementation(libs.playServices.location)
-    implementation(libs.sentry.android)
     implementation(libs.androidx.lifecycle.runtime.testing)
     debugImplementation(platform(libs.compose.bom))
     debugImplementation(libs.compose.ui.test.manifest)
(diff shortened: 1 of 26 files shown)


androidApp/build.gradle.kts
```diff
@@ -33,7 +39,20 @@ sentry {
...
+            val bareKMPConfig = configurations.detachedConfiguration(libs.sentry.kmp.get())
+            val resolvedDependencies = bareKMPConfig.incoming.resolutionResult
+            val resolvedModuleVersions =
+                resolvedDependencies.allComponents.mapNotNull { it.moduleVersion }
+            val transitiveSentryCore =
+                resolvedModuleVersions.find { it.module.toString() == "io.sentry:sentry" }
+            checkNotNull(transitiveSentryCore) {
+                    "Could not find io.sentry:sentry among ${resolvedModuleVersions.joinToString(prefix = "[", postfix = "]")}"
+                }
+                .version
```
boringcactus, line 53: I hate Gradle. I tried like a dozen times to force both `shared` and `androidApp` to use the latest version of the Sentry Java SDK, but none of the inexplicably numerous ways to do that that the Gradle docs mention were working, so the best we can do is detecting what Sentry Java version Sentry KMP is depending on and using that as the Sentry Java version to install via the Sentry Android Gradle plugin.

boringcactus: https://github.com/getsentry/sentry-kotlin-multiplatform/issues/450

PR #1276, opened 2025-08-27
fix(Android): handle setSheetPadding applying while animateToOverview queued

boringcactus:
### Summary

_Ticket:_ [🤖 | Map animation race condition?](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210805266121129?focus=true)

Ugh.

iOS

android

### Testing

Checked that some steps that used to sometimes reproduce this issue (have a favorite far away from your current location, expand the favorites sheet, tap that favorite) now do not reproduce this issue. Added a test which fails without the lazy `animateToOverview` options and passes with it.


BrandonTR (approved):
Nice finds!

PR #1280, opened 2025-08-28
fix?(ViewModel): move all other input state out of event flow

boringcactus:
### Summary

This is the last thing I can think of that might actually help with the root cause here. Any event that just writes to a `mutableStateOf` doesn’t need an event to do that.

### Testing

Checked that tests still pass. This has never been an issue we’ve had a reasonable ability to reproduce, so that’s about all I can do.


KaylaBrady:
@boringcactus this change is making me wonder... do we still need events at all? Like for the ErrorBannerVM, is there still value in having Event.ClearState and Event.SetSheetRoute be events?

PR #1281, opened 2025-08-28
fix?(ViewModel): stop crashing on event buffer overflows

boringcactus:
### Summary

When I was first setting up the Molecule infrastructure, the reason I thought it made sense to crash if there was an event buffer overflow was that the buffer shouldn’t overflow unless we’ve messed up somehow, like sending to a ViewModel but forgetting to read its state. In the absence of an identifiable cause for our event buffer overflows, we may conclude that sometimes the buffer just overflows and it needs to not be a huge deal.

I maintain my belief that event buffer overflows would cause absurdly irritating bugs if they just happened silently, so what I want is to continue crashing on event buffer overflows in debug builds (so that we can discover them if they are severe enough to notice for ourselves) but to just discard the oldest event if the event buffer overflows in release builds (so that users get a better experience than just the app crashing, and so that we can finally move on from this buffer overflow issue). Getting the answer to the question “is this a debug build or a release build?” all the way into the event flow construction is not fun, and doing all this just to maintain our ability to discover event buffer overflows if they happen to us will only pay off if we actually wind up overflowing an event buffer in our own development process. Maybe #1282 will be enough instead.

I think discarding the oldest event is better than discarding the newest event. If we were comfortable discarding the newest event, we could also record that in Sentry by continuing to use `SUSPEND` but just doing `Sentry.captureMessage` instead of `error` if the `trySend` fails, but there is no mechanism I can discover which will let us both detect that the buffer has overflowed and discard the oldest event.

Pairs nicely with #1280, which moves more input state changes out of the event flow, so that it’s less likely that some significant input state change will be lost.

### Testing

Checked that tests pass on both iOS and Android.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/TestUtils.kt
@@ -12,6 +12,7 @@ import androidx.compose.ui.test.assert
 import androidx.compose.ui.test.captureToImage
 import com.mbta.tid.mbta_app.analytics.Analytics
 import com.mbta.tid.mbta_app.analytics.MockAnalytics
+import com.mbta.tid.mbta_app.dependencyInjection.KoinName
 import com.mbta.tid.mbta_app.dependencyInjection.MockRepositories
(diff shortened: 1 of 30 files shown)


boringcactus:
Tracked down the actual bug in #1285. We are not surrendering today.

PR #1282, opened 2025-08-28
fix?(ViewModel): completely ignore event buffer overflows

boringcactus:
### Summary

If we magically knew that we would never actually have an event buffer overflow in local development, then there’d be no point doing #1281, and this would be adequate.

Pairs nicely with #1280 to reduce the likelihood that the event that gets dropped is vital input data.

### Testing

Not actually tested, it’s 5:30 pm and I need to log off.


boringcactus:
Tracked down the actual bug in #1285. We are not surrendering today.

PR #1284, opened 2025-08-29
fix(ViewModel): apply timeout to event processing

boringcactus:
### Summary

This serves as a backstop in case an event gets stuck for a long time in a `suspend fun` call.

iOS

android

### Testing

Added a test on the MapViewModel. Checked that existing tests pass.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
```diff
@@ -193,75 +194,73 @@ public class MapViewModel(
...
-                                followPuck(null)
-                            }
-                            is State.TripSelected -> {
-                                when (event.type) {
-                                    RecenterType.CurrentLocation -> followPuck(null)
-                                    RecenterType.Trip -> {
-                                        val currentState = state as State.TripSelected
-                                        handleViewportCentering(currentState, density)
-                                    }
+        EventSink(eventHandlingTimeout = 10.seconds, sentryRepository = sentryRepository) { event ->
```
EmmaSimon, line 197: 10 seems a bit excessive, isn't our animation duration 1 second?

EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
boringcactus: The animation duration is one second, and #1285 sets the timeout for a single animation holding the lock to be two seconds, but this timeout includes the time spent waiting for the lock, and if there are multiple animations that were already queued, it could reasonably take several seconds before the event handler finishes.

PR #1295, opened 2025-09-04
fix(android): Address a batch of flaky test

KaylaBrady:
### Summary

_Ticket:_ [CI | android flaky tests again](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211209894303557?focus=true)

What is this PR for?
This PR addresses the 4 itemized android tests that have been recently flaking and preventing PRs from merging.
It also broadens the usage of `waitUntilDefaultTimeout` to increase the timeout used across tests.

iOS

android

### Testing

What testing have you done?
* For each of the itemized fixed tests, ran each one for 300 repetitions using [RetryableComposeTestRule](https://github.com/mbta/mobile_app/blob/main/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/testUtils/RetryableComposeTestRule.kt) and confirmed there were no failures.

### Note
Seeing some failures on this PR and elsewhere on shared VM tests. I'm taking a closer look at those, but putting this up for review as-is to hopefully save us from a few CI failures in the meantime.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/SaveFavoritesFlowTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/SaveFavoritesFlowTest.kt
@@ -15,12 +15,16 @@ import com.mbta.tid.mbta_app.model.RouteCardData
 import com.mbta.tid.mbta_app.model.RouteStopDirection
 import com.mbta.tid.mbta_app.usecases.EditFavoritesContext
 import com.mbta.tid.mbta_app.utils.TestData
+import com.mbta.tid.mbta_app.viewModel.IToastViewModel
 import com.mbta.tid.mbta_app.viewModel.MockToastViewModel
 import com.mbta.tid.mbta_app.viewModel.ToastViewModel
+import dev.mokkery.MockMode
+import dev.mokkery.mock
+import dev.mokkery.verify
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
-import org.junit.Ignore
+import kotlinx.coroutines.runBlocking
 import org.junit.Rule
 import org.junit.Test
 import org.koin.compose.KoinContext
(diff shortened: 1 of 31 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/component/SaveFavoritesFlowTest.kt
```diff
@@ -376,36 +380,42 @@ class SaveFavoritesFlowTest {
...
-        val toastVM = MockToastViewModel()
-        var displayedToast: ToastViewModel.Toast? = null
-        toastVM.onShowToast = { displayedToast = it }
+    fun testFavoritingToastFallbackText() = runBlocking {
+        val toastVM = mock<IToastViewModel>(MockMode.autofill)
 
         val busRoute = TestData.getRoute("15")
         val busStop = TestData.getStop("17861")
+        // no TestData for getting the appropriate labels for this favorite
+        val koin = testKoinApplication()
```
KaylaBrady, line 395: This was probably the least intuitive of the 4 itemized tests. I first tried switching to mokkery to get a more helpful error message. The failure was because the message of the toast was the same as for the test above - `<b>Outbound 15 bus</b> at <b>Ruggles</b> added to Favorites`. I don't know how that was happening without the KoinContext being set since it should come from global data, but explicitly adding a koin context without global data resolved the issue.

KaylaBrady: Huh, I wonder if something recently changed? This particular test failure was strange because it would happen even if I only ran this one test repeatedly -- I'm not sure where else the `TestData` could have been injected from.

EmmaSimon: On iOS, I updated a bunch of tests to use the koin mocks because before they were able to get away with just passing in all the repositories but with the global modifier, they aren't. Most of them work fine but a few are having trouble. But yeah the fact that it's happening when running the test by itself makes me more skeptical that's what's happening, super weird.

KaylaBrady: Curiously now I'm not able to run that test at all locally 😵‍💫 maybe something about my config was entirely borked when running it initially, will try to get this running again and understand it a bit more.

BrandonTR (approved):

PR #1297, opened 2025-09-05
feat(iOS): Add favorites loading modifier

EmmaSimon:
### Summary

_Ticket:_ Prep for [Move StopDetailsViewModel to shared code](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210525354862284?focus=true)

Add a `.favorites` modifier similar to `.global`, which takes a binding of favorites data and populates it from the repo. This is a bit different from global data in that that has a shared flow in the repo which is consumed and can automatically update the bindings when the value changes. This needs to be told when to update the favorites data, so it uses a second boolean binding which the consuming view can change when it wants to fetch an update. There's probably a more elegant solution and we should probably get rid of this or completely rewrite it if/when we get around to implementing the shared favorites fetching/updating system we talked about recently.

One awkward thing that seemed a bit risky is the addition of the `LoadedFavorites.last` singleton which just updates with the last loaded value whenever it's loaded from the repo, so that there's less of a noticeable delay before the new state loads in. I could make this more robust/complicated if anyone thinks it's necessary (maybe injected through koin, or a VM like the route card data one??), but this seemed like a workable MVP, and especially if we rethink and rewrite our favorites handling soon, I didn't want to over engineer it.

iOS

### Testing

Existing tests pass, added new tests for the modifier, and a bonus test for the global modifier that I forgot.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Favorite.kt
```diff
@@ -2,7 +2,11 @@ package com.mbta.tid.mbta_app.model
 
 import kotlinx.serialization.Serializable
 
-@Serializable public data class Favorites(val routeStopDirection: Set<RouteStopDirection>? = null)
+@Serializable
+public data class Favorites(val routeStopDirection: Set<RouteStopDirection>? = null) {
+    public fun isFavorite(rsd: RouteStopDirection): Boolean =
```
boringcactus, line 7:
```suggestion
    public fun isFavorite(@ObjCName(swiftName = "_") rsd: RouteStopDirection): Boolean =
```

This will remove the need for the FavoritesExtension.

iosApp/iosApp/Pages/StopDetails/StopDetailsUnfilteredView.swift
```diff
@@ -26,6 +28,7 @@ struct StopDetailsUnfilteredView: View {
     @EnvironmentObject var settingsCache: SettingsCache
 
     @State var global: GlobalResponse?
+    @State var loadingFavorites = true
```
boringcactus, line 31:
```suggestion
```

I don’t think this state is used in this view.

iosApp/iosApp/Utils/Modifiers/FavoritesModifier.swift
```diff
@@ -0,0 +1,52 @@
...
+    static var last = Favorites(routeStopDirection: [])
+}
+
+struct FavoritesModifier: ViewModifier {
+    var favoritesRepository: IFavoritesRepository = RepositoryDI().favorites
+    @Binding var favorites: Favorites
+    @Binding var awaitingUpdate: Bool
+
+    func loadFavorites() {
+        guard awaitingUpdate else { return }
```
boringcactus, line 22: I think the fact that the `.onAppear()` won’t load the favorites if `awaitingUpdate` is `false` is counterintuitive and a little risky, and I’d rather move this check into the `onChange` in the short term.

boringcactus (commented):
Huh. I guess I hadn’t actually thought through how much of an improvement even this mildly janky SwiftUI version of our Android Compose state helpers would be; this is good to be working on.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Favorite.kt
EmmaSimon: Ahh, I was trying to do that with an empty string and it was failing, so I assumed it wasn't possible, nice!

PR #1305, opened 2025-09-09
test(Android): load Koin mocks into global Koin so they propagate to ViewModels

boringcactus:
### Summary

_Ticket:_ [🤖 | Track this Trip | Sheet contents](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210501361218460?focus=true), tangentially

Since the `KoinContext` is only inherited through Compose, it doesn’t make it down to the ViewModels.

iOS

android

### Testing

Checked that all the tests still pass.


EmmaSimon (approved):
Nice, appreciate the simplification 🙏

PR #1307, opened 2025-09-10
feat(iOS): Switch over to shared stop and trip VMs

EmmaSimon:
### Summary

_Ticket:_ [Move StopDetailsViewModel to shared code](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210525354862284?focus=true), follow up to #1273

This rips out the iOS specific `StopDetailsViewModel` and replaces it with the shared stop and trip details VMs. I pulled out some VM specific handling out into modifiers, and had to rearrange a lot of the state and params in stop details, but functionality should be exactly the same.

iOS

### Testing

Deleted iOS VM tests that are now covered by shared tests, and a few irrelevant tests that were testing VM functionality from the views. Got all existing tests passing with the new mock VMs


iosApp/iosApp/Pages/Map/HomeMapView.swift
```diff
@@ -123,9 +128,9 @@ struct HomeMapView: View {
             .onChange(of: nearbyVM.alerts) { _ in
                 handleGlobalMapDataChange(now: now)
             }
-            .onChange(of: nearbyVM.routeCardData) { _ in
```
EmmaSimon, line 126: Previously, `nearbyVM.routeCardData` was being set with both the nearby route cards, but also, kinda inappropriately, with the stop details rcd, and since the map only cares about the stop details rcd, it was just doing this filtering to distinguish them. Now, this uses the molecule `RouteCardDataViewModel` directly, which is only updated with stop details rcd (the name isn't the best tbh). I could probably remove the stop details nav check now, not sure if it's really adding any additional safety, unless we start updating the `RouteCardDataViewModel` also with nearby rcd.

iosApp/iosApp/Pages/StopDetails/StopDetailsFilteredView.swift
```diff
@@ -110,8 +116,12 @@ struct StopDetailsFilteredView: View {
                 loadingBody()
             }
         }
+        .task {
+            for await model in stopDetailsVM.models {
+                alertSummaries = model.alertSummaries as? [String: AlertSummary?] ?? [:]
```
EmmaSimon, line 121: `model.alertSummaries` has a type of `[String: Any]`, which was slightly awkward to work around.

PR #1321, opened 2025-09-15
fix: remove unnecessary use of OkHttp internal API

boringcactus:
### Summary

_Ticket:_ none

This appears to have snuck in in #557, and I don’t think it has any effect at all, since we aren’t otherwise using the JVM synchronization primitives (there’s no `wait` for this `notifyAll` to unblock).

Should unblock #1141 and #1311.

iOS

android

### Testing

Checked that the app still builds.


EmmaSimon (approved):
Ugh, thanks for digging into this, I noticed that it was an issue with that dependabot PR and found that we were using the deprecated internal API, but assumed there was a good reason for it and wasn't sure how to replace it.

PR #1327, opened 2025-09-18
fix(MapVM): use actual current time for alerts if alerts change

boringcactus:
### Summary

_Ticket:_ [🤖 | Reflect alerts on the map as quickly as in the sheet](https://app.asana.com/1/15492006741476/project/1201654106676769/task/1211292609179112?focus=true)

iOS

android

### Testing

Wrote a new test to reproduce the issue that went from failing to passing with this change.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
@@ -154,7 +154,7 @@ public class MapViewModel(
 
     @Composable
     override fun runLogic(): State {
-        val timerTime by timer(updateInterval = 300.seconds, clock = clock)
+        val alertCheckTimer by timer(updateInterval = 300.seconds, clock = clock)
         val globalData by globalRepository.state.collectAsState()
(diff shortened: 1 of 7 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
```diff
@@ -152,7 +154,7 @@ public class MapViewModel(
 
     @Composable
     override fun runLogic(): State {
-        val now by timer(updateInterval = 300.seconds)
+        val timerTime by timer(updateInterval = 300.seconds, clock = clock)
```
EmmaSimon, line 157: nitpick: Since this is only ever used for the alert checks, should we rename it to something like `alertCheckTimer`?

EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModel.kt
boringcactus: Sure.

PR #1329, opened 2025-09-18
refactor: centralize logic common to all channel subscriptions

boringcactus:
### Summary

_Ticket:_ [SwiftPhoenixClient/Channel.swift: tried to join multiple times. 'join' can only be called a single time per channel instance](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210905903879490?focus=true), indirectly

All of our channel subscriptions work basically the same way, and I’ve thought of a race condition that could cause the problems we’re seeing in that common structure, so it’d be nice if fixing it could be done only once.

iOS

android

### Testing

Checked that the existing tests still pass.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/PredictionsForStopsChannel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/PredictionsForStopsChannel.kt
@@ -3,24 +3,12 @@ package com.mbta.tid.mbta_app.phoenix
 import com.mbta.tid.mbta_app.json
 import com.mbta.tid.mbta_app.model.response.PredictionsByStopJoinResponse
 import com.mbta.tid.mbta_app.model.response.PredictionsByStopMessageResponse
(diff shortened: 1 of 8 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/PredictionsForStopsChannel.kt
```diff
@@ -5,31 +5,31 @@ import com.mbta.tid.mbta_app.model.response.PredictionsByStopJoinResponse
...
 
-internal class PredictionsForStopsChannel {
-    companion object {
-        val topic = "predictions:stops"
-
-        val newDataEvent = "stream_data"
-
-        fun topicV2(stopIds: List<String>) = "predictions:stops:v2:${stopIds.joinToString(",")}"
+internal object PredictionsForStopsChannel {
+    class V1(stopIds: List<String>) : ChannelSpec {
```
EmmaSimon, line 9: Do we need to keep v1 around anymore? Might be a good time to rip it out.

EmmaSimon (approved):
Nice, great consolidation 🙏

PR #1333, opened 2025-09-19
feat(iOS): tap vehicle to track trip

boringcactus:
### Summary

_Ticket:_ [ | Track this Trip | Tap vehicle on map to open](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211363468635850?focus=true)

Nice and straightforward with reference to #1309.

iOS

android

### Testing

Added a test for the new logic to match the test for the old logic.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
@@ -365,7 +365,7 @@ fun MapAndSheetPage(
 
         if (routeId != null) analytics.tappedVehicle(routeId)
 
-        val newTripFilter = TripDetailsFilter(tripId, vehicle.id, stopSequence)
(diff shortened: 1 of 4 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -365,7 +365,7 @@ fun MapAndSheetPage(
 
         if (routeId != null) analytics.tappedVehicle(routeId)
 
-        val newTripFilter = TripDetailsFilter(tripId, vehicle.id, stopSequence)
+        val newTripFilter = TripDetailsFilter(tripId, vehicle.id, stopSequence, true)
```
boringcactus, line 368: This `selectionLock = true` got lost in #1309; it probably won’t matter, since I doubt we’ll need to release without track this trip, but it’s worth putting back just in case.

PR #1356, opened 2025-09-30
feat: "Trip Complete" card on trip details

EmmaSimon:
### Summary

_Ticket:_ [[QA & release track this trip]](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211412336640357?focus=true)

When a trip finishes, show a card saying that the trip is done, rather than getting into a weird "completing another trip" route ladder state.

Simulator Screenshot - iPhone 16 Pro - 2025-09-30 at 14 43 43 Screenshot_20250930_143958

iOS

android

### Testing

Added unit tests on both platforms

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripDetailsView.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripDetailsView.kt
@@ -17,6 +17,7 @@ import com.mbta.tid.mbta_app.analytics.Analytics
 import com.mbta.tid.mbta_app.android.ModalRoutes
 import com.mbta.tid.mbta_app.android.component.DebugView
 import com.mbta.tid.mbta_app.android.state.getGlobalData
+import com.mbta.tid.mbta_app.android.tripDetails.TripCompleteCard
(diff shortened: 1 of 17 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/stopDetails/TripDetailsView.kt
```diff
@@ -235,35 +238,43 @@ fun TripDetails(
...
-                headerSpec,
-                now,
-                alertSummaries,
-                globalResponse,
-                onTapStop,
-                onOpenAlertDetails,
-                route,
-                routeAccents,
-            )
+        val hasCompletedTrip =
```
boringcactus, line 241: It seems like this will also be true in the finishing another trip state, so “hasCompletedTrip” is a confusing name for it.

EmmaSimon: Yeah, that's true, it's just that you can't get into the track this trip view when finishing another trip. How about `onOtherTrip` or `vehicleOnOtherTrip` or something?

boringcactus (approved):

PR #1609, opened 2026-03-19
feat: Notifications beta opt in flow

EmmaSimon:
### Summary

_Ticket:_ [Notifications | beta opt-in flow](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1212944835007237?focus=true)

Add a flow to funnel users into enabling the notifications beta, then ask for feedback contact and display a hint on the favorites page if it's enabled.

This included a few minor changes that added up a bit, and all the newly added strings really ballooned Localizeable.xcstrings, so the line count is pretty scary. I could try to split out a separate iOS PR, but a lot of the changes to shared will require a chunk of those changes in order to compile.

At a high level, the changes were:
* A couple new types added to `ToastAction` for allowing actions to be triggered by tapping the body of the toast (with or without a close button)
* A highlighted section param on the more page
* `OnboardingRepository` updated to handle loading and saving all the one time visible toast/dialog/hint state to preferences (some of these could have been split out into another repo maybe, but they're all quite onboarding-esque, and they interact with the notifications onboarding page)
* `NotificationsBetaViewModel` with a state containing booleans for if the toast and dialog should be visible
* A modifier in iOS and a composable in Android for displaying the toast or feedback dialog if appropriate
* `FavoritesViewModel` updated to include a boolean state for showing the hint
* The favorites page updated to show the hint if it's true
* Notifications onboarding animations updated to restart when the app is backgrounded (they would happily play in the background and be done by the time you returned from filling out the feedback form)
* A targeting class that uses Firebase installation IDs to get a true boolean for some percentage of users
* A debug only button that wipes out the saved toast, dialog, and hint state, sets the notification onboarding page back to pending, and sets a preference to override the ID based targeting and show you the toast, for QA purposes

Screenshot_20260319_142437 Screenshot_20260319_142451 Screenshot_20260319_142541

iOS

android

### Testing

Tested the flow manually on both platforms, tested the targeting manually by deleting app data repeatedly to generate new IDs, added unit tests for shared and UI code


BrandonTR (commented):
Just had one optional nitpick. This looks really good!
One question, I think I remember this being brought up but can't remember the answer: Do we want to show the dialog again if the user turns notifications off and on?

BrandonTR (approved):

PR #1657, opened 2026-04-10
fix(RouteCardData): Include both directions at all temporary terminals

KaylaBrady:
### Summary

No ticket, noticed while looking into [Notifications QA | trip-specific suspensions alerts replace all departures](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213983085131248?focus=true) while a planned CR disruption is taking place.

What is this PR for?

Currently, a planned CR disruption is ~omitted from the app for the first affected stop - not shown in stop details, favorites, or nearby transit. 

This was due to subway-only temporary terminal logic applied in https://github.com/mbta/mobile_app/pull/739. That subway-only condition was added because at the time, not all routes had typical patterns, and we were seeing arrival-only directions displayed at regular ferry & bus terminals. Now, all routes have a typical pattern, so we don't need to special-case subway anymore. 

iOS

android

### Testing

What testing have you done?
| Before | After |
|--------|--------|
|image | image |

* Also confirmed that for the ferry & bus routes included in the old PR, the arrival-only direction is still omitted.

### DIFF
--- a/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
+++ b/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
@@ -5427,121 +5427,119 @@ class RouteCardDataTest {
         }
 
     @Test
-    fun `RouteCardData routeCardsForStopList filters out any arrival only for non-subway routes`() =
-        runBlocking {
-            val objects = ObjectCollectionBuilder()
+    fun `RouteCardData routeCardsForStopList filters out any arrival only`() = runBlocking {
+        val objects = ObjectCollectionBuilder()
 
-            val longWharf = objects.stop { id = "Boat-Long" }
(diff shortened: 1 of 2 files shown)


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -5427,121 +5427,119 @@ class RouteCardDataTest {
         }
 
     @Test
-    fun `RouteCardData routeCardsForStopList filters out any arrival only for non-subway routes`() =
-        runBlocking {
-            val objects = ObjectCollectionBuilder()
+    fun `RouteCardData routeCardsForStopList filters out any arrival only`() = runBlocking {
```
KaylaBrady, line 5430: just reformatted due to shorter name

boringcactus (approved):

PR #1658, opened 2026-04-10
fix: omit typical shuttles with no service

BrandonTR:
### Summary

_Ticket:_ [Omit shuttle buses with no service today](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213822527827960?focus=true)

Omit shuttles with no service today.

iOS

android

### Testing

Added tests for new logic.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LineOrRoute.kt
```diff
@@ -96,6 +96,16 @@ public sealed class LineOrRoute {
                 is Route -> this.route.type.isSubway()
             }
 
+    internal val isShuttle: Boolean
+        get() =
+            when (this) {
+                is Line ->
+                    this.routes.any {
+                        it.id.idText.startsWith("Shuttle")
```
BrandonTR: Should this be hardcoded to false instead?

KaylaBrady: Yeah that seems safer! The line data type does include all pre-defined shuttle routes (ex: https://api-v3.mbta.com/lines/line-Green?include=routes), so probably best to just hardcode.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/LineOrRoute.kt
```diff
@@ -96,6 +96,16 @@ public sealed class LineOrRoute {
...
+                    } // Should this just hard code to false?
+                is Route -> this.id.idText.startsWith("Shuttle")
```
KaylaBrady: suggestion(non-blocking): Could use this.route.isShuttle

KaylaBrady (approved):

PR #1659, opened 2026-04-10
fix: Trip-specific suspension & cancellation alerts take over just the affected trip

KaylaBrady:
### Summary

_Ticket:_ [Notifications QA | trip-specific suspensions alerts replace all departures](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213983085131248?focus=true)

What is this PR for?

WIP - getting closer, but need to block trip details 
image

iOS

android

### Testing

What testing have you done?

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -54,7 +54,10 @@ internal constructor(
     public fun allClear(atTime: EasternTimeInstant): Boolean =
         activePeriod.all { it.end != null && it.end < atTime }
 
-    public fun significance(atTime: EasternTimeInstant?): AlertSignificance {
+    public fun significance(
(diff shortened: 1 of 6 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -54,7 +54,10 @@ internal constructor(
     public fun allClear(atTime: EasternTimeInstant): Boolean =
         activePeriod.all { it.end != null && it.end < atTime }
 
-    public fun significance(atTime: EasternTimeInstant?): AlertSignificance {
+    public fun significance(
+        atTime: EasternTimeInstant?,
+        tripId: String? = null,
```
KaylaBrady, line 59: I'm not crazy about adding tripId in here. I could instead put more logic into DisplayAlert, because ultimately the effect I want is that the cancellation for this trip is displayed as a takeover alert.

PR #1662, opened 2026-04-13
fix: request stop and trip maps only when actually needed

boringcactus:
### Summary

_Ticket:_ [reduce excessive calls to /api/trip/map-friendly, /api/stop/map ](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213993311573670?focus=true)

I’m glad this was such an easy fix.

iOS

android

### Testing

Checked with the local backend that the data is no longer fetched more often than it needs to be. Added a unit test for the stop data specifically and confirmed that it fails on `main` but passes on this PR.


KaylaBrady (approved):
Sweet!

PR #1663, opened 2026-04-13
refactor(Alert): Separate intrinsicSignificance from significanceAtTime, new significanceForTrip

KaylaBrady:
### Summary

What is this PR for?
In service of https://github.com/mbta/mobile_app/pull/1659

While working on that PR, I was tempted to add a new `tripId` param to the existing `significance(atTime: EasternTimeInstant?)` fun. That didn't feel quite right to me though - depending on the call site, we want either: 
1. significance independent of the current time
2. significance considering the current time
3. significance considering the currently selected trip

This PR makes those 3 separate functions. Now that I've split it all out, I could see merging 1&2 back into a `significanceAtTime` fun that takes an optional time, but I think it is a bit clearer and perhaps easier to choose which one you really want in a particular place with the separate names

iOS

android

### Testing

What testing have you done?
* Added unit tests for the new fun

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -54,46 +54,50 @@ internal constructor(
     public fun allClear(atTime: EasternTimeInstant): Boolean =
         activePeriod.all { it.end != null && it.end < atTime }
 
-    public fun significance(atTime: EasternTimeInstant?): AlertSignificance {
-        val intrinsicSignificance =
-            when (effect) {
-                // suspensions or shuttles can reasonably apply to an entire route
-                in setOf(Effect.Shuttle, Effect.Suspension) -> AlertSignificance.Major
-                // detours and closures are only major if they specify stops
-                in setOf(
-                    Effect.StationClosure,
-                    Effect.StopClosure,
-                    Effect.DockClosure,
-                    Effect.Detour,
-                    Effect.SnowRoute,
-                ) -> if (hasStopsSpecified) AlertSignificance.Major else AlertSignificance.Secondary
-                // service changes are always secondary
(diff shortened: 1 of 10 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -497,7 +497,8 @@ constructor(val trip: Trip, val stops: List<Entry>, val startTerminalEntry: Entr
                             )
                         } &&
                         // there's no UI yet for secondary alerts in trip details
-                        alert.significance(entryTime) >= AlertSignificance.Major
+                        (entryTime?.let { alert.significanceAtTime(it) }
+                            ?: alert.intrinsicSignificance) >= AlertSignificance.Major
```
KaylaBrady, line 501: This is the one place it is kind of icky to have separated # 1 and # 2

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -54,51 +54,65 @@ internal constructor(
...
     }
 
+    public fun significanceForTrip(tripId: String): AlertSignificance {
+
+        if (anyInformedEntitySatisfies { checkTrip(tripId) }) {
+            if (effect == Effect.Cancellation) {
+                return AlertSignificance.Major
+            }
+        }
+        return intrinsicSignificance
```
KaylaBrady, line 113: Now that I use this, I think I should be falling back to significanceForTime(now) when the tripId doesn't match the alert so that we treat shuttles later today as secondary rather than major.

KaylaBrady: So maybe I've worked myself back into the signature of `significance(atTime, tripId?) after all

PR #1674, opened 2026-04-14
fix: Try downgrading ruby to fix iOS builds

EmmaSimon:
### Summary

_Ticket:_ No ticket

Follow up #1665, the upgraded ruby version seems to have broken iOS builds.


EmmaSimon:
Dev orange build passed for this branch, so I think this should work (it's upgraded from what it was previously, and the rubocop upgrade works with it)

KaylaBrady (approved):

PR #1675, opened 2026-04-15
fix(StopDetailsFiltered): Don't duplicate cancellation alerts

KaylaBrady:
### Summary

_Ticket:_ [Notifications QA | fix presentation of cancellation](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213994843939443?focus=true)

What is this PR for?

This PR updates DisplayAlerts to treat trip-specific cancellations as takeovers. The custom trip cancellation card is then shown only if there is not a takeover alert present. This change includes some refactors to significance, which I think will help with understanding the different ways to slice significance.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran in iOS & android

imageimage

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
@@ -497,7 +497,8 @@ constructor(val trip: Trip, val stops: List<Entry>, val startTerminalEntry: Entr
                             )
                         } &&
                         // there's no UI yet for secondary alerts in trip details
-                        alert.significance(entryTime) >= AlertSignificance.Major
+                        (entryTime?.let { alert.significance(it) }
+                            ?: alert.intrinsicSignificance) >= AlertSignificance.Major
                 }
             if (alert == null) return null
             return UpcomingFormat.Disruption(alert, route?.let { MapStopRoute.matching(it) })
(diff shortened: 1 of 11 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -497,7 +497,8 @@ constructor(val trip: Trip, val stops: List<Entry>, val startTerminalEntry: Entr
                             )
                         } &&
                         // there's no UI yet for secondary alerts in trip details
-                        alert.significance(entryTime) >= AlertSignificance.Major
+                        (entryTime?.let { alert.significance(it) }
+                            ?: alert.intrinsicSignificance) >= AlertSignificance.Major
```
KaylaBrady, line 501: This is the most awkward change from making `significance` take a non-nullable time. Overall though I found separately naming  `intrinsicSignificance` helpful for my understanding of what functions were doing and determining whether I wanted the intrinsicSignificance or significance at time.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/DisplayAlertTest.kt
```diff
@@ -77,6 +77,29 @@ class DisplayAlertTest {
         assertEquals(AlertCardSpec.Takeover, DisplayAlert(tripAlert).cardSpec(now, true, "trip1"))
     }
 
+    @Test
+    fun `cardSpec cancellation for selected trip is takeover`() = runBlocking {
+        val tripAlert =
+            objects.alert {
+                id = "hereForTargetTrip"
+                effect = Effect.Cancellation
+                activePeriod = mutableListOf(Alert.ActivePeriod(now.minus(10.minutes), null))
```
boringcactus: I think it would make sense to just call `activePeriod()` and `informedEntity()` instead of setting the lists.

boringcactus (approved):

PR #1676, opened 2026-04-15
fix(RouteCardData): Treat trip-specific major as secondary in alert card

KaylaBrady:
### Summary

_Ticket:_ Follow-up on [Notifications QA | trip-specific suspensions alerts replace all departures](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213983085131248?focus=true)

What is this PR for?

It was a little strange to not see a trip-specific cancellation or suspension reflected in favorites or nearby transit when it was previously. This updates the secondary alert logic to include major trip-specific alerts. 

iOS

android

### Testing

What testing have you done?
* Updated unit test, ran locally
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -182,8 +182,10 @@ public data class RouteCardData(
 
         private fun secondaryAlertToDisplay(atTime: EasternTimeInstant) =
             alertsHere.firstOrNull {
-                it.significance(atTime) < AlertSignificance.Major &&
-                    it.significance(atTime) >= AlertSignificance.Secondary
+                (it.significance(atTime) < AlertSignificance.Major &&
+                    it.significance(atTime) >= AlertSignificance.Secondary) ||
+                    (it.significance(atTime) == AlertSignificance.Major &&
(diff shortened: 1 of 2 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -182,8 +182,10 @@ public data class RouteCardData(
 
         private fun secondaryAlertToDisplay(atTime: EasternTimeInstant) =
```
KaylaBrady, line 183: Note: `secondaryAlertToDisplay` overloads the term "secondary", which is also a significance. This already returned non-secondary alerts (downstream), and after this change expands into alerts that aren't secondary significance. What do you think about renaming this concept? Something like `warningAlert`?

BrandonTR (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
BrandonTR: `warningAlert` makes sense, I'm struggling to come up with anything better 😅

KaylaBrady: TY! I'll throw that change into a separate PR so that it can be a pure refactor

PR #1678, opened 2026-04-16
feat(Notifications): Use string fallback message when failure to parse summary

KaylaBrady:
### Summary

_Ticket:_ [Add handling of alert header fallback notification payload](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214048781923338?focus=true)

What is this PR for?

Changes the unknown summary type from `fallback: AlertSummary` to `fallback: String` so that it is easier to manage backwards compatibility of introducing new summary types in the future.

iOS

android

### Testing

What testing have you done?
* Added one unit test for `fromUserInfo`, but unsure how else to unit test, please lmk if you have suggestions!
* Deployed and ran against https://github.com/mbta/mobile_app_backend/pull/460 with a well-formatted fallback and a poorly-formatted summary. Confirmed I received notifications for the well-formatted and didn't crash on the poorly-formatted.

imageimage


androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
```diff
@@ -22,11 +23,16 @@ class MBTAGoMessagingService : FirebaseMessagingService() {
...
+        try {
+            super.onMessageReceived(message)
+
+            val workRequest =
+                OneTimeWorkRequestBuilder<NotificationWorker>()
+                    .setInputData(PushNotificationPayload.messageToWorkData(message.data))
+                    .build()
+            WorkManager.Companion.getInstance(this).enqueue(workRequest)
+        } catch (t: Throwable) {
+            Sentry.captureMessage("Error enqueueing notification: $t")
```
KaylaBrady, line 35: I never actually saw any errors register in Sentry, so this may not do anything, but seems worth having just in case. 

I opted to use Sentry directly here rather than using Sentry or ISentryRepository directly. I thought there was a good reason at the time, but now I'm less sure... maybe it had to do with parity on iOS where we can't instantiate in the analogous class, and also using Sentry directly in `PushNotificationPayload.fromUserInfo`?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
```diff
@@ -50,6 +50,10 @@ public abstract class Analytics {
             }
     }
 
+    public fun notificationsFallback() {
```
KaylaBrady, line 53: since I wasn't having luck seeing anything in sentry, I figured I'd add this to analytics instead since it is a metric product is interested in. Assuming the other analytics added to the same classes are delivered as expected.

boringcactus (approved):

PR #1679, opened 2026-04-16
feat: correctly display multi-route trips on each route

boringcactus:
### Summary

_Ticket:_ [Ensure multi-route schedules and predictions are correctly displayed](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213913595242981?focus=true)

Consumes https://github.com/mbta/mobile_app_backend/pull/461 and preserves the multiple routes that a trip is scheduled/predicted on.

iOS

android

### Testing

Checked that a trip with an added route will show on both its route and its added route at its scheduled time. Added unit tests to ensure that objects that would previously have been lost as duplicates are now preserved.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -236,7 +236,7 @@ public data class RouteCardData(
                 val trip = formattedTrip.trip
                 if (context.isStopDetails() || trip.isUpcoming()) {
                     val existingPatterns =
-                        potentialService.getOrPut(Pair(trip.trip.routeId, trip.headsign)) {
+                        potentialService.getOrPut(Pair(trip.routeId, trip.headsign)) {
                             mutableSetOf()
                         }
                     if (trip.trip.routePatternId != null) {
@@ -366,8 +366,7 @@ public data class RouteCardData(
                     formattedTrips.map { format ->
                         val trip = format.trip
                         val route =
-                            if (shouldIncludeRoute) globalData?.getRoute(trip.trip.routeId)
-                            else null
+                            if (shouldIncludeRoute) globalData?.getRoute(trip.routeId) else null
                         LeafFormat.Branched.BranchRow(
                             route,
                             trip.headsign,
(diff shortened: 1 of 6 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -584,7 +582,7 @@ public data class RouteCardData(
     internal fun distanceFrom(position: Position): Length =
         this.stopData.first().stop.distanceFrom(position)
 
-    override fun toString(): String = "[RouteCardData]"
+    //    override fun toString(): String = "[RouteCardData]"
```
KaylaBrady: oops! I did this too recently. [Made a ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1214098708693408?focus=true) for a spotless rule to prevent this

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
```diff
@@ -35,6 +35,8 @@ constructor(
 
     val id: String = "${trip.id}-${prediction?.stopSequence ?: schedule?.stopSequence}"
 
+    public val routeId: Route.Id = prediction?.routeId ?: schedule?.routeId ?: trip.routeId
```
KaylaBrady, line 44: praise: Thanks for adding this, love the removals of all the `trip.routeId` calls! Maybe worth a comment about why this doesn't just use `trip.routeId` so a future reader will understand the multi-route trip implication

KaylaBrady (approved):
LGTM once the toString is back!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
boringcactus: oops!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
boringcactus: Good call.

PR #1681, opened 2026-04-17
feat: record current locale alongside notification subscriptions

boringcactus:
### Summary

_Ticket:_ [Store top supported language for notifications fallback](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214048781923336?focus=true)

Saves the current locale so that we can eventually use it for fallback text. Pairs with https://github.com/mbta/mobile_app_backend/pull/464.

iOS

android

### Testing

Updated unit tests. Manually verified that every language on every OS is correctly recorded.


EmmaSimon (approved):
Clever solution!

PR #1685, opened 2026-04-21
feat: Update trip summary behavior

EmmaSimon:
### Summary

_Ticket:_ [Notifications QA | Trip specific suspension and shuttle alert summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214072175090608?focus=true)
Backend PR: https://github.com/mbta/mobile_app_backend/pull/470

⚠️ Don't merge before the associated backend PR is ready

Fix a few edge cases with trip specific shuttle, suspension, and stop closure legibility, and added a new "this trip" identity.

<table>
  <tr>
    <td>Screenshot_20260421_151559</td>
    <td>Screenshot_20260421_151635</td>
  </tr>
</table>

iOS

android

### Testing

Added unit tests for new "this trip" identity, and the updated notification trip time identity. We didn't have tests for `FormattedAlert` directly on either platform and were using `AlertCardTests` to test them indirectly. I added some tests for the new text on alert cards, but also added some tests directly for `FormattedAlert`, just to simplify things. However, I just added tests for the new updated trip summary cases, I didn't go back and add tests for all the other existing behavior already covered by `AlertCardTests`.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/util/FormattedAlertTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/util/FormattedAlertTests.kt
@@ -1,0 +1,190 @@
+package com.mbta.tid.mbta_app.android.util
+
+import androidx.compose.ui.platform.LocalResources
+import androidx.compose.ui.test.junit4.createComposeRule
+import com.mbta.tid.mbta_app.model.Alert
+import com.mbta.tid.mbta_app.model.RouteType
+import com.mbta.tid.mbta_app.model.TripShuttleAlertSummary
+import com.mbta.tid.mbta_app.model.TripSpecificAlertSummary
+import com.mbta.tid.mbta_app.utils.EasternTimeInstant
+import kotlin.test.assertEquals
+import kotlin.time.Duration.Companion.hours
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.Month
+import org.junit.Rule
+import org.junit.Test
+
+class FormattedAlertTests {
+    @get:Rule val composeTestRule = createComposeRule()
+
+    @Test
+    fun testTripSpecificShuttle() = runTest {
+        val now = EasternTimeInstant(2026, Month.APRIL, 10, 15, 0, 0)
+        val summary =
+            TripShuttleAlertSummary(
+                TripShuttleAlertSummary.SingleTrip(
+                    now.plus(1.hours),
+                    RouteType.COMMUTER_RAIL,
+                    null,
+                ),
+                "Porter",
+                "North Station",
+            )
+        val format = FormattedAlert(null, summary)
+        val pattern = "Shuttle buses replace the 4:00\\sPM train from Porter to North Station"
+        composeTestRule.setContent {
+            val summaryString = format.alertCardMajorBody(LocalResources.current).toString()
+            assert(Regex(pattern).matches(summaryString))
+        }
+    }
+
+    @Test
+    fun testTripSpecificDownstreamShuttle() = runTest {
+        val now = EasternTimeInstant(2026, Month.APRIL, 10, 15, 0, 0)
+        val summary =
+            TripShuttleAlertSummary(
+                TripShuttleAlertSummary.SingleTrip(
+                    now.plus(1.hours),
+                    RouteType.COMMUTER_RAIL,
+                    "Concord",
+                ),
+                "Porter",
+                "North Station",
+            )
+        val format = FormattedAlert(null, summary)
+        val pattern =
+            "4:00\\sPM train from Concord is replaced by shuttle buses from Porter to North Station"
+        composeTestRule.setContent {
+            val summaryString = format.alertCardMajorBody(LocalResources.current).toString()
+            assert(Regex(pattern).matches(summaryString))
+        }
+    }
+
+    @Test
+    fun testThisTripSpecificShuttle() = runTest {
+        val summary =
+            TripShuttleAlertSummary(
+                TripShuttleAlertSummary.ThisTrip(RouteType.COMMUTER_RAIL),
+                "Porter",
+                "North Station",
+            )
+        val format = FormattedAlert(null, summary)
+        val expected = "Shuttle buses replace this train from Porter to North Station"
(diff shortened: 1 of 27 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/util/FormattedAlertTests.kt
```diff
@@ -0,0 +1,190 @@
...
+import com.mbta.tid.mbta_app.model.TripSpecificAlertSummary
+import com.mbta.tid.mbta_app.utils.EasternTimeInstant
+import kotlin.test.assertEquals
+import kotlin.time.Duration.Companion.hours
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.Month
+import org.junit.Rule
+import org.junit.Test
+
+class FormattedAlertTests {
```
KaylaBrady, line 17: Thank you for adding these tests!!

iosApp/iosApp/Utils/FormattedAlertTripShuttleExtension.swift
```diff
@@ -0,0 +1,87 @@
...
+            )
+        }
+    }
+
+    func tripShuttleAlertSummary(alertSummary: TripShuttleAlertSummary) -> AttributedString {
+        let identity = alertSummary.tripIdentity
+        let identityString = Self.summaryTripShuttleIdentity(tripIdentity: identity)
+
+        return if case let .singleTrip(singleTrip) = onEnum(of: identity), singleTrip.fromStop != nil {
+            AttributedString.tryMarkdown(String(
```
KaylaBrady, line 56: thought(non-blocking): I'm interested to see how these translations play out, with strings w/ args within strings w/ args

iosApp/iosApp/Utils/FormattedAlertTripSpecificExtension.swift
```diff
@@ -0,0 +1,139 @@
...
+//
+//  Created by esimon on 4/17/26.
+//  Copyright © 2026 MBTA. All rights reserved.
+//
+
+import Foundation
+import Shared
+
+extension FormattedAlert {
+    static func summaryTripIdentity(tripIdentity: TripSpecificAlertSummaryTripIdentity) -> String {
```
KaylaBrady, line 13: thought(non-blocking): These fns are pretty straightforward enum => string building, but seeing them duplicated does make me wish we could do it in shared...

iosApp/iosAppTests/Pages/AlertDetails/AlertDetailsTests.swift
```diff
@@ -48,8 +48,8 @@ final class AlertDetailsTests: XCTestCase {
         XCTAssertNotNil(try sut.inspect().find(text: stop2.name))
         XCTAssertNotNil(try sut.inspect().find(text: stop3.name))
         XCTAssertNotNil(try sut.inspect().find(text: "Full Description"))
-        XCTAssertNotNil(try sut.inspect().find(text: alert.description_!))
-        XCTAssertNotNil(try sut.inspect().find(text: alert.header!))
+        XCTAssertNotNil(try sut.inspect().find(text: XCTUnwrap(alert.description_)))
+        XCTAssertNotNil(try sut.inspect().find(text: XCTUnwrap(alert.header)))
```
KaylaBrady, line 52: TIL XCTUnwrap

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/AlertSummary.kt
```diff
@@ -321,9 +336,14 @@ public sealed class AlertSummary {
             // same disrupted stops, or if multiple branches are disrupted
             val firstStops = affectedPatternStops.values.firstOrNull { it.size > 1 } ?: return null
             val orderedStops = firstStops.mapNotNull { global.stops[it] }
+            val downstream = affectedPatternStops.values.all { !it.contains(stopId) }
```
KaylaBrady, line 339: thought(non-blocking): I also added a boolean representation of downstream in [DisplayAlert](https://github.com/mbta/mobile_app/blob/1ef97809b8a20e420272aa2d3d21caea6138304a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/DisplayAlert.kt#L13). This makes me wonder if changing our representation of Alerts at the RouteCardData to have more contextual information baked in would make it easier to work with alerts across both places... definitely a bigger refactor than we'd want to consider now, but something we may want to keep an eye on

KaylaBrady (approved):
Nice! A few non-blocking stray thoughts, but no concerns with this approach!

iosApp/iosApp/Utils/FormattedAlertTripShuttleExtension.swift
EmmaSimon: Yeaaaaah, I'm a bit concerned too, but all the alternatives seem even more unwieldy.

iosApp/iosAppTests/Pages/AlertDetails/AlertDetailsTests.swift
EmmaSimon: I was having issues with SwiftFormat for Xcode freezing up constantly, so I switched to the now built into xcode swift-format, which did this automatically 😅 It also disagrees a in a few cases with SwiftFormat in the commit hooks (and does some more forceful but compatible formatting in some cases), which is slightly annoying, but not completely unworkable.

iosApp/iosApp/Utils/FormattedAlertTripShuttleExtension.swift
KaylaBrady: Totally. The one alternative that I think would be viable if we need it is to just use the alert header, but only if we find that it is translated in a meaningfully better way.

EmmaSimon: Yeah, I don't think we'll really know if that's generally true for a pretty long time. It would be the most reliable to have a separate string for each possible permutation, but having the number of strings be exponential based on the params is scary, even if it would be kinda manageable in practice (for now).

PR #1686, opened 2026-04-21
fix: check routeType when checking alert informed entities

KaylaBrady:
### Summary

_Ticket:_ [Fix: mode-wide alert appeared on other modes, too](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1212966333558790?focus=true)

What is this PR for?
We currently only compare alert informed entities by route ID and don't check routeType. We [work around that on the backend](https://github.com/mbta/mobile_app_backend/blob/721639f298451e67fb66355fc6c1110ddc352ea4/lib/mbta_v3_api/alert.ex#L243) by expanding mode-specific alerts with a routeType specified into informed entities for each route. 

We still wanted to represent the mode-specific check on the frontend though for consistency & correctness. 

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally against https://github.com/mbta/mobile_app_backend/pull/467, which removed the exploding logic. This pointed me to the changes needed for `AlertAssociatedStop` b/c all stops on the map were marked as closed when I tested with a ferry alert!

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -348,9 +348,9 @@ internal constructor(
 
             fun checkRoute(routeId: Route.Id?, routeType: RouteType?) {
                 if (!isSatisfied) return
+                checkRouteType(routeType)
                 if (routeId == null) return
                 checkRouteIdIn(listOf(routeId))
-                checkRouteType(routeType)
             }
 
             fun checkRouteIdIn(routeIds: Collection<Route.Id>) {
@@ -363,7 +363,9 @@ internal constructor(
 
             fun checkRouteType(routeType: RouteType?) {
                 if (!isSatisfied) return
-                if (routeType == null) return
+                if (routeType == null && this@InformedEntity.routeType != null) {
+                    isSatisfied = false
+                }
                 if (this@InformedEntity.routeType == null) return
                 if (this@InformedEntity.routeType != routeType) {
                     isSatisfied = false
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -363,8 +363,7 @@ internal constructor(
 
             fun checkRouteType(routeType: RouteType?) {
                 if (!isSatisfied) return
-                if (routeType == null && this@InformedEntity.routeType != null) return
-
+                if (routeType == null) return
                 if (this@InformedEntity.routeType == null) return
                 if (this@InformedEntity.routeType != routeType) {
                     isSatisfied = false
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -343,25 +343,25 @@ internal constructor(
             }
 
             fun checkRoute(route: Route?) {
-                checkRoute(route?.id)
-                checkRouteType(route?.type)
+                checkRoute(route?.id, route?.type)
             }
 
-            fun checkRoute(routeId: Route.Id?) {
+            fun checkRoute(routeId: Route.Id?, routeType: RouteType?) {
                 if (!isSatisfied) return
                 if (routeId == null) return
-                checkRouteIn(listOf(routeId))
+                checkRouteIn(listOf(routeId), routeType)
             }
 
-            fun checkRouteIn(routeIds: Collection<Route.Id>) {
+            fun checkRouteIn(routeIds: Collection<Route.Id>, routeType: RouteType?) {
                 if (!isSatisfied) return
                 if (this@InformedEntity.route == null) return
                 if (this@InformedEntity.route !in routeIds) {
                     isSatisfied = false
                 }
+                checkRouteType(routeType)
             }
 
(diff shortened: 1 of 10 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/AlertSummary.kt
```diff
@@ -406,7 +406,7 @@ public sealed class AlertSummary {
                                         // `affectedStops` only includes parent stops, here we check
                                         // if the child stops on each pattern are affected
                                         checkStop(stopOnTrip)
-                                        checkRoute(pattern.routeId)
+                                        checkRoute(pattern.routeId, routes.firstOrNull()?.type)
```
KaylaBrady, line 409: Logically I know it would be safe to do checkRoute(routes.firstOrNull()) here.... but it feels strange to do that instead of using the routeId from the pattern on the off-chance that something goes wrong.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -342,19 +342,39 @@ internal constructor(
...
+                if (this@InformedEntity.route == null && this@InformedEntity.routeType == null)
+                    return
+                if (this@InformedEntity.route != null && this@InformedEntity.route != routeId) {
+                    isSatisfied = false
+                }
+                if (
+                    this@InformedEntity.routeType != null &&
+                        routeType != null &&
+                        this@InformedEntity.routeType != routeType
+                ) {
```
EmmaSimon: It might be a little more straightforward to split this into `checkRouteId` and `checkMode`/`checkRouteType`, then keep `checkRoute` and `checkRouteIn` with the same signatures, but use the separate checks within them. I'm also wondering if there's a reason not to have a single `checkRouteId` (maybe `checkRouteIdIn`?) implementation that takes a collection like `checkRouteIn`, then in `checkRoute(RouteId)`, call that with the single ID in a list. It just seems a bit redundant to have a fully duplicated method with the only difference being `!=` vs `!in`

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -342,13 +342,15 @@ internal constructor(
                 }
             }
 
+            fun checkRoute(route: Route?) {
+                checkRoute(route?.id)
+                checkRouteType(route?.type)
+            }
+
             fun checkRoute(routeId: Route.Id?) {
```
EmmaSimon: Sorry I think my original comment was too vague and confusing, if we have two different `checkRoute` methods with different signatures, they should do the same thing fundamentally, I'd be more comfortable with this being:

```
fun checkRoute(route: Route?) {
  checkRoute(route?.id, route?.type)
}

fun checkRoute(routeId: Route.Id?, routeType: RouteType?) {
  checkRouteId(routeId)
  checkRouteType(routeType)
}

fun checkRouteId(routeId: Route.Id?) {
  if (!isSatisfied) return
  if (routeId == null) return
  checkRouteIn(listOf(routeId))
}
```

KaylaBrady: Ah I see, thanks! I ended up doing away with `checkRouteId` - if dealing with a single route, I think using either of the `checkRoute` options makes it more likely that we don't forget to separately checkRouteType.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -342,19 +342,30 @@ internal constructor(
...
             }
 
-            fun checkRoute(routeId: Route.Id?) {
+            fun checkRoute(route: Route?) {
+                checkRoute(route?.id, route?.type)
+            }
+
+            fun checkRoute(routeId: Route.Id?, routeType: RouteType?) {
                 if (!isSatisfied) return
                 if (routeId == null) return
```
EmmaSimon, line 352: I think you need to put `checkRouteType` above this return in case a nil route ID is passed with a non-nil route type. I didn't realize it at first either, but that was why I split out the separate `checkRouteId` method 😅 But it should be the same effect to just move the route type check above.

KaylaBrady: Ah TY, I should test these more directly too.

KaylaBrady: I struggled with this more than I thought I would. 
For an alert that has routeType specified, I think these are the right assertions, because elsewhere null values passed to a `checkX` return true. But my brain just wants these to be false. 
```
 assertTrue(alert.anyInformedEntitySatisfies { checkRoute(Route.Id("1"), null) })
 assertTrue(alert.anyInformedEntitySatisfies { checkRoute(null, null) })
```

EmmaSimon: Yeah I think it makes sense for these to be true, null passed in is basically a wildcard, so it should match any value that the informed entity contains. I'm a little mixed up too though, the value starting as true then being proved false makes it hard to wrap your head around. Why is the `&& this@InformedEntity.routeType != null` necessary? If both values are null, wouldn't it just return on the next line anyway?

KaylaBrady: True, it is not necessary!

EmmaSimon (approved):
Thanks for working through all of that!!

PR #1688, opened 2026-04-21
fix: make read-only copies of ObjectCollectionBuilder object maps in models

boringcactus:
### Summary

_Ticket:_ [iOS: Fix flaky shared tests](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213997621491057)

The test that was failing sometimes was creating new objects while a background task was iterating over existing objects, so the ConcurrentModificationException was real, although it would only be possible in tests. This could have been fixed by rearranging the test, but then a later test would accidentally do the same thing and create a similar race condition; the more robust fix is to make immutable copies when creating (for example) an `AlertsStreamDataResponse` out of an `ObjectCollectionBuilder` so that no test can exhibit a race condition like this again. (I wish I were writing Rust, where accidentally cascading mutability like this is really difficult to do.)

iOS

android

### Testing

Ran the iOS shared tests quite a few times locally and did not see any failures.


KaylaBrady (approved):
Sneaky!

PR #1689, opened 2026-04-22
fix(Alert.recurrenceRange): count all days in active period

KaylaBrady:
### Summary

_Ticket:_ [GL-B Alert recurrence display issue](https://github.com/mbta/mobile_app/pull/new/kb-recurrence-fix)

What is this PR for?
We were incorrectly displaying a GL-B alert as ending on 4/24, because that is when the first active period ends. In reality, it is recurring weekdays over 2 disjoint periods. I think part of the issue was that [The backend](https://github.com/mbta/mobile_app_backend/blob/721639f298451e67fb66355fc6c1110ddc352ea4/lib/mbta_v3_api/alert.ex#L233) collapses consecutive periods together, whereas the frontend was expecting separate periods for each day. I'm concerned that there could be other ramifications of removing the backend collapsing logic (introduced [here](https://github.com/mbta/mobile_app_backend/pull/328) to address a different active period display issue), so this just makes the recurrence check handle cases where a period is > 1 day.

Backend PR: https://github.com/mbta/mobile_app_backend/pull/469

Note: [outstanding discussion](https://mbta.slack.com/archives/C082U4THN5A/p1776886055791379?thread_ts=1776858827.168749&cid=C082U4THN5A) as to whether we can get the data representation for a recurrence directly from Alerts
iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally for alert w/ same dates as the affected GL-B alert & confirmed it displays as a recurrence.
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -447,8 +447,12 @@ internal constructor(
         val days: Set<DayOfWeek>,
         val endDayKnown: Boolean,
     ) {
+        /**
+         * If the alert is for > 1 day and it is active for all days in the range, treat it as daily
+         */
         public val daily: Boolean =
-            days == (start.serviceDate..end.serviceDate).map { it.dayOfWeek }.toSet()
+            days == (start.serviceDate..end.serviceDate).map { it.dayOfWeek }.toSet() &&
+                days.size > 1
 
         public val fromStartOfService: Boolean = start.local.time == LocalTime(3, 0)
         public val toEndOfService: Boolean =
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -3,7 +3,6 @@ package com.mbta.tid.mbta_app.model
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.utils.EasternTimeInstant
 import kotlin.time.Duration.Companion.hours
-import kotlinx.datetime.DatePeriod
 import kotlinx.datetime.DayOfWeek
 import kotlinx.datetime.LocalDate
 import kotlinx.datetime.LocalTime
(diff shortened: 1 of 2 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -460,32 +459,22 @@ internal constructor(
...
+                        DayOfWeek.entries.toSet()
+                    } else {
+                        (it.startServiceDate..it.endServiceDate).map { it.dayOfWeek }.toSet()
+                    }
                 }
-            }
-        val allAlertDaysContiguous =
-            alertDays
-                .windowed(size = 2, step = 1) { (a, b) -> (b - a) == DatePeriod(days = 1) }
-                .all { it }
```
KaylaBrady: Would especially like a close eye on this removal. The previous logic would always return all days of the week if the alert days were continuous. I don't quite understand that special case; if all days are continuous, but only 3 days, why would we want to include the full set of days?

This is also the source of the +1s in the existing tests - there were only 6 days covered by those active periods, but they were expecting to see all 7 days listed in `RecurrenceInfo`

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -460,32 +459,22 @@ internal constructor(
...
-                    it.startServiceDate
-                } else {
-                    return@recurrenceRange null
+        val daysOfWeek =
+            activePeriod
+                .flatMap {
+                    if (it.endServiceDate == null) {
+                        DayOfWeek.entries.toSet()
+                    } else {
+                        (it.startServiceDate..it.endServiceDate).map { it.dayOfWeek }.toSet()
```
KaylaBrady, line 474: Note: This does bake in some assumptions about the limitations in place when creating a recurring alert: you select the days of the week, and then the date range. So one alert wouldn't have one active period that is weekdays, then separate active periods for the next 4 saturdays.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -460,32 +459,37 @@ internal constructor(
...
+            lastPeriod.endServiceDate != null &&
+                (firstPeriod.startServiceDate..lastPeriod.endServiceDate)
+                    .map { it.dayOfWeek }
+                    .toSet() == seenDaysOfWeek
+
+        // If all the days in the range have been seen, then include all days.
+        // This indicates that the alert is "daily".
+        val daysOfWeek =
+            if (allDaysInRangeSeen && seenDaysOfWeek.size > 1) {
                 DayOfWeek.entries.toSet()
```
KaylaBrady, line 490: Including all the days of the week if seenDaysOfWeek are continuous makes me kind of uneasy, but it was the existing behavior. I tried instead to have this return only the actually included days and modify `RecurrenceInfo.daily`, but I backed it out to minimize the change. Please lmk if you think that is worth adding back in!

EmmaSimon: If you were happier with that approach when you had it implemented, I think it makes sense to have it as a follow up. But it also seems like it could get a bit clunky to conditionally change `RecurrenceInfo.daily`, how did you have it working?

EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
KaylaBrady: I think it was ultimately something like 
```
 public val daily: Boolean =
            days == (start.serviceDate..end.serviceDate).map { it.dayOfWeek }.toSet() && days.count > 1
```
I do like it, but it is on the clunky side

EmmaSimon: That's not too bad, I'd be in favor, doesn't it also mean that we can get rid of the extra check in `recurrenceRange()`? This seems way more concise than that.

KaylaBrady: I'll follow up with that then, thanks!

PR #1697, opened 2026-04-29
feat: Consolidate and standardize error banner data errors

EmmaSimon:
### Summary

_Ticket:_ [Standardize data errors across the app](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213823872544150?focus=true)

This moves all socket related error banner setting and clearing logic into `ChannelOwner`. This means that since channel messages need to be parsed before we're sure if there's an error or not, the channel owner needs to be generic to the message type, so that it can handle parsing internally. This is complicated a bit further since our predictions channel parses a different type when joining and when receiving subsequent messages. I created an `AsymmetricChannelOwner` which takes two types, and implemented `ChannelOwner` internally as one of these but with both types being the same.

I also updated the `ApiResult.Error` messages for socket errors to include additional data about the message that the channel failed on, which should show up in the frequent error banner sentry logs and will help us debug socket connection issues.

Other than the socket changes, I also updated the `MapViewModel` rail route shape static fetch to use the standardized `fetchApi` helper, which all other repo requests already use, which handles error banner setting and clearing for static requests.

Also deleted a couple long unused state helpers.

iOS

android

### Testing

Updated existing unit tests to pass. Also ran on both platforms, manually caused connection errors by turning off the internet, and confirmed that the expected banner keys were set and cleared properly.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/ChannelOwner.kt
```diff
@@ -1,49 +1,134 @@
...
+        spec: ChannelSpec,
+        parseMessage: (String) -> X,
+        handleResult: (ApiResult<X>) -> Unit,
+        errorKey: String,
+    ) = owner.connect(spec, parseMessage, parseMessage, handleResult, handleResult, errorKey)
+
+    fun disconnect() = owner.disconnect()
+}
+
+internal class AsymmetricChannelOwner<X : Any, Y : Any>(
```
boringcactus: These might benefit from more descriptive names, like `JoinData` and `MessageData`.

boringcactus (approved):

PR #1698, opened 2026-04-30
fix: station/stop bypass/closure language

BrandonTR:
### Summary

_Ticket:_ [Notifications QA | update language for station bypass/closure alerts](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213994843939454?focus=true)

Update stop/station closure/bypass to consistently be stop skipped. Also updates summary to be `[trains/buses] will not stop at [affected stop(s)] [timeframe]` instead of "station closed at...."

iOS

android

### Testing

Added tests for new logic, fixed existing tests


iosApp/iosApp/Utils/FormattedAlert.swift
```diff
@@ -360,13 +407,31 @@ struct FormattedAlert: Equatable {
...
+                let skippedArgs = [
+                    Self.summaryAffectedMode(effect: alertSummary.effect),
+                    Self.summarySkippedEffect(
+                        stops: location,
+                        timeframe: String(timeframe.trimmingPrefix(.horizontalWhitespace))
+                    ),
+                ]
+                return AttributedString.tryMarkdown(String(format:
+                    NSLocalizedString(
+                        "%1$@%2$@",
```
EmmaSimon: Could you include the space after the mode here, rather than appending it individually wherever the modes are set?

iosApp/iosApp/Utils/FormattedAlert.swift
```diff
@@ -360,13 +407,31 @@ struct FormattedAlert: Equatable {
...
                 sentenceCaseEffect,
-                Self.summaryLocation(effect: alertSummary.effect, location: alertSummary.location),
-                Self.summaryTimeframe(timeframe: alertSummary.timeframe),
+                location,
+                timeframe,
                 Self.summaryRecurrence(recurrence: alertSummary.recurrence),
             ]
-            if alertSummary.isUpdate {
+            if alertSummary.effect.stopSkipped {
+                let skippedArgs = [
```
EmmaSimon: Why set the args in a list rather than just put them as params to the string format? We're only doing it for the other args because the same ones are used in two places. It might even be worth it to just move the stop skipped if statement and return above the setting of `args` entirely, so we're not unnecessarily generating the recurrence, but that's a nitpick.

iosApp/iosApp/Utils/FormattedAlert.swift
```diff
@@ -344,6 +347,50 @@ struct FormattedAlert: Equatable {
...
+                "will not stop at %@ %@",
+                comment: "Alert effect denoting station bypass, ex “will not stop at [Back Bay and Ruggles] [today]”"
+            ),
+            stops,
+            timeframe
+        )
+    }
+
+    static func summaryAffectedStops(stops: [String]) -> String {
+        stops.map { "**\($0)**" }.reduce(nil) { lhs, rhs in
```
EmmaSimon: non-blocking: We should maybe check with design if we're expected to support 3+ stops

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -7362,6 +7409,53 @@
         }
       }
     },
+    "Buses" : {
+      "comment" : "Mode of travel prefixing alert summary e.g.: \"Buses will not stop at...\"",
+      "localizations" : {
+        "es" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "autobuses"
```
EmmaSimon: Capitalize

iosApp/iosApp/Localizable.xcstrings
```diff
@@ -25151,6 +25296,53 @@
         }
       }
     },
+    "Trains" : {
+      "comment" : "Mode of travel prefixing alert summary e.g.: \"Trains will not stop at...\"",
+      "localizations" : {
+        "es" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "trenes"
```
EmmaSimon: Capitalize

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/AlertSummaryTest.kt
```diff
@@ -20,18 +20,20 @@ import kotlinx.datetime.atTime
...
 import kotlinx.serialization.json.JsonObjectBuilder
+import kotlinx.serialization.json.add
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.decodeFromJsonElement
 import kotlinx.serialization.json.encodeToJsonElement
 import kotlinx.serialization.json.put
+import kotlinx.serialization.json.putJsonArray
 import kotlinx.serialization.json.putJsonObject
 
 class AlertSummaryTest {
```
EmmaSimon, line 31: Can you add a new test to check that a closure alert properly sets an `AffectedStops` location?

iosApp/iosApp/Utils/FormattedAlert.swift
BrandonTR: I mostly just did it to be consistent with the rest but I agree with your suggestions!

iosApp/iosApp/Utils/FormattedAlert.swift
BrandonTR: Checking with Adam

BrandonTR: He suggested we limit to 3 stops and if there's more than that we just say "multiple stops". Going ahead and adding this

EmmaSimon (approved):
Looks good, pending the multiple stops changes! Feel free to put up a follow up PR for that if that's easier.

iosApp/iosApp/Utils/FormattedAlert.swift
```diff
@@ -359,16 +359,23 @@ struct FormattedAlert: Equatable {
...
-            if let lhs { String(
-                format: NSLocalizedString(
-                    "%1$@ and %2$@",
-                    comment: "Joins two stops into a list, ex “[Back Bay] and [Ruggles]”"
-                ),
-                lhs,
-                rhs
-            ) } else { rhs }
-        } ?? ""
+        if stops.count > 3 {
```
EmmaSimon, line 362: Was it meant to be > 3 or >= 3? If we're supporting 3, I think the concat logic would produce "X and Y and Z", rather than like, "X, Y, and Z", which is a little awkward.

BrandonTR: Fixed the formatting! Apparently iOS and Android both have the same utility to achieve this which is nice!

EmmaSimon (approved):

iosApp/iosApp/Utils/FormattedAlert.swift
EmmaSimon: That's sick, nice find!

PR #1707, opened 2026-05-05
fix: Address branching stop route card data issues with alerts and atypical routes with service

EmmaSimon:
### Summary

_Ticket:_ [[Red Line at JFK]](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214494522221793?focus=true)

Thanks to @boringcactus for the fix 🙏 

This was happening when a branching stop had a disruption on one branch and only atypical schedules on the other.

### Testing

Added unit test for this case


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -228,12 +228,14 @@ public data class RouteCardData(
                 mutableMapOf()
             val isBus = representativeRoute.type == RouteType.BUS
             val tripsToConsider =
-                if (isBus && context != Context.StopDetailsFiltered)
-                    tripsUpcoming.take(TYPICAL_LEAF_ROWS)
-                else tripsUpcoming
+                tripsUpcoming
+                    .run { if (any { it.second != null }) filter { it.second != null } else this }
```
KaylaBrady: suggestion: this seems a little hard to read to me. Maybe worth a comment as to why we are doing this?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -523,8 +529,9 @@ public data class RouteCardData(
                     else -> TYPICAL_LEAF_ROWS
                 }
             val tripsToShow =
-                if (countTripsToDisplay != null) allTripsToShow.take(countTripsToDisplay)
-                else allTripsToShow
+                allTripsToShow
+                    .mapNotNull { it.second }
```
KaylaBrady: suggestion: allTripsToShow seems like a misnomer now that we have to filter at the end, maybe worth renaming to something like `allTripsWithFormat

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
```diff
@@ -324,37 +324,43 @@ internal fun List<UpcomingTrip>.isArrivalOnly(): Boolean {
...
 
     // If the last non-cancelled trip is not flagged as the last trip, we need to set the last
     // trip flag on the cancelled trip to false, and set it to true on the actual last trip
-    return formattedTrips.mapNotNull {
+    return formattedTrips.map {
         if (it == lastNonCancelledTrip) {
-            return@mapNotNull it.trip.format(now, route, context, true)
-        } else if (it.lastTrip) {
-            return@mapNotNull it.trip.format(now, route, context, false)
+            it.copy(second = it.first.format(now, route, context, true))
```
KaylaBrady: suggestion(non-blocking): May be worth destructuring the first / second here for readability

KaylaBrady (approved):
Awesome! Just a few minor readability suggestions

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
boringcactus: Good idea.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
boringcactus: Sure.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt
boringcactus: Might even be worth using a bespoke data class so we can have clearer names than just `first` and `second`.

PR #1709, opened 2026-05-05
feat: move notification deep link calculation to the backend

boringcactus:
### Summary

_Ticket:_ [[fix notification deep linking and icon and stuff]](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214547310290612)

iOS

android

### Testing

Checked that things work, except for the things that don’t.


androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
```diff
@@ -28,18 +22,16 @@ class MainActivity : ComponentActivity() {
...
-            AnalyticsProvider(Firebase.analytics)
-                .notificationClicked(pushNotificationPayload, stillActive)
-            deepLinkStateFlow.value = pushNotificationPayload.getDeepLinkState()
-        } else {
-            val deepLinkUri: Uri? = intent?.data?.takeIf { intent.action == Intent.ACTION_VIEW }
-            deepLinkStateFlow.value = deepLinkUri?.let { DeepLinkState.from(it.toString()) }
-        }
+        val deepLinkPath =
+            if (intent?.hasExtra("deep_link_path") == true) {
+                // TODO notification sound
```
boringcactus: Android notifications don’t play any sound for me, and that might be something specific to my phone (I did check that it’s not on vibrate) but it might be something we actually need to fix.

iosApp/iosApp/IOSApp.swift
```diff
@@ -127,11 +75,9 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
...
         Messaging.messaging().appDidReceiveMessage(userInfo)
-        if let rawPayload = userInfo[PushNotificationPayload.companion.launchKey] as? String {
-            let payload = PushNotificationPayload.companion.deserialize(rawPayload: rawPayload)
-            let stillActive = payload.isStillActive()
-            AnalyticsProvider.shared.notificationClicked(payload: payload, stillActive: stillActive)
-            Self.notificationDeepLinkOwner.notificationDeepLink = payload.getDeepLinkState()
+        // TODO: if you open the notification when the app is not active in the background, this doesn’t work
+        if let deepLinkPath = userInfo["deep_link_path"] as? String {
+            Self.notificationDeepLinkOwner.notificationDeepLink = .companion.from(url: deepLinkPath)
         }
```
boringcactus: If I open a notification when the app is in the foreground or when I just sent the app to the background, I get the behavior I expect, but if I force close the app from the task switcher and then I open a notification, it just opens the app without opening the deep link.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
KaylaBrady: Confirmed I did hear notification sound

BrandonTR: I think we can probably remove this comment then

BrandonTR (approved):
One nitpick otherwise lgtm!

androidApp/src/main/java/com/mbta/tid/mbta_app/android/MainActivity.kt
```diff
@@ -28,18 +22,16 @@ class MainActivity : ComponentActivity() {
...
+                intent.getStringExtra("deep_link_path")
```
EmmaSimon: Maybe put the `"deep_link_path"` string into a constant

EmmaSimon (commented):
Are we okay removing the analytics events? I thought Melody mentioned that we might be able to preserve some of that, but it seems like it might not be possible to record clicks on Android when opening from background.

PR #1714, opened 2026-05-07
feat: use trip_headsign field

BrandonTR:
### Summary

_Ticket:_ [Use new trip_headsign field](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214110029669184?focus=true)

Adds changes to utilize the new `trip_headsign` field. I think this is all that's needed?

### Testing

Added a test for new headsign override in UpcomingTest.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -431,7 +427,9 @@ public data class RouteCardData(
                         if (shouldIncludeRoute) globalData?.getRoute(upcomingTrip.routeId) else null
                     LeafFormat.Branched.BranchRow(
                         route,
-                        upcomingTrip.trip.headsign,
+                        upcomingTrip.trip
+                            .headsign, // Why is this getting headsign directly from the trip?
+                        // Should it just be UpcomingTrip.headsign now?
```
BrandonTR: ^

KaylaBrady: Great catch! Yes, I think this is the right change. 

AFAIK schedule.stopHeadsign is only really used on CR. It may be worth doing a spot check for a branched CR that uses stop headsigns (trying to look at Franklin/Foxboro Line at Forest Hills for reference, since I have vague memories of stop headsigns being used at Forest Hills, and am seeing maybe a separate issue. Will follow up in slack)

BrandonTR: Did a spot check and it does seem to be using the correct headsign in this scenario! Stop headsigns didn't exist for everything at this stop for CR but it did for the Providence & Forge Park trip

KaylaBrady: Sweet, thanks for checking!

KaylaBrady (approved):
Oops sorry, thought I approved this on first review!

PR #1715, opened 2026-05-08
fix(a11y): More readable buttons when show button borders is on

KaylaBrady:
### Summary

_Ticket:_ [Accessibility Bug: Readability with Button Shape setting](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213958079429272?focus=true)

What is this PR for?

When the "Show Borders" accessibility setting is on, particularly with "High Contrast" also on, many buttons in our app were unreadable when the sheet was lowered. 

This PR improves the readability of key interactive elements when that those settings are on. 
I may have gotten carried away with this in a way that is no longer in the spirit of the settings, so would definitely appreciate a second opinion on whether this changes too much. 

iOS

android

### Testing

| Before | After |
|--------|--------|
|image|image|
|image| image|
| image | image |

What testing have you done?
* Ran locally with the settings enabled, confirmed that the direction picker, departure tiles, and route cards are all more readable. 

* I also modified the trip stops & search results, but maybe I should leave those as-is since they are maybe less obviously interactive? I left the alert card as-is as an example of what these would look like with the setting on, which maybe isn't so bad. Poking around a few other apps, I do not see these button border artifacts though when the setting is on, so I'm not sure how important the shading is to preserve.


iosApp/iosApp.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -33,26 +33,26 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/firebase/firebase-ios-sdk",
```
KaylaBrady, line 34: oops! Will remove

iosApp/iosApp/Pages/StopDetails/DepartureTile.swift
```diff
@@ -81,11 +81,15 @@ struct DepartureTile: View {
...
+            .padding(.horizontal, 10)
+            .padding(.vertical, 10)
+            .frame(maxWidth: maxTileWidth, minHeight: 56, maxHeight: .infinity)
+            .withProminentButtonPaddingRemoved()
         }
-        .padding(.horizontal, 10)
-        .padding(.vertical, 10)
-        .frame(maxWidth: maxTileWidth, minHeight: 56, maxHeight: .infinity)
-        .background(isSelected ? Color.fill3 : Color.routeColorContrast)
+        // using borderedProminent for readable layout with "Show Borders" setting.
```
EmmaSimon: Does this mean that it's considered borderedProminent all the time, so when the setting is on, it doesn't change it at all, or that only when the setting is turned on, it's changed in a way that makes more sense?

iosApp/iosApp/Utils/Modifiers/ProminentButtonOffsetPadding.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Copyright © 2026 MBTA. All rights reserved.
+//
+
+import SwiftUI
+
+struct RemoveProminentButtonPadding: ViewModifier {
+    func body(content: Content) -> some View {
+        content
+            .padding(.horizontal, -10)
+            .padding(.vertical, -8)
```
EmmaSimon: I assume there must not be a built in way to determine what padding is added? I'm a little cautious in case iOS decides to change these values in some future release.

iosApp/iosApp/ContentView.swift
```diff
@@ -554,6 +554,9 @@ struct ContentView: View {
                     VStack {
                         navSheetContents
                             .presentationDetents([.small, .medium, .almostFull], selection: $selectedDetent)
+                            // Setting background color helps preserev contrast with accessibility setting "Show
```
EmmaSimon: Typo in "preserve"

EmmaSimon (commented):
I don't have any issues with this approach, I think not messing up the contrast and layout is more important than applying whatever styling iOS thinks might be helpful, and most of our buttons are already oversized and don't need help becoming larger tap targets.

iosApp/iosApp/Utils/Modifiers/ProminentButtonOffsetPadding.swift
KaylaBrady: Oh very good point. I was reaching for `borderedProminent` as kind of a last resort since nothing else worked. I had tried just applying `tint` instead of `background` in DepartureTile & DirectionPicker before, but now I'm realizing that `tint` works as part of these other changes when applied in addition to background. So I'll just do that in these two places too and remove this!

It still feels wacky to need tint & background both, so there is probably a more conventional fix out there, but that will at least be a less-hacky fix than this one.

EmmaSimon: Oh nice, we can delete this modifier then, right? I'm not too bothered by needing to set both. Could we even set the tint to something reasonable that does differentiate the buttons without ruining contrast when the setting is turned on? I'm assuming that the tint would only show up when the setting is enabled.

KaylaBrady: Oh ty, I deleted it locally but must have messed something up with the project generation. 

I'm hesitant to change tint to anything else since the text color can be different based on the route selected, and I'm not really sure if the tinted capsule area is something that is actually desirable here. I haven't found another app that adds capsules when the setting is toggled on for elements besides little buttons like our close & back. Maybe most comparably, notes & contacts as two apps with lists of tappable things do not where the items in that list don't get marked with extra capsules.

EmmaSimon: Yeah, that makes sense! I think part of it is just that we're using buttons for things that SwiftUI assumes you wouldn't use buttons for.

EmmaSimon (approved):

PR #1717, opened 2026-05-08
fix: clear stale favorites

boringcactus:
### Summary

_Ticket:_ [Sentry Bug | Empty static RouteCardData](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213986119380531)

This Sentry message was never a bug; it was added to indicate favorites that are stale and will probably be cleared, but that doesn’t even happen very often, and the corresponding Sentry message to indicate that favorites were actually cleared fell through the cracks of our triage process (and also appears to have never fired at all on iOS).

The mystery of the disappearing favorites turns out to have been a bug hiding behind a bug, dodging tests with a third bug. Fundamentally, the logic for checking if the favorited direction was missing was backwards, so directions that did exist were the ones selected for deletion. If this happened every time, I think we would have caught it fairly quickly, but we also weren’t correctly waiting for the favorites to load before trying to clear favorites, so it was rare on Android and impossible on iOS for anything to actually happen at all. We should have caught this in tests, but our tests were only looking for a change that was created explicitly in the mock, so the actual functionality wasn’t being tested at all. All three of those are fixed.

iOS

android

### Testing

Checked with some temporary logging that if I create a stale favorite, it gets cleared, on both Android and iOS. Updated unit tests.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpdateSubscriptionsRequests.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpdateSubscriptionsRequests.kt
@@ -10,7 +10,7 @@ public data class WriteSubscriptionsRequest
 internal constructor(
     @SerialName("fcm_token") val fcmToken: String,
     val subscriptions: List<SubscriptionRequest>,
-    val locale: String,
+    val locale: String?,
 )
 
 @Serializable
@@ -18,7 +18,7 @@ public data class UpdateAccessibilityRequest
 internal constructor(
     @SerialName("fcm_token") val fcmToken: String,
     @SerialName("include_accessibility") val includeAccessibility: Boolean,
-    val locale: String,
+    val locale: String?,
 )
 
 @Serializable
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
@@ -69,7 +69,7 @@ public interface IFavoritesViewModel {
         defaultDirection: Int,
         fcmToken: String?,
         includeAccessibility: Boolean,
-        locale: String,
+        locale: String?,
     )
 }
 
@@ -103,7 +103,7 @@ public class FavoritesViewModel(
             val defaultDirection: Int,
             val fcmToken: String?,
             val includeAccessibility: Boolean,
-            val locale: String,
+            val locale: String?,
         ) : Event
     }
 
@@ -134,6 +134,7 @@ public class FavoritesViewModel(
             mutableStateOf(null)
         }
 
+        var fcmTokenForClearingStaleFavorites: String? by remember { mutableStateOf(null) }
         var hadOldPinnedRoutes: Boolean by remember { mutableStateOf(false) }
         var shouldShowFirstTimeToast: Boolean by remember { mutableStateOf(false) }
         var shouldShowNotificationsHint: Boolean by remember { mutableStateOf(false) }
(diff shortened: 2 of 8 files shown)


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModelTest.kt
```diff
@@ -930,13 +928,7 @@ internal class FavoritesViewModelTest : KoinTest {
...
         val favoritesAfter = buildFavorites { routeStopDirection(route1.id, stop1.id, 0) }
 
-        val favoritesRepo = mock<IFavoritesRepository>(MockMode.autofill)
-
-        everySuspend { favoritesRepo.getFavorites() } sequentially
-            {
-                returns(favoritesBefore)
-                repeat { returns(favoritesAfter) }
-            }
+        val favoritesRepo = MockFavoritesRepository(favorites = favoritesBefore)
```
boringcactus, line 931: The `MockFavoritesRepository` will change what it returns if `setFavorites` is called, so it already behaves in the way that we want here.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpdateSubscriptionsRequests.kt
```diff
@@ -10,15 +10,15 @@ public data class WriteSubscriptionsRequest
 internal constructor(
     @SerialName("fcm_token") val fcmToken: String,
     val subscriptions: List<SubscriptionRequest>,
-    val locale: String,
+    val locale: String?,
```
boringcactus, line 13: Passing the locale through when clearing stale favorites is a bit tricky, and the backend will preserve the user’s existing locale setting if a new request comes in that doesn’t specify the locale, so making the locale explicitly optional seems reasonable.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -280,6 +240,48 @@ public class FavoritesViewModel(
...
+                updateFavorites(
+                    staleFavorites.mapValues { null },
+                    EditFavoritesContext.StaleCheck,
+                    // There's no real default direction, but it's only used for analytics
+                    0,
+                    fcmToken,
+                    false,
+                    locale = null,
+                )
+                sentryRepository.captureMessage("Clearing stale favorites") {
```
boringcactus, line 262: I’m not sure if we should preserve these Sentry messages, since they’re for debugging functionality that I believe I’m fixing.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -280,6 +240,48 @@ public class FavoritesViewModel(
...
+
+            fcmTokenForClearingStaleFavorites = null
+
+            val staleFavorites = getStaleFavorites(resolvedFavorites.keys, globalData)
+            if (staleFavorites.isNotEmpty()) {
+                updateFavorites(
+                    staleFavorites.mapValues { null },
+                    EditFavoritesContext.StaleCheck,
+                    // There's no real default direction, but it's only used for analytics
+                    0,
```
KaylaBrady: question(non-blocking): Would this negatively affect the accuracy of our analytics then?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
KaylaBrady: I think probably worth keeping to make sure we still see them only in legitimate cases (we should at least see this for that 47 stop that we are seeing the empty static route card data for). Probably worth making a ticket to remove this after it has been released for a while though. 

If anything, maybe worth removing the empty static card message, since that no longer results in a crash and the cause will be resolved by this issue?

KaylaBrady (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
boringcactus: If we’re leaving one in for a bit, we may as well leave both.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
boringcactus: The only thing it’s used for is the `is_default_direction` field on the `updated_favorites` event, but I can separate that out a bit more explicitly.

PR #1726, opened 2026-05-12
feat: manual socket reconnection in error banner (infinite shimmer loading mitigation)

KaylaBrady:
### Summary

_Ticket:_ [Investigate: Stuck on shimmer load sometimes on startup](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214148953775143?focus=true)

What is this PR for?

Paul shared a video of the infinite shimmer loading, where only the `joinPredictions` error message was present. It shows up after about 10 seconds, which is the default timeout for joining a prediction in SwiftPhoenixClient. Since the app loads perfectly after backgrounding & reopening, it seems like it has to do with the socket connection. 

Sockets are supposed to immediately retry connection, but maybe there is some edge case or we are hitting where it does not (https://github.com/davidstump/SwiftPhoenixClient/issues/214 from 2022 reports "A little more investigation: It seems that generally, channels do reconnect. Socket disconnects are a little less reliable, although not 100% so". I didn't dig too far into what might be a problem upstream since it seemed like a reasonable/safe enough solution to add socket re-connecting to the error banner retry button.

iOS

android

### Testing

What testing have you done?
* I am not sure how to test this in an automated way, but in both iOS & android, hardcoded a bad socket URL and confirmed socket showed up in the error banner with debug mode on, and that hitting the refresh button called `socket.attach()`
* Ran with Network Link Conditioner under Very Bad Network & 100% loss. I didn't see socket errors every time in these scenarios, but confirmed that loading data recovers automatically or when refresh is pressed after the network is restored.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/phoenix/PhoenixSocketWrapper.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/phoenix/PhoenixSocketWrapper.kt
@@ -9,6 +9,7 @@ import org.phoenixframework.Socket
 value class PhoenixSocketWrapper(private val socket: Socket) : PhoenixSocket {
 
     init {
+        socket.timeout = 6_000
         socket.reconnectAfterMs = { tries ->
             if (tries > 9) 2_000
             else listOf(10L, 50L, 100L, 150L, 200L, 250L, 500L, 1_000L, 2_000L)[tries - 1]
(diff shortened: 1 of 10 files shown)


iosApp/iosApp/ContentView.swift
```diff
@@ -64,6 +64,16 @@ struct ContentView: View {
         } action: { height in
             contentHeight = height
         }
+        .withScenePhaseHandlers(
```
KaylaBrady, line 67: One thing I wondered about a possible cause - race condition with socket.attach() being called here and `joinPredictions` being called in an inner-child view before the socket is attached. It seems safe to move this onChange handler higher up in the order, even if it doesn't directly solve a problem here. 

If there is actually a race condition going on, I think it would be quite gnarly to resolve directly.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/phoenix/PhoenixSocketWrapper.kt
```diff
@@ -9,6 +9,7 @@ import org.phoenixframework.Socket
 value class PhoenixSocketWrapper(private val socket: Socket) : PhoenixSocket {
 
     init {
+        socket.timeout = 3_000
```
KaylaBrady, line 12: p99 in the last 30 days (as recorded by backend) is 184ms. max is 6.7 seconds

boringcactus (approved):

PR #1733, opened 2026-05-18
feat: send analytics event when opening notification

boringcactus:
### Summary

_Ticket:_ [Notifications | Click analytics](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214587659492421)

Manually sends an event to Google Analytics when a notification is clicked, based on the metadata added in https://github.com/mbta/mobile_app_backend/pull/515.

iOS

android

### Testing

Checked in Firebase debug view that this works on both platforms.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
@@ -72,6 +72,15 @@ public abstract class Analytics {
         )
     }
 
+    public fun notificationClicked(analyticsLabel: String) {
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/analytics/Analytics.kt
```diff
@@ -72,6 +72,15 @@ public abstract class Analytics {
         )
     }
 
+    public fun notificationClicked(analyticsLabel: String) {
```
KaylaBrady, line 75: suggestion(non-blocking): Could use a small test for this

KaylaBrady (approved):
Great!

PR #1745, opened 2026-05-20
fix(iOS): Trip stops border radius

EmmaSimon:
### Summary

_Ticket:_ [ | Stop list border radius](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214753534941129?focus=true)

While obsessively zooming in to check the borders, I also noticed that there was a 1pt padding on the TripStops background color that was also slightly messing up the radius.

### Testing

Verified that the border radius on trip stops and route details is correct, and verified that things still look okay with the show borders accessibility setting turned on


KaylaBrady (approved):
TY for catching my oversight! 🔎

PR #1741, opened 2026-05-19
feat: small transition polish items

KaylaBrady:
### Summary

_Ticket:_ [Transition polish dev pass](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214591136371873?focus=true)

What is this PR for?

This PR includes a few small changes to transitions
* (ios): Don't trigger shimmer loading when flipping directions in stop details
* Flood route color, show direction toggles & route pill in stop details before realtime data loads
* Shimmer => skeleton only in nearby, favorites 

More to come after design pass! 

iOS
~- [ ] If you added any user-facing strings on iOS, are they included in Localizable.xcstrings?~
 ~ - [ ] Add temporary machine translations, marked "Needs Review"~

android
~- [ ] All user-facing strings added to strings resource in alphabetical order~
~- [ ] Expensive calculations are run in `withContext(Dispatchers.Default)` where possible (ideally in shared code)~

### Testing

What testing have you done?
* Ran in simulators on both platforms w/ Network Link Conditioner on 3G to test w/ dramatically slow loads
* Added iOS unit tests for direction picker visibility when no realtime stop data yet

PR #1747, opened 2026-05-21
fix(ios.StopDetailsFilteredView): Fix broken direction picker toggle on GL

KaylaBrady:
### Summary

What is this PR for?
Follow up to https://github.com/mbta/mobile_app/pull/1741. In changing the DirectionPicker constructor while moving where the DirectionPicker was called, I had mistakenly always set the stopFilter `routeId` field to the `routeId`.  Really it is `routeOrLineId` (as indicated by the type, not the name). This removes the special DirectionPicker constructor that was being used in the StopDetailsFilterView for consistency with the android implementation, and similarly just copies the old filter & changes directionId rather than building one from scratch.

iOS

android

### Testing

What testing have you done?
* Added a unit test
* Ran locally, confirmed I can now toggle GL directions

### DIFF
--- a/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
+++ b/iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
@@ -167,7 +167,7 @@ struct NearbyTransitView: View {
                         pushNavEntry: { _ in },
                         showStopHeader: true
                     )
-                    .loadingPlaceholder()
+                    .loadingPlaceholder(withShimmer: false)
                 }
             }
             .padding(.vertical, 4)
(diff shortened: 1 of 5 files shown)


iosApp/iosApp/Pages/NearbyTransit/NearbyTransitView.swift
```diff
@@ -167,7 +167,7 @@ struct NearbyTransitView: View {
                         pushNavEntry: { _ in },
                         showStopHeader: true
                     )
-                    .loadingPlaceholder()
+                    .loadingPlaceholder(withShimmer: false)
```
KaylaBrady, line 170: Noticed this was missing from the previous PR too

boringcactus (approved):

PR #1750, opened 2026-05-21
feat: show extra stop if nearest stop on pattern has a major alert

boringcactus:
### Summary

_Ticket:_ [Show an additional nearby stop if the nearest one is closed](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210112078012935)

This isn’t particularly complicated, although SwiftUI’s lack of Compose’s `remember` makes it tricky on iOS.

|iOS|Android|
|---|-------|
|Simulator Screenshot - iPhone 14 - 2026-05-21 at 15 10 50|Screenshot_20260521_152036|

On iOS, there’s a weird issue where, when panning the map, the old set of loaded stops flashes in for a bit before the sheet goes back to loading and then to the new stops. This is new on this PR, and I’m not really sure how to fix it. I plan to take a closer look in the morning, but if you can catch it before I get to it, thanks a bunch.

iOS

android

### Testing

Manually checked on both Android and iOS that the new behavior works. Added tests for the logic and at the UI level to test the state management.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
@@ -7,13 +7,13 @@ import androidx.lifecycle.ViewModel
 import com.mbta.tid.mbta_app.model.RouteCardData
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
+import com.mbta.tid.mbta_app.model.response.NearbyResponse
 import com.mbta.tid.mbta_app.model.response.PredictionsStreamDataResponse
 import com.mbta.tid.mbta_app.model.response.ScheduleResponse
 import com.mbta.tid.mbta_app.repositories.INearbyRepository
 import com.mbta.tid.mbta_app.utils.EasternTimeInstant
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import org.koin.core.component.KoinComponent
 import org.maplibre.spatialk.geojson.Position
@@ -21,10 +21,8 @@ import org.maplibre.spatialk.geojson.Position
 class NearbyTransitViewModel(private val nearbyRepository: INearbyRepository) :
     KoinComponent, ViewModel() {
     var loading by mutableStateOf(false)
-    var nearbyStopIds by mutableStateOf<List<String>?>(null)
+    var nearbyResponse by mutableStateOf<NearbyResponse?>(null)
     var routeCardData by mutableStateOf<List<RouteCardData>?>(null)
-
-    private var fetchNearbyTask: Job? = null
 
     fun getNearby(
         globalResponse: GlobalResponse,
@@ -32,18 +30,15 @@ class NearbyTransitViewModel(private val nearbyRepository: INearbyRepository) :
         setLastLocation: (Position) -> Unit,
         setIsTargeting: (Boolean) -> Unit,
     ) {
-        CoroutineScope(Dispatchers.IO).launch {
-            if (loading) {
-                fetchNearbyTask?.cancel()
-            }
-            val stopIds = nearbyRepository.getStopIdsNearby(globalResponse, location)
-            nearbyStopIds = stopIds
+        CoroutineScope(Dispatchers.Default).launch {
+            nearbyResponse = nearbyRepository.getStopIdsNearby(globalResponse, location)
             setLastLocation(location)
             setIsTargeting(false)
         }
     }
 
     fun loadRouteCardData(
+        stopIds: List<String>?,
         global: GlobalResponse?,
         location: Position?,
         schedules: ScheduleResponse?,
@@ -51,7 +46,6 @@ class NearbyTransitViewModel(private val nearbyRepository: INearbyRepository) :
         alerts: AlertsStreamDataResponse?,
         now: EasternTimeInstant,
     ) {
-        val stopIds = nearbyStopIds
         if (global == null || location == null || stopIds == null) {
             return
         }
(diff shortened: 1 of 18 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
```diff
@@ -72,11 +66,6 @@ class NearbyTransitViewModel(private val nearbyRepository: INearbyRepository) :
 
     fun reset() {
         loading = false
-        nearbyStopIds = null
-    }
-
-    override fun onCleared() {
-        super.onCleared()
-        fetchNearbyTask?.cancel()
```
KaylaBrady, line 80: question: Was onCleared ever used?

iosApp/iosApp/ViewModels/NearbyViewModel.swift
```diff
@@ -206,18 +212,28 @@ class NearbyViewModel: ObservableObject {
...
             nearbyState.loadedLocation = location
             nearbyState.loading = false
             isTargeting = false
         }
     }
 
+    func filterNearbyStops(global: GlobalResponse, alerts: AlertsStreamDataResponse?, atTime: EasternTimeInstant) {
+        Task { @MainActor [weak self] in
+            // wait for any getNearbyStops calls to have finished
+            let _ = await fetchNearbyTask?.result
```
KaylaBrady: It makes me nervous for these tasks to be interdependent, and generally how the use of tasks can introduce race conditions here (like seems to have been the problem with the route card jitter). 

question: If we also call filterNearbyStops whenever the list of nearby stops change, do we need to wait for fetchNearby here?

KaylaBrady (commented):
I haven't gotten to do a full review and need to sign off for a bit, so just leaving a few comments first. 

I added a commit that I think resolves the jitteriness & freezing card issue you were seeing. It may not have been exclusively introduced by this PR, but I think the extra async behavior may have exacerbated it. Left in one potential problem comment that I didn't address, and also things worked without addressing it, but maybe something we'd want to think more about in a spin-off / if we ever do get around to consolidating the nearby VM.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewModel.kt
boringcactus: It’s probably called by some Android ViewModel state management. Since `fetchNearbyTask` was unused anyway, deleting it just leaves `override fun onCleared() { super.onCleared() }`, and then there’s no point overriding it at all.

boringcactus:
I’m not seeing a commit from you on this branch. (Sorry that the auto generated branch names make it hard to figure out which branch is which.)

iosApp/iosApp/ViewModels/NearbyViewModel.swift
boringcactus: I had hoped that waiting for fetchNearby to finish would resolve the state mess, and it briefly looked like it worked, but it appears not to have.

KaylaBrady:
Oops sorry, just pushed!

iosApp/iosApp/ViewModels/NearbyViewModel.swift
boringcactus: I think it still makes sense, though. If we call getNearbyStops because the location changed, and then while that’s loading we call filterNearbyStops because the time changed, we don’t want the filterNearbyStops call to pick up the old nearbyResponse and use it to overwrite the new stopIds. It would be a lot simpler if SwiftUI just had `remember`, though. Maybe we should migrate the NearbyViewModel soon.

KaylaBrady: Yeah I do think this is reason to prioritize the migration. 

Still thinking about this a little bit and if there is anything small we might want to change here in the meantime. Fuzzy on main actor stuff, is it OK that we're doing an `await` in a main actor task? Does that block other rendering work?

iosApp/iosApp/ViewModels/NearbyViewModel.swift
```diff
@@ -206,18 +213,32 @@ class NearbyViewModel: ObservableObject {
...
-            let stopIds = nearbyRepository.getStopIdsNearby(
+            nearbyState.stopIds = nil
+            let nearbyResponse = nearbyRepository.getStopIdsNearby(
                 global: global,
                 location: location.positionKt
             )
+            nearbyState.nearbyResponse = nearbyResponse
             lastLoadedLocation = location
-            nearbyState.stopIds = stopIds
+            nearbyState.stopIds = nearbyResponse.filter(globalData: global, alerts: alerts, atTime: atTime)
```
KaylaBrady, line 223: question: is it an issue that stopIds are being set directly to the value filtered by alerts? If the nearby stops are A, B, C, and filtering produces stops B and C, if an alert changes such that we should show A, would we still be able to? This might be contrived.

KaylaBrady: What if NearbyState included just the pure un-edited list of nearby stops, and onChange of those stops, alerts, now, etc., in the view called something like NearbyVM.filterNearbyStops which set a separate published value for the filtered stops? I think that could potentially resolve the other race condition thread too

iosApp/iosApp/ViewModels/NearbyViewModel.swift
boringcactus: `await` shouldn’t block anything; the task should just suspend until the result is available.

iosApp/iosApp/ViewModels/NearbyViewModel.swift
boringcactus: I’m not quite sure what you mean, but I think you might be describing the way that this already works. We store the response and the filtered stop IDs so that we can re-filter the response with different parameters later, and we calculate the filtered stop IDs immediately after calculating the response because we don’t want to show a stale stop list while we wait for the onChange to fire.

boringcactus:
a08dffe appears to fix the state issues for me, thanks!

iosApp/iosApp/ViewModels/NearbyViewModel.swift
KaylaBrady: Oh yes, I am describing how this already works 🤦‍♀️ was reading too fast and conflating `nearbyState` with `nearbyResponse`.

KaylaBrady (approved):

PR #1753, opened 2026-05-27
test(ios): Add first automated accessibility test

KaylaBrady:
### Summary

_Ticket:_ [Set up automated accessibility testing for iOS](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1206196351096728?focus=true)

What is this PR for?

This PR adds a run of the ios accessibility auditor. We previously had some functionality to create mock data for UI tests, but that went stale since we moved away from UI tests. This updates that functionality using the newer `TestData`. 

Some notes:
* It seemed like there were a lot of false positives for contrast & dynamic text sizes, so I turned those two tests off. The dynamic text size one I think would be particularly nice to have automated if we could get it
* I also hit those while running the XCode manual accessibility audit tool. It is maybe easier to walk through those as a human and decide what is real or not. I do think we could consider just checking with that tool manually as needed. 
* The only "real" issue detected was for the star button tap size, will flag with design
* If we want to continue this approach of accessibility tests, we can add additional tests for the favorites flow & more page.

iOS

android

### Testing

What testing have you done?
* Ran test locally & watched that it ran
* Fixed star button tap target issue as discussed with design, confirmed it looks correct in the UI


EmmaSimon (approved):
Nice, this is cool 👀

androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
```diff
@@ -72,6 +73,9 @@ class MBTAGoMessagingService : FirebaseMessagingService() {
             defaultSoundUri,
             AudioAttributes.Builder().setUsage(AudioAttributes.USAGE_NOTIFICATION_EVENT).build(),
         )
+
+        channel.lockscreenVisibility = Notification.VISIBILITY_PRIVATE
```
KaylaBrady: Oops this is something unrelated I was playing around with, will back out!

PR #1755, opened 2026-05-28
fix?: Attempt to resolve eager error banner

BrandonTR:
### Summary

_Ticket:_ [Investigate: error banner seems to appear too often](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214494522221798?focus=true)

**Investigation:**
* There seems to be race conditions happening with the active state in our molecule VM’s
    * When you come back from background we attempt to reconnect to things like a vehicle channel. If you go back to nearby before things have had a chance to settle it’ll sometimes continue trying to reconnect since the state was set to active and our VM’s are singletons. TripDetailsViewModel -> getTripData ->  subscribeToVehicle is one instance I’ve seen this happen.
    * We only check sheetRoute for predictions — propagating this to everything might help error banner presentation specifically but doesn’t address the underlying issues
* We do attempt to clear errors through mechanisms like ChannelOwner and fetchApi. 
* The issues are exacerbated when connection is unstable.
* I thought maybe the rapid succession of calling onInactive and then onBackground when the app is going to the background could contribute to the race conditions so I attempted to throttle it so that doesn’t happen. I don’t think this is the worth focusing on, throttling didn’t help and we often do similar things in both instances. (e.g. leaving predictions, vehicle, etc.)

**Note:** These findings are mostly based on testing with iOS

**What’s changed in this PR:**
* Added a mutex to the error banner repo to try and prevent some race conditions
* Made ChannelOwner a little more forgiving on errors with a grace period.
* Increased retry time for static requests to 100 milliseconds instead of 1

These changes seem to at least help but don’t entirely resolve the issue as that seems to be a fairly gnarly thing to tackle and I’m not entirely sure the best course of action to take and I am tired of looking at this 😃 

### Testing

~~Working on tests for ChannelOwner changes.~~


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/ChannelOwner.kt
```diff
@@ -46,6 +48,14 @@ internal class AsymmetricChannelOwner<JoinData : Any, MessageData : Any>(
     private val errorBannerStateRepository: IErrorBannerStateRepository,
 ) {
     internal var channel: PhoenixChannel? = null
+    internal var lastMessageTimestamp: EasternTimeInstant? = null
+    private val gracePeriod = 30.seconds
```
KaylaBrady: @BrandonTR before you get too far on the tests for this, I'm wondering what the relationship between this and the `checkPredictionsStale` functionality is, if any. That already has a 2 minute grace period for if predictions have gone stale before showing an error message. Do we need this separate grace period?

BrandonTR: This should apply to every subscription instead of just predictions. The thinking behind this was to mitigate spamming the error banner if someone has an unstable connection but is able to receive messages at irregular frequencies.

KaylaBrady: Ah, right! Could this approach result in hiding the error banner when we shouldn't though? If it has been 25 seconds since my last message and I get an error, we wouldn't show it. Once I reach 31 seconds though, if things are still broken, would I have any way of knowing?

KaylaBrady: @BrandonTR sorry for the scattered thoughts - if the above is an issue and not one that is straightforward to fix, I think it would be OK to back out this change & punt on it for the follow-up ticket

BrandonTR: I think that is a risk so I'll back it out

KaylaBrady:
@BrandonTR that failing test is covered in the next [flaky test roundup](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1214097786423207?focus=true) - just pulled that for refinement

KaylaBrady (approved):
Sweet! Can you add any ideas you have for follow up to a ticket for next refinement?

One thing that your point on race conditions made me think is that maybe we *don't* actually want a global notion of ErrorBannerViewModel, and actually having a separate VM tied to each sheet could prevent interference / lingering references when moving between sheets. I think there was a good reason why we made it global to start, but now I don't remember it...

PR #1757, opened 2026-06-02
fix(RouteCardData.format): Don't limit # of predictions in stop details when branched

KaylaBrady:
### Summary

_Ticket:_ [Looking at GL truncated prediction list](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215285475661021?focus=true)

iOS

android

### Testing

What testing have you done?
* Added unit test
* Ran locally & confirmed at Boylston I now see all predictions
image

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -450,7 +450,7 @@ public data class RouteCardData(
             val predictionsUnavailableBranches =
                 if (remainingRowsToShow > 0) {
                     nonDisruptedHeadsigns
-                        // count as predictions unavailable only if there are no predictionsß
+                        // count as predictions unavailable only if there are no predictions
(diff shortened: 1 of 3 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -402,12 +402,22 @@ public data class RouteCardData(
                 )
             }
 
+            val maxRowsToShow =
+                if (context == Context.StopDetailsFiltered) {
+                    // We want to show every upcoming trip, plus a row for any headsign with a
+                    // disruption or unavailable predictions. This comfortably overestimates the #
+                    // of
+                    // rows to show.
+                    upcomingTrips.size + dataByHeadsign.size
```
KaylaBrady, line 411: I considered an alternative of having maxRowsToShow be nil for StopDetailsFiltered and conditionally calling `take` wherever this is used. That seemed more repetitive, so I opted for this instead.

BrandonTR (approved):

PR #1765, opened 2026-06-03
fix?(TripDetailsViewModel): Remove tripPredictionsLoaded param, only generate stop list if predictions exist

KaylaBrady:
### Summary

_Ticket:_ [Hide trip details until predictions loaded](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215327861081436?focus=true)

What is this PR for?

This PR removes the `tripPredictionsLoaded` param, which I think is left over from an older implementation and is effectively always true here, even if `tripPredictions` is null. This instead checks if `tripPredictions` exists before generating the stop list. 

Note: If we want to show the stop list in the event that tripPredictions fail to load for some reason, then I think the clearest approach would be to change `tripPpredictions` to `ApiResult<PredictionStreamResponse>?` so that we can easily differentiate pre-load vs loaded vs error cases without relying on an additional param that we have to separately manage.

iOS

android

### Testing

What testing have you done?
* Ran locally with Network Link Conditioner set to 3G. Confirmed that the trip details shimmer loads until predictions are available.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/composeStateHelpers/subscribeToTripPredictions.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/composeStateHelpers/subscribeToTripPredictions.kt
@@ -54,7 +54,9 @@ internal fun subscribeToTripPredictions(
     fun onReceive(message: ApiResult<PredictionsStreamDataResponse>) {
         onAnyMessageReceived()
         when (message) {
-            is ApiResult.Ok -> predictions = message.data
+            is ApiResult.Ok -> {
+                predictions = message.data
+            }
             is ApiResult.Error ->
                 println("Trip predictions stream failed to join: ${message.message}")
         }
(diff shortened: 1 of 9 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/composeStateHelpers/subscribeToTripPredictions.kt
```diff
@@ -75,8 +77,9 @@ internal fun subscribeToTripPredictions(
         }
     }
 
-    LaunchedEffect(tripId) { predictions = null }
     DisposableEffect(tripId, active) {
+        predictions = null
+
```
KaylaBrady, line 82: Having `predictions = null` in a separate LaunchedEffect was causing test failures now that we tests have moved away from using `tripPredictionsLoaded`. 

By adding logs, I could see that this DisposableEffect was called before & predictions were set, but then the `predictions = null` effect was called afterwards. 

I would imagine this isn't just a testing artifact and could have caused missing predictions for users, though I think they would have reappeared whenever the next push happened, so it might have just seemed like a blip.

EmmaSimon (approved):

PR #1766, opened 2026-06-04
fix(TripDetailsStopList): Don't match alerts on other routes

KaylaBrady:
### Summary

_Ticket:_ [Fix: Seemingly unrelated alert appears in Trip Details](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215327861081439?focus=true)

What is this PR for?

Resolve issue where we display an unrelated alert in trip details for added trips. 
The cause is that we are resolving route id from the schedules, and since added trips don't have schedules, the route id is resolving to nil, which for the purposes of entity matching is effectively a wildcard. 

As [discussed in slack 
](https://mbta.slack.com/archives/C062QNAJZ2M/p1780598166770059) I do want to make the typing for wildcard more explicit to help prevent this kind of bug going forward, but that is getting pretty gnarly so I'm going to do that separately.
iOS

android

### Testing

What testing have you done?
* Added a unit test.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
@@ -481,7 +481,6 @@ constructor(val trip: Trip, val stops: List<Entry>, val startTerminalEntry: Entr
             directionId: Int,
         ): UpcomingFormat.Disruption? {
             val entryTime = (entry.prediction ?: entry.schedule)?.stopTime ?: fallbackTime
-            // route ID & entryRouteType are nil, which is messing with alert matching I think
             val entryRouteType = route?.type
 
             val alert =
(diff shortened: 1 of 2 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
```diff
@@ -206,14 +206,13 @@ constructor(val trip: Trip, val stops: List<Entry>, val startTerminalEntry: Entr
                 val routeId =
                     tripPredictions?.trips?.values?.singleOrNull()?.routeId
                         ?: tripSchedules?.routeId()
+                        ?: trip.routeId
```
KaylaBrady, line 209: Maybe someone can remind me why we are resolving the routeId from the predictions & schedules in favor of reading it directly from the trip? I can't really imagine when we wouldn't want to use it from the trip

BrandonTR (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripDetailsStopList.kt
KaylaBrady: @EmmaSimon @boringcactus - any edge cases for `trip.routeId` I should be thinking about here?

EmmaSimon: Hm, no I can't think of anything off the top of my head. Maybe there are some added trip situations where a route ID doesn't exist? Or where the trip changes mid route to a new branch? But it's still a fallback only if the others are null, I can't imagine it could be worse than nothing.

PR #1769, opened 2026-06-05
refactor: Explicit wildcard type for alert informed entity matching

KaylaBrady:
### Summary

_Ticket:_ No ticket, follow up to https://github.com/mbta/mobile_app/pull/1766

What is this PR for?

The linked PR was a fix for a sneaky bug; when we are matching alert informed entities, null is effectively a wildcard. Sometimes, we'd have a null data field (route id for example) because of an invalid data scenario, not because we'd intentionally want to wildcard match on that data field

This PR makes the typing more strict to hopefully prevent situations where null data fields are unknowingly passed to alert matching as a wildcard; if we actually want a wildcard match when we have null data, we have to explicitly transform it into the new Wildcard type.

This change is the clunkiest on the `checkRoute(routeId, routeType)` predicate, where previously both values were nullable. I opted to make both required, which results in some early returns in places where we don't have both route id & type. As long as global data is up to date we should always have both pieces of data, so I feel good about doing this, but please let me know if there is an edge case I am missing that would warrant keeping them nullable! I did toy around with making routeId required and routeType nullable, which I think is doable but awkward.

iOS

android

### Testing

What testing have you done?
* Updated typing of unit tests & confirmed otherwise that they all pass as expected.


BrandonTR (approved):
Nice!

PR #1780, opened 2026-06-10
feat: Add channel debug last update times

EmmaSimon:
### Summary

_Ticket:_ [Display age of last update for predictions & alerts in debug mode](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215281963134654?focus=true)

Add to the debug view a list of connected topics, each with the last time that the channel received a successful event.

Screenshot_20260610_143639

iOS

android

### Testing

Verified the behavior on iOS and Android, added unit tests for the new DebugRepository

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
@@ -8,6 +8,7 @@ import com.mbta.tid.mbta_app.repositories.IAccessibilityStatusRepository
 import com.mbta.tid.mbta_app.repositories.IAlertsRepository
 import com.mbta.tid.mbta_app.repositories.IConfigRepository
 import com.mbta.tid.mbta_app.repositories.ICurrentAppVersionRepository
+import com.mbta.tid.mbta_app.repositories.IDebugRepository
 import com.mbta.tid.mbta_app.repositories.IErrorBannerStateRepository
 import com.mbta.tid.mbta_app.repositories.IFavoritesRepository
 import com.mbta.tid.mbta_app.repositories.IGlobalRepository
@@ -65,6 +66,7 @@ public fun repositoriesModule(repositories: IRepositories): Module {
     return module {
         single<IConfigRepository> { repositories.config }
         single<ITabPreferencesRepository> { repositories.tabPreferences }
+        single<IDebugRepository> { repositories.debug }
         single<IErrorBannerStateRepository> { repositories.errorBanner }
         single<IGlobalRepository> { repositories.global }
         single<ILastLaunchedAppVersionRepository> { repositories.lastLaunchedAppVersion }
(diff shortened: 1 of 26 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/ChannelOwner.kt
```diff
@@ -119,9 +126,17 @@ internal class AsymmetricChannelOwner<JoinData : Any, MessageData : Any>(
                 }
                 channel.onFailure {
                     handleResult(ApiResult.Error(message = "${SocketError.FAILURE} - $it"))
+                    CoroutineScope(dispatcher).launch {
+                        debugRepository.clearChannelStatus(spec.topic)
```
KaylaBrady, line 130: question: Why clear the status on failure?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
```diff
@@ -65,6 +66,7 @@ public fun repositoriesModule(repositories: IRepositories): Module {
     return module {
         single<IConfigRepository> { repositories.config }
         single<ITabPreferencesRepository> { repositories.tabPreferences }
+        single<IDebugRepository> { repositories.debug }
```
KaylaBrady, line 69: question: Does having a single DebugRepository subject us to the same race condition issues we've been experiencing with the ErrorBanner? Are there downsides to making this a factory instead?

thedecaftony:
Is the debug mode only available on the simulator/emulator? is this available as a particular build?

EmmaSimon:
> Is the debug mode only available on the simulator/emulator? is this available as a particular build?

@thedecaftony Debug mode is an option you can turn on in the More page that's only available in staging and dev orange builds, it shows relevant IDs and map debug information

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/phoenix/ChannelOwner.kt
EmmaSimon: These are failures that completely disconnect the channel, so it seemed more true to show the channel as disconnected (not listed) rather than stale

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/dependencyInjection/AppModule.kt
EmmaSimon: It's possible yeah, but I think this is a lot safer since the disconnect should always come after the last received success for a given topic. If a transition causes a reconnection to the same topic, then maybe it could connect the new one, then disconnect the old one and erase the successful connection time, but the next success would overwrite it. But I think in most cases we retain the same channel if a transition would use the exact same one though, so maybe this isn't even possible. I also wouldn't want to implement that before the error banner changes are in place so that the implementations mirror each other if we do decide to do it.

KaylaBrady (approved):

PR #1781, opened 2026-06-10
chore: Manual dependency updates

KaylaBrady:
### Summary

_Ticket:_ [Manual dependency updates](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211115350805370?focus=true)

What is this PR for?

A few manual updates for things dependabot doesn't catch. 

**Note for Review**: 
I've updated the next [manual update ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1215606746972640?focus=true) to remove the below steps I believe to be unnecessary, please confirm that is the case.

Items that are mentioned in the ticket but not included in this PR or the next manual update ticket
* mapbox for Android is updated: done recently in https://github.com/mbta/mobile_app/pull/1729. The next version 11.24 recommends a minimum android sdk version of 31, whereas we are min 28.
* buildSrc spotless is updated to match non-buildSrc spotless: we only define the spotless version in one place now
* pinned Sentry SDK version is updated to the latest version that supports all sentry dependencies: I don't think this is necessary anymore
* Android Gradle Plugin & Gradle itself updated: we get AGP bumps automatically (https://github.com/mbta/mobile_app/pull/1723) and gradle bumps automatically (https://github.com/mbta/mobile_app/blob/main/gradle/wrapper/gradle-wrapper.properties)
* buildSrc serialization plugin is updated to match Gradle version's embedded Kotlin: already up to date 

### Testing

What testing have you done?
* Confirmed android and ios still start in simulator as expected
* Will follow-up to make sure we get dependabot PRs for swift going forward

### DIFF
--- a/config.swiftformat
+++ b/config.swiftformat
@@ -1,2 +1,6 @@
 --disable wrapMultilineStatementBraces
+--disable trailingCommas
+--disable redundantViewBuilder
+--disable wrapFunctionBodies
+--disable wrapPropertyBodies
 --maxwidth 120
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,14 @@ updates:
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
       - dependency-name: "com.android.library"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
+  - package-ecosystem: pre-commit
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+    rebase-strategy: disabled
+    cooldown:
+      default-days: 7
   - package-ecosystem: swift
     directory: /iosApp
     schedule:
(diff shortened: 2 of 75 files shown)


config.swiftformat
```diff
@@ -1,2 +1,6 @@
 --disable wrapMultilineStatementBraces
+--disable trailingCommas
```
KaylaBrady, line 2: disabled these new rules from the latest SwiftFormat version based mostly on preference. 
redundantViewBuilder I disabled b/c I saw mention in the changelog of it having issues in an earlier version and I don't remember / haven't looked into the consequences of removing the annotation when it is actually needed.

thedecaftony (approved):
Is there a formatter that we use for the project?

KaylaBrady:
@thedecaftony we use [SwiftFormat](https://github.com/nicklockwood/swiftformat) on the iOS side and [Spotless](https://github.com/diffplug/spotless) for kotlin. Both Run as part of our pre-commit hoooks automatically

boringcactus (approved):

PR #1783, opened 2026-06-12
feat: expose details in error banner debug view

BrandonTR:
### Summary

_Ticket:_ [Log details in error banner message](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215281963134651?focus=true)

Expose error banner states details in a new accordion debug view. The accordion is pretty jank on iOS, we might have been better off doing like a sheet or something with the details. Might still be worth doing this so we can see all the info unhindered

Screenshot_20260611_182850

Simulator Screenshot - iPhone Air - 2026-06-11 at 18 24 51

iOS

android

### Testing

Manual


iosApp/iosApp/ComponentViews/DebugView.swift
```diff
@@ -34,3 +34,53 @@ struct DebugView<Content: View>: View {
...
+    var body: some View {
+        if settingsCache.get(.devDebugMode) {
+            ZStack {
+                Rectangle()
+                    .strokeBorder(Color(.text), style: .init(lineWidth: 2, dash: [10]))
+                DisclosureGroup(
+                    isExpanded: $expanded,
+                    content: {
+                        VStack(alignment: .leading) {
+                            content()
```
EmmaSimon: Hm, are these scrollable or does the content just get cut off like in the screenshot? If it does then I think you're right that a button to open a scrollable sheet would be better.

BrandonTR: I did try making it scrollable to see how that would work. It did work for the most part but if the error banner overflowed like it does in the screenshot it would still get cut off. I'll go ahead and see about putting it into a sheet

BrandonTR: Simulator Screenshot - iPhone Air - 2026-06-12 at 14 37 18
I like this much better! Will change Android as well.

BrandonTR: Done
Screenshot_20260612_170452

Screenshot_20260612_171108

EmmaSimon (approved):
Nice, thanks for pivoting!

PR #1785, opened 2026-06-15
feat(ErrorBannerStateRepository): Key errors by sheet

KaylaBrady:
### Summary

_Ticket:_ [ErrorBanner & VM Race Conditions](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215251110653563?focus=true)

What is this PR for?

This PR is an attempt to address some of the error banner race conditions by associating each error with the SheetRoute it came from. This is in the spirit of the proposed approach to consider ("strong reference to current sheet route in the VM [so that we know which sheet is active and can know when to ignore unrelated errors]"). 

I opted to identify the active page by the class of the SheetRoute rather than the SheetRoute itself since we don't alwaysre-fetch data when the SheetRoute changes. For example, in the following scenario: 
1. Open on Nearby
2. Navigate to StopDetailsFiltered for Park St Outbound (schedules fetch fails) 
3. Navigate to StopDetailsFiltered for Park St Inbound

If we cleared the errors between 2 and 3, the user wouldn't have a way to retry the schedules fetch. With the approach that this PR takes, they would still see the error banner & retry button. There may be cases still where this over-reports errors to users, but it should cut down on most of the noise, and I think still better to over-report than under-report.

How PRs will be split for this change:
1. pure refactor PR from this one one that changes errorKey from String to ErrorKey, with all being set to Permanent - https://github.com/mbta/mobile_app/pull/1788
2. [This PR] updates shared & android implementation to configure ErrorKeys to have the correct `sheets` value where appropriate and make `ErrorBannerStateRepository` respect the sheets
3. iOS ErrorKeys have the correct `sheets` value

~This will likely have conflict with the work to move to a shared NearbyTransitViewModel, so I may wait to do #3 until that is done.~

iOS

android

### Testing

What testing have you done?
* Updated & added unit tests
* Ran locally with 3G simulator and confirm ErrorBanner no longer shows errors for a previously opened page


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/MapAndSheetPage.kt
```diff
@@ -340,7 +340,11 @@ fun MapAndSheetPage(
                     is SheetRoutes.StopDetails -> currentNavEntry.stopFilter
                     is SheetRoutes.TripDetails -> currentNavEntry.filter.stopFilter
                     else -> null
-                }
+                },
+
+            // We re-use share the stream between stop & trip details, so we'd want errors to
+            // persist across both pages
+            errorKey = ErrorKey(KeyType.Permanent, "VehiclesViewModel.subscribeToVehicles"),
```
EmmaSimon: I don't know if this is better necessarily, but instead of `Permanent` vs `PageSpecific`, could we instead pass a set of `SheetRoutes`? Then you wouldn't need a sealed class, an empty set could mean the same thing as `Permanent`, and in cases like this where a stream is limited to multiple specific pages, we could just pass two or more.

KaylaBrady: Yeah that seems worth trying! My only concern about it is actually the same as my concern about this approach overall - it is kind of hardcoding some implementation details from elsewhere in the code in a weird way. But I like that it'll be more flexible for this case!

EmmaSimon: Yeah that's fair. I don't necessarily think we have a better way to deal with it though, and `SheetRoutes` are high level enough that I don't think it's totally unreasonable to reference them in places where they're indirectly related

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/ErrorBanner.kt
```diff
@@ -68,8 +68,9 @@ fun ErrorBanner(vm: IErrorBannerViewModel, modifier: Modifier = Modifier) {
                 },
                 button = {
                     RefreshButton(label = stringResource(R.string.reload_data)) {
-                        errorState.action()
+                        val action = errorState.action
                         vm.clearState()
+                        action()
```
KaylaBrady, line 73: This was another possible source of race conditions I found- clearing the state after running the retry action could lead to prematurely clearing any errors from the rerun.

boringcactus (approved):

PR #1788, opened 2026-06-15
refactor(ErrorBannerStateRepository): errorKey String => ErrorKey

KaylaBrady:
### Summary

Ticket: [ErrorBanner & VM Race Conditions](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215251110653563?focus=true)

What is this PR for?

Pure refactor in support of https://github.com/mbta/mobile_app/pull/1785. 

This changes every string `errorKey` to a new `ErrorKey` with a set of `SheetRoutes` that it applies too (set as suggested by @EmmaSimon  [in the follow up PR ](https://github.com/mbta/mobile_app/pull/1785#discussion_r3421473425)- thank you!). In the follow-up PR, I will convert most of these empty set keys to `PageSpecific` and add logic to ignore or clear errors that aren't for the current page.

iOS

android

### Testing

What testing have you done?
* Updated tests & confirmed existing tests pass - this is a pure refactor
* Confirmed app still runs as expected

### DIFF
--- a/iosApp/iosApp/Utils/Modifiers/GlobalModifier.swift
+++ b/iosApp/iosApp/Utils/Modifiers/GlobalModifier.swift
@@ -12,7 +12,7 @@ import SwiftUI
 struct GlobalModifier: ViewModifier {
     var globalRepository: IGlobalRepository = RepositoryDI().global
     @Binding var global: GlobalResponse?
-    let errorKey: String
+    let errorKey: ErrorKey
 
     @MainActor
     func activateGlobalListener() async {
@@ -27,7 +27,7 @@ struct GlobalModifier: ViewModifier {
         }
         Task {
             await fetchApi(
-                errorKey: "\(errorKey).loadGlobal",
+                errorKey: errorKey.withSuffix(suffix: "loadGlobal"),
                 getData: { try await globalRepository.getGlobalData() },
                 onRefreshAfterError: loadGlobal
             )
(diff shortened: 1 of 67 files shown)


iosApp/iosApp/Utils/Modifiers/GlobalModifier.swift
```diff
@@ -27,7 +27,7 @@ struct GlobalModifier: ViewModifier {
         }
         Task {
             await fetchApi(
-                errorKey: "\(errorKey).loadGlobal",
+                errorKey: errorKey.withSuffix(suffix: "loadGlobal"),
```
EmmaSimon, line 30: nitpick: Should either have a `.` in front, or `withSuffix` should include the dot in the interpolated string (and it would then need to be removed from the other usages)

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/composeStateHelpers/subscribeToVehicle.kt
```diff
@@ -9,20 +9,21 @@ import androidx.compose.runtime.setValue
...
     vehicleId: String?,
-    errorKey: String,
+    errorKey: ErrorKey,
     active: Boolean,
     errorBannerRepository: IErrorBannerStateRepository = koinInject(),
     vehicleRepository: IVehicleRepository = koinInject(),
 ): Vehicle? {
     var vehicle: Vehicle? by remember { mutableStateOf(null) }
-    val errorKey = "$errorKey.subscribeToVehicle"
+    val errorKey = errorKey.withSuffix("w+")
```
EmmaSimon: Typo?

EmmaSimon (approved):
A couple nitpicks, but overall this seems like a good approach!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/composeStateHelpers/subscribeToVehicle.kt
KaylaBrady: Bad regex replace, thanks!

PR #1789, opened 2026-06-15
feat(Android): Shared NearbyVM

EmmaSimon:
### Summary

_Ticket:_ [Move NearbyViewModel to shared code](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210525354862280?focus=true)

We never switched the Nearby VM over to a shared molecule VM when we were converting all the other ones. This creates a new shared nearby VM, and switches over to it only on Android, iOS is coming as a follow up.

android

### Testing

Manually verify that the behavior is the same, add tests for the new VM, and update existing tests


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewTest.kt
```diff
@@ -327,6 +536,6 @@ class NearbyTransitViewTest : KoinTest {
 
         alerts = AlertsStreamDataResponse(mapOf(alert.id to alert))
 
-        composeTestRule.onNodeWithText("Central").assertIsDisplayed()
+        composeTestRule.waitForIdle()
```
KaylaBrady, line 452: question: why not assert for same text?

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewTest.kt
```diff
@@ -181,6 +187,91 @@ class NearbyTransitViewTest : KoinTest {
                     )
             },
         )
+        val now = EasternTimeInstant.now()
+        val nearbyVM =
```
KaylaBrady, line 267: question(non-blocking): Any part of these 3 new mock VMs that can be shared across tests?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
```diff
@@ -42,39 +36,34 @@ fun NearbyTransitView(
...
-            nearbyVM.getNearby(globalResponse, targetLocation, setLastLocation, setIsTargeting)
-        }
+    val state by nearbyViewModel.models.collectAsStateWithLifecycle()
+    val (favorites) = manageFavorites()
+
+    LaunchedEffect(now) { nearbyViewModel.setNow(now) }
+    LaunchedEffect(alertData) { nearbyViewModel.setAlerts(alertData) }
+    LaunchedEffect(targetLocation) { nearbyViewModel.setLocation(targetLocation) }
+
+    LaunchedEffect(Unit) { nearbyViewModel.setActive(active = true, wasSentToBackground = false) }
```
KaylaBrady, line 50: question(non-blocking): It was tangential to my PR so I was wondering this already, and is tangential to this PR, but is there ever a case where `active` and `wasSentToBackground` aren't opposites?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
```diff
@@ -42,39 +36,34 @@ fun NearbyTransitView(
...
+
+    LifecycleResumeEffect(Unit) {
+        nearbyViewModel.setActive(active = true, wasSentToBackground = false)
+        onPauseOrDispose { nearbyViewModel.setActive(active = false, wasSentToBackground = true) }
     }
-    val stopIds =
```
KaylaBrady, line 54: love seeing all this logic removed from the view 🎉

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
```diff
@@ -42,39 +36,34 @@ fun NearbyTransitView(
...
-        remember(nearbyVM.nearbyResponse, globalResponse, alertData, now) {
-            if (globalResponse != null)
-                nearbyVM.nearbyResponse?.filter(globalResponse, alertData, now)
-            else null
-        }
-    val schedules = getSchedule(stopIds, "NearbyTransitView.getSchedule")
```
KaylaBrady, line 60: suggestion: I think getSchedule & subscribeToPredictions can be fully removed now too (these came up in my error key changes)

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
```diff
@@ -3,21 +3,21 @@ package com.mbta.tid.mbta_app.android.pages
...
 @Composable
 fun NearbyTransitPage(
     nearbyTransit: NearbyTransit,
     onOpenStopDetails: (String, StopDetailsFilter?) -> Unit,
     openSearch: () -> Unit,
-    nearbyViewModel: NearbyTransitViewModel,
+    nearbyViewModel: INearbyViewModel,
     errorBannerViewModel: IErrorBannerViewModel,
 ) {
-    val targetLocation by managedTargetLocation(nearbyTransit) { nearbyViewModel.reset() }
```
KaylaBrady, line 20: question: why is reset removed?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NearbyViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+
+        var active: Boolean by remember { mutableStateOf(false) }
+
+        val errorKey = "NearbyViewModel"
+        val globalData = getGlobalData(errorKey)
+        val schedules = getSchedules(stopIds, errorKey)
+        val predictions =
+            subscribeToPredictions(
+                stopIds,
+                SheetRoutes.Favorites,
```
KaylaBrady: question: why favorites vs NearbyTransit?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NearbyViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+                active,
+                errorKey,
+                onAnyMessageReceived = { awaitingPredictionsAfterBackground = false },
```
KaylaBrady, line 92: question: is `awaitingPredictionsAfterBackground` related to the `active` field at all?

KaylaBrady (commented):
So excited for this! Added a few minor questions, mostly just being conservative while noticing things that look different before.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitViewTest.kt
EmmaSimon: The old test was running everything through the full old VM route card data logic, which would have replaced Harvard with Central based on the alert, but now since the VM is fully mocked, changing the alerts doesn't change the route card data. I just updated it to check that the VM was getting the new alert data, rather than also going through the trouble of manually changing the route card data in the test.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
EmmaSimon: I thought that there were some cases where they could both be set to false, but I'm not certain off the top of my head.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/nearbyTransit/NearbyTransitView.kt
EmmaSimon: Ah right, I thought of that at one point then forgot to remove them, good catch!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NearbyViewModel.kt
EmmaSimon: Sloppy copy pasting 😅

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/NearbyTransitPage.kt
EmmaSimon: It didn't seem necessary, the favorites VM isn't doing anything to clear the routeCardData when the location changes, so I didn't include it here either. I can try adding a `LaunchedEffect(location) { routeCardData = null }` to see what happens though.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NearbyViewModel.kt
EmmaSimon: Not really, `active` is used to connect and disconnect from channels when the app is backgrounded or the app changes nav route, and `awaitingPredictionsAfterBackground` is used to suppress the stale predictions error banner when returning from background, until the predictions channel responds.

KaylaBrady (approved):

PR #1794, opened 2026-06-18
test: Flaky test roundup

KaylaBrady:
### Summary

_Ticket:_ [Flaky test roundup](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214097786423207?focus=true)

What is this PR for?

Address a few flaky tests that have been causing CI failures. They are each a little different - will call them out specifically. 

iOS

android

### Testing

What testing have you done?
Ran locally w/out failures repeatedly using for loops & `RetryableComposeTestRule`. We'll see how they do in CI!


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
```diff
@@ -96,7 +94,7 @@ class HomeMapViewTests {
     }
 
     @Test
-    fun testRecenterNotShownWhenPermissionsAndAtDefaultCenter(): Unit = runBlocking {
+    fun testRecenterNotShownWhenFollowing(): Unit = runBlocking {
```
KaylaBrady, line 97: Previously this test would have the recenter button shown then hidden. It seems more straightforward to just test that it is hidden when in following mode

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/cache/ResponseCache.kt
```diff
@@ -62,14 +62,14 @@ internal class ResponseCache<T : Any>(
 
     private val lock = Mutex()
 
-    private suspend fun getPossiblyStaleData(postprocess: (T) -> T): Response<T>? {
```
KaylaBrady, line 65: When looking at the test failure, I noticed that postprocess was being run twice (there was an extra item [0,1,2] in the flow). Removing the calls to postprocess when reading the data from the cache because it was already run when putting the data in the cache.

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/WorldCupServiceTest.kt
```diff
@@ -1,36 +0,0 @@
...
-        // failing
-        try {
-            client.get("https://mbta.com")
-        } catch (_: IOException) {
-            return@runTest
-        }
-        val response = client.get(WorldCupService.scheduleUrl)
-        assertTrue(response.status.isSuccess())
-        val responseBody: String = response.body()
-        assertContains(responseBody, "Jun 13")
```
KaylaBrady, line 28: This test was flaking and mostly tests dotcom behavior. Since the world cup is time limited, seemed worth just removing this test entirely.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
KaylaBrady, line 118: Note: While trying different possible fixes for this, I also tried switching to the new `createComposeTestRule` function since the one used here is deprecated, and [the new ones use the standard coroutine dispatcher](https://developer.android.com/develop/ui/compose/testing/migrate-v2). That produced some errors, and seems like they could have been mapbox related so I didn't pursue it further, but we may want to dedicate some standalone time to adopt it.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/map/HomeMapViewTests.kt
thedecaftony: What is following mode?

KaylaBrady: Following mode is the default when you have location services turned on - if you move, the map will pan to "follow" you, and nearby transit will update to your new location automatically. 

If you manually pan the map the map & nearby transit will be fixed to that panned location, and we will render the `RecenterButton` on the map so that you can get back to follow mode.

This test was flaky because it started in manual mode then switched to follow mode, but just starting in follow mode is more reliable and I think is in line with the spirit of the original test.

BrandonTR (approved):

PR #1795, opened 2026-06-18
Set notification visibility to public to show content when Android ph…

thedecaftony:
…ones hide sensitive content notifications

### Summary

_Ticket:_ [Show notifications on android lock screen w/ privacy setting](https://app.asana.com/1/15492006741476/project/1215456342910576/task/1215199524799205?focus=true)

What is this PR for?
Show notification content when Android device is set to hide sensitive content on notifications

iOS

android

### Testing

What testing have you done?
- Tested with the emulator, changing the notification settings to hide sensitive content from notifications, tested that it failed with no changes => implemented changes and tested that it show the content, and it did

[Back end change for notifications when app is running on the background](https://github.com/mbta/mobile_app_backend/pull/551)

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
@@ -59,6 +59,7 @@ class MBTAGoMessagingService : FirebaseMessagingService() {
                 .setAutoCancel(true)
                 .setSound(defaultSoundUri)
                 .setContentIntent(pendingIntent)
+                .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
 
         val notificationManager = NotificationManagerCompat.from(applicationContext)
 


boringcactus (commented):
Unfortunately, this change won’t actually fix this issue most of the time. For some reason, in Firebase on Android, how the notification gets created depends on whether or not the app is still running in the foreground when the notification comes in. Ours is a notification with data, so per [the Firebase docs](https://firebase.google.com/docs/cloud-messaging/android/receive-messages), if the app is running in the background (or not at all), Firebase will create the notification itself based on the [parameters](https://github.com/mbta/mobile_app_backend/blob/362ad9bd3f9636f993a403fcd12e3266f728c877/lib/mobile_app_backend/notifications/deliverer.ex#L56-L57) we sent in the backend, but if the app is running in the foreground, we have to create the notification ourselves, and that’s the only case that hits this code path. I think setting `visibility: "PUBLIC"` in that spot in the backend would fix everything else.

boringcactus (approved):

PR #1801, opened 2026-06-18
feat: introduce StopCardData for grouping favorites by stop

boringcactus:
### Summary

_Ticket:_ [Group favorites by stop | new data structure](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215721826698287)

It’s conceivable that at some point we will realize we can’t actually just reuse the `RouteCardData.Leaf` fully as-is, but I can’t think of a reason why.

iOS

android

### Testing

Added new shared test and updated existing tests. Most existing tests were just updated to calculate the new `StopCardData` from the old `RouteCardData`, which will stop making as much sense as we move favorites to grouped by stop, but some existing tests were updated to define a `Leaf` that both data structures are explicitly set to use, which will still be defensible after we’ve moved (although it’ll probably be more succinct to define the `Leaf` inline in the `StopCardData` the same way they were previously defined inline in the `RouteCardData`, which is why I didn’t use this strategy everywhere).

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
@@ -11,6 +11,7 @@ import com.mbta.tid.mbta_app.analytics.Analytics
 import com.mbta.tid.mbta_app.model.FavoriteSettings
 import com.mbta.tid.mbta_app.model.RouteCardData
 import com.mbta.tid.mbta_app.model.RouteStopDirection
+import com.mbta.tid.mbta_app.model.StopCardData
 import com.mbta.tid.mbta_app.model.response.AlertsStreamDataResponse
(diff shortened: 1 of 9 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/FavoritesViewModel.kt
```diff
@@ -340,7 +353,9 @@ public class FavoritesViewModel(
             shouldShowFirstTimeToast,
             shouldShowNotificationsHint && favorites?.isNotEmpty() == true,
             routeCardData,
+            stopCardData,
```
thedecaftony, line 356: Will this make this make the card visible or will this just provide the data to another view for it to render it after?

boringcactus: This just provides the data; the next ticket to get the UX structure in place will be where we actually display the card.

EmmaSimon (approved):
Looks good, glad this ended up being relatively straightforward

PR #1802, opened 2026-06-22
feat: add local environment alongside dev-orange/staging/prod

boringcactus:
### Summary

_Ticket:_ [make it easier to run the app pointing at a backend instance running locally](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215836694743717)

Noticed while helping Tony get set up that this could be easier than it is. It took a little creativity to allow swapping in a different URL to use whatever LAN IP actually works from real devices, but this seems like it’s both easy to use and acceptable to maintain.

iOS

android

### Testing

Checked that the local app works on Android and iOS. Weirdly, it doesn’t work on my real iPhone because "The Internet connection appears to be offline.", but it works fine on an iOS emulator and my real Android phone, and I couldn’t figure out what was causing that error.


shared/build.gradle.kts
```diff
@@ -184,6 +187,20 @@ spotless {
     }
 }
 
+run {
+    val env = EnvReader()
+
+    val localBackendOrigin = env["LOCAL_BACKEND_ORIGIN"] ?: "http://127.0.0.1:4000"
```
EmmaSimon: Won't we still need to change this between `10.0.2.2` and `127.0.0.1` every time we want to switch back and forth between android and iOS? Since afaik all of us use the standard localhost IP and default port, could we make the fallback switch between localhost or the android alias based on platform? Then if you're doing something weird you can still set the env var, but we won't need to set it most of the time.

EmmaSimon (commented):
Thanks for getting this done, this has always been an annoying step 🙏

shared/build.gradle.kts
boringcactus: Hm. I use my real LAN IP so that I can run the Android app on my real Android phone, since the Android emulator runs poorly on my computer, but I guess that might not be possible in the office. I can split things up so Android and iOS use different values and can be set independently.

EmmaSimon: Ahh, yeah that makes sense. I usually use the emulator, but it doesn't run great on my computer either. 🥲

thedecaftony (approved):
Thanks for putting up this change

PR #1803, opened 2026-06-22
feat(iOS): Shared NearbyVM

EmmaSimon:
### Summary

_Ticket:_ [Move NearbyViewModel to shared code](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210525354862280?focus=true)

iOS Follow up to https://github.com/mbta/mobile_app/pull/1789

This was trickier on iOS, I needed to split out the navigation related behavior into a new `NavigationManager` class. At first I was using it as an environment object, but that was making tests annoying, so I prop drilled it everywhere instead. I also moved the `isTargeting` and `lastLocation` fields into `ViewportProvider`.

iOS

### Testing

Updated all the tests to use the new nav object and VM (this is where the bulk of lines in this PR come from). I also removed a handful of tests which were just indirectly testing the VM behavior.


iosApp/iosApp/Pages/AlertDetails/AlertDetailsPage.swift
```diff
@@ -96,20 +98,20 @@ struct AlertDetailsPage: View {
...
-        .onChange(of: nearbyVM.alerts) { _ in updateAlert() }
+        .onChange(of: alerts) { _ in updateAlert() }
         .onReceive(timer) { input in now = input }
         .onReceive(inspection.notice) { inspection.visit(self, $0) }
         .enableInjection()
     }
 
     private func updateAlert() {
-        guard let alerts = nearbyVM.alerts else { return }
+        guard let alerts else { return }
```
thedecaftony, line 108: How are alerts being set? before I think it was checking that nearbyVM.alerts existed, is now using the alerts parameter from the struct? if that's the case shouldn't the code look like `guard alerts != nil else { return }` or does this do the same?

EmmaSimon: Yep exactly, this is the [optional unwrapping syntax](https://www.swiftbysundell.com/articles/swifts-new-shorthand-optional-unwrapping-syntax/), it basically checks that alerts is non-nil, if it is, then `alerts` is reassigned as a non-optional param and you can reference it without nil chaining below the guard, and if it's nil, the guard returns.

thedecaftony: Thank you for the explanation, makes sense

iosApp/iosApp/Utils/Modifiers/AlertsModifier.swift
BrandonTR, line 1: I like this!

BrandonTR (approved):
Looks good, nice work!

iosApp/iosApp/Utils/Modifiers/AlertsModifier.swift
EmmaSimon: I was too scared/lazy to set it up like `global` so that it could be used anywhere alerts are required with a single shared stream, to avoid prop drilling. I also didn't want to add that additional complexity, but it might be nice to get it working at some point.

PR #1804, opened 2026-06-23
fix(iOS): Don't set presentationBackground

EmmaSimon:
### Summary

_Ticket:_ [Fix iOS 17](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215928252721120?focus=true)
Introduced in #1715 

There seems to be a bug in iOS 17.4+ where this in combination with `presentationBackgroundInteraction` was causing the sheet to lock up entirely and prevent any interaction within the app. As far as I can tell, adding `presentationBackground` doesn't actually do anything when the accessibility toggle is turned on in iOS 17.4. ~WIP: Trying it out on iOS 26 to see if it makes a difference there.~ It doesn't seem to do anything in iOS 26 either.

iOS

### Testing

Verified in the simulator that the app is now functional on iOS 17.4

### DIFF
--- a/iosApp/iosApp/ContentView.swift
+++ b/iosApp/iosApp/ContentView.swift
@@ -557,11 +557,8 @@ struct ContentView: View {
                     VStack {
                         navSheetContents
                             .presentationDetents([.small, .medium, .almostFull], selection: $selectedDetent)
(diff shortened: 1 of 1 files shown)


boringcactus (approved):
Wild.

PR #1806, opened 2026-06-23
refactor: move direction from RouteStopData into leaf

boringcactus:
### Summary

_Ticket:_ [Group favorites by stop | Favorites View UX](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215746862487051)

This would have made sense as part of #1801 but I wasn’t thinking quite far enough ahead. I think it’s a good idea regardless, though.

iOS

android

### Testing

Checked and updated existing tests.


iosApp/iosApp/ComponentViews/RouteCard/RouteCardDepartures.swift
```diff
@@ -21,43 +21,41 @@ struct RouteCardDepartures: View {
...
-                    .padding(.leading, 16)
-                    .padding(.trailing, 8)
-                    .padding(.vertical, 10)
-                    .accessibilityHint(Text("Open for more arrivals"))
-                    if index < stopData.data.count - 1 {
-                        HaloSeparator()
+                        tripFilter: nil
+                    ),
+                    action: { entry in
+                        print(entry)
```
EmmaSimon: Is this a debug print that got committed?

EmmaSimon (approved):

iosApp/iosApp/ComponentViews/RouteCard/RouteCardDepartures.swift
boringcactus: It probably was originally, but it was already here before this change. It appears to have been added [more than a year ago](https://github.com/mbta/mobile_app/commit/776ed81429dd20eab839e35ebef71792a5d8c426#diff-8207b0d538d3b1060204ec265903efb7d98e1f33410afeb2a1f999def5403161R41).

PR #1823, opened 2026-06-29
feat: removing 120 min limit cutoff

thedecaftony:
### Summary

_Ticket:_ [Nearby - remove 120 min cutoff](https://app.asana.com/1/15492006741476/project/1215456342910576/task/1214831841586110?focus=true)

What is this PR for?
Removing the 120 min limit cutoff for trips that are schedule over 120 minutes can be displayed in the app

iOS

android

### Testing

What testing have you done?
- Unit testing
- Checked that trips schedules over 120 min were visible
- Shared finding with the team in slack

| Before | After |
| --- | --- |
| Screenshot_20260625_134916 | Screenshot_20260625_120253  |
| Screenshot_20260625_133943 | Screenshot_20260625_134132 |

### DIFF
--- a/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
+++ b/shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
@@ -2518,7 +2518,7 @@ class RouteCardDataTest {
         }
 
     @Test
-    fun `RouteCardData routeCardsForStopList hides rare direction with no predictions in next 120 min`() =
+    fun `RouteCardData routeCardsForStopList shows rare direction with no predictions in next 120 min`() =
         runBlocking {
             val objects = ObjectCollectionBuilder()
 
@@ -2595,7 +2595,24 @@ class RouteCardDataTest {
                                             subwayServiceStartTime = null,
                                             alertsDownstream = emptyList(),
                                             context = context,
-                                        )
+                                        ),
+                                        RouteCardData.Leaf(
+                                            lineOrRoute = lineOrRoute1,
+                                            stop = stop1,
+                                            direction = Direction(1, route1),
+                                            routePatterns = listOf(deviationInbound),
+                                            stopIds = setOf(stop1.id),
+                                            upcomingTrips =
+                                                listOf(
+                                                    objects.upcomingTrip(deviationInboundPrediction)
+                                                ),
+                                            alertsHere = emptyList(),
+                                            allDataLoaded = false,
+                                            hasSchedulesToday = false,
+                                            subwayServiceStartTime = null,
+                                            alertsDownstream = emptyList(),
+                                            context = context,
+                                        ),
                                     ),
                                 )
                             ),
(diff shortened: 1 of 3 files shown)


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
KaylaBrady: suggestion: this comment should be updated as well to reflect that it should be shown (or the comment can just be removed)

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -3375,9 +3393,10 @@ class RouteCardDataTest {
                     trip = objects.trip(scheduleLater)
                     departureTime = time + 121.minutes
                 }
+            // stop id has to be different to current stop otherwise is considered boarding
             val predictionPastPrediction =
                 objects.prediction {
-                    stopId = stop1.id
+                    stopId = stop2.id
```
KaylaBrady: issue: I think it is important that this is stop1 so that we can assert that predictions for this stop in the past are not shown. I do think there is an important comparison against `now` that we need to preserve to ensure past predictions are filtered out (and I'm wondering now how there isn't an existing user facing issue on `StopDetailsFiltered` and `Favorites` where cutoff time is already null 🤔)

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
thedecaftony: That's right, I'll update it

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
thedecaftony: I need to revisit this,  I was running on the assumption that [this code](https://github.com/mbta/mobile_app/blob/480a294feb3aa4e8a93c7a4b249b09c175cc9208/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt#L91) was running after my change, but it shouldn't.

I thought that the reason why this was showing up was because the stopId in the prediction was the same as in the vehicleId. I guess after my change validation is gone and it would show predictions in the past that don't match prediction.stopId == vehicle.stopId

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
```diff
@@ -3396,7 +3393,7 @@ class RouteCardDataTest {
             // stop id has to be different to current stop otherwise is considered boarding
```
KaylaBrady, line 3393: suggestion: this comment can now be removed too

KaylaBrady (approved):
Nice!

shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/model/RouteCardDataTest.kt
thedecaftony: Good catch, will remove it as a follow up

PR #1836, opened 2026-07-02
test(Android): make tests more robust on real devices

boringcactus:
### Summary

_Ticket:_ [Make android tests pass locally on device](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215072780897372)

Ran into some issues locally outside of the ones mentioned in Slack; this should take care of everything. Split into separate logical changes for easier review.

iOS

android

### Testing

Checked that tests pass on my physical Android phone.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -158,33 +158,23 @@ spotless {
         custom(
             "use custom default wait timeout in android UI tests",
             object : Serializable, FormatterFunc.NeedsFile {
-                override fun applyWithFile(text: String, file: File): String {
-                    if (
-                        !file.path.contains("androidTest") ||
-                            file.name == "ComposeTestRuleExtension.kt" ||
-                            file.name == "RetryableComposeTestRule.kt"
+                val replacements =
+                    mapOf(
+                        "waitUntil(" to "waitUntilDefaultTimeout(",
+                        "waitUntil {" to "waitUntilDefaultTimeout {",
+                        "waitUntilAtLeastOneExists(" to "waitUntilAtLeastOneExistsDefaultTimeout(",
+                        "waitUntilDoesNotExist(" to "waitUntilDoesNotExistDefaultTimeout(",
+                        "waitUntilExactlyOneExists(" to "waitUntilExactlyOneExistsDefaultTimeout(",
+                        "waitUntilNodeCount(" to "waitUntilNodeCountDefaultTimeout(",
+                        ".assertIsDisplayed()" to ".assertCanBeDisplayed()",
                     )
+
+                override fun applyWithFile(text: String, file: File): String {
+                    if (!file.path.contains("androidTest") || file.path.contains("testUtils"))
                         return text
-                    val lines = text.lines()
-                    for (line in lines.withIndex()) {
-                        for (fnText in
-                            arrayOf(
-                                "waitUntil(",
-                                "waitUntil {",
-                                "waitUntilAtLeastOneExists(",
-                                "waitUntilDoesNotExist(",
-                                "waitUntilExactlyOneExists(",
-                                "waitUntilNodeCount(",
-                            )) {
-                            val column = line.value.indexOf(fnText, 0, false)
-                            if (column != -1) {
-                                throw IllegalStateException(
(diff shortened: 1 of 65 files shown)


androidApp/build.gradle.kts
```diff
@@ -158,33 +158,23 @@ spotless {
...
-                            file.name == "RetryableComposeTestRule.kt"
+                val replacements =
+                    mapOf(
+                        "waitUntil(" to "waitUntilDefaultTimeout(",
+                        "waitUntil {" to "waitUntilDefaultTimeout {",
+                        "waitUntilAtLeastOneExists(" to "waitUntilAtLeastOneExistsDefaultTimeout(",
+                        "waitUntilDoesNotExist(" to "waitUntilDoesNotExistDefaultTimeout(",
+                        "waitUntilExactlyOneExists(" to "waitUntilExactlyOneExistsDefaultTimeout(",
+                        "waitUntilNodeCount(" to "waitUntilNodeCountDefaultTimeout(",
+                        ".assertIsDisplayed()" to ".assertCanBeDisplayed()",
```
KaylaBrady, line 169: question: What do you think about being more conservative here and replacing with `assertIsAlreadyOnScreen`? Seeing `assertCanBeDisplayed` in so many places has me wondering if it could cause us to miss real layout problems in tests. I could see it being used for higher-level view / page tests, but for something like RoutePillTest it doesn't seem meaningful to change. 

If `assertIsAlreadyOnScreen` is the default and someone ends up with a failure locally, I think it would be easier to detect that it might just be a scrolling problem and judgement can be used to tell if `assertCanBeDisplayed` would be appropriate or if there is a real problem.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetailsTest.kt
```diff
@@ -62,25 +62,33 @@ class AlertDetailsTest {
...
-        composeTestRule.onNodeWithText("Unruly Passenger").assertIsDisplayed()
+        composeTestRule.onNodeWithText("Red Line Suspension").assertCanBeDisplayed()
+        composeTestRule.onNodeWithText("Unruly Passenger").assertCanBeDisplayed()
         composeTestRule
-            .onNode(hasTextMatching(Regex("Wednesday, Jan 22, 10:35\\sAM")))
-            .assertIsDisplayed()
-        composeTestRule.onNodeWithText("3 affected stops").assertIsDisplayed()
+            .onNode(
+                hasTextMatching(
+                    Regex("Wednesday, (Jan 22|22 Jan), 10:35\\sAM", RegexOption.IGNORE_CASE)
```
KaylaBrady, line 70: question: Why ignore case? And why do different devices present these dates differently? This seems like something I'd expect us to be able to control entirely

KaylaBrady (commented):
Thank you for the commit breakdowns!

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/alertDetails/AlertDetailsTest.kt
boringcactus: On my phone, in these tests, this datetime shows up as “Wednesday, 22 Jan, 10:35 am”. I have no clue why my phone does that, but I do know that when we format datetimes, we pick what to show but it’s the device that picks how to show it, and for internationalization that’s the correct thing to be doing. If users set their phone to, for example, 24-hour time, we want to respect that, and I believe our current approach does (although handling that in these tests if any of us set our phones to 24-hour time would be annoying).

androidApp/build.gradle.kts
boringcactus: The Asana ticket mentions test failures that need scrolling, but I don’t see any, so I don’t know which tests actually need it. The acceptance criteria indicate that scrolling should be included “by default”, so that’s what I’ve done here. I think that kind of layout problem is unlikely to occur and would be caught in QA anyway, so my inclination would be to err on the side of being confident that tests won’t randomly break.

KaylaBrady: Fair enough!

KaylaBrady (approved):

PR #1846, opened 2026-07-08
Filter out arrival only atypical routes

thedecaftony:
### Summary

_Ticket:_ [Filter out arrival-only trips for stops that only serve non-typical trips](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1216302146592360?focus=true)

What is this PR for?
Filter out arrival only atypical routes

iOS

android

### Testing

What testing have you done?
- Tested that atypical routes are being filtered out after the change

| Before | After |
| -- | -- |
| IMG_0001 | Simulator Screenshot - iPhone 17 Pro - 2026-07-08 at 11 39 25 |

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
@@ -1211,8 +1211,11 @@ public data class RouteCardData(
             // If we don’t have any upcoming trips to tell us whether or not service is
             // arrival-only, trust the typical-last-stop-on-route-pattern state.
             val shouldBeFilteredAsArrivalOnly =
-                this.isTypicalLastStopOnRoutePattern(stop, globalData) &&
-                    (this.upcomingTrips?.isArrivalOnly() ?: true)
+                this.isLastStopOnRoutePattern(
+                    stop,
+                    globalData,
+                    listOf(RoutePattern.Typicality.Typical, RoutePattern.Typicality.Atypical),
+                ) && (this.upcomingTrips?.isArrivalOnly() ?: true)
 
             val hasUnseenTypicalPattern =
                 routePatterns?.any {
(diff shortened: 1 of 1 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
```diff
@@ -1228,12 +1231,13 @@ public data class RouteCardData(
...
+        private fun isLastStopOnRoutePattern(
             stop: Stop,
             globalData: GlobalResponse,
+            typicalityList: List<RoutePattern.Typicality>,
         ): Boolean {
             return this.routePatterns
-                ?.filter { it.typicality == RoutePattern.Typicality.Typical }
+                ?.filter { typicalityList.contains(it.typicality) }
                 ?.map { it.representativeTripId }
                 ?.takeUnless { it.isEmpty() }
```
EmmaSimon: I'm not sure if we just want to consider typical and atypical, the wording is confusing, but usually when we refer to patterns as non-typical rather than atypical, that means patterns of any typicality, including deviation and diversion. But we don't want any non-typical patterns to be used here if a typical pattern exists. I think instead of just filtering for these two typicalities, we should instead filter for typical patterns first, and only when no typical patterns exist at that stop, look at all patterns regardless of typicality. Maybe something like:

```suggestion
            return (
                this.routePatterns
                ?.filter { it.typicality == RoutePattern.Typicality.Typical }
                ?.takeUnless { it.isEmpty() } 
                ?: this.routePatterns?.takeUnless { it.isEmpty() }
            )?.map { it.representativeTripId }
```

thedecaftony: I updated the code to something similar to the suggestion

EmmaSimon (approved):
Nice!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/RouteCardData.kt
EmmaSimon: Yours is cleaner, good call on `ifEmpty`, I didn't remember that existed

PR #1852, opened 2026-07-09
refactor: pull Matcher out of Alert.InformedEntity

boringcactus:
### Summary

_Ticket:_ [Alert Summary Consolidation | Make the frontend display backend generated summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215637601111273)

If we’re going to use `Matcher` for selecting alert summaries, this seems like a reasonable move, and with `AnyOf` it lets us simplify a few things.

iOS

android

### Testing

Checked that tests still pass.

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
@@ -1,6 +1,5 @@
 package com.mbta.tid.mbta_app.model
 
-import com.mbta.tid.mbta_app.model.Alert.InformedEntity.Matcher
 import com.mbta.tid.mbta_app.model.RoutePattern.Typicality
 import com.mbta.tid.mbta_app.model.response.GlobalResponse
 import com.mbta.tid.mbta_app.utils.EasternTimeInstant
@@ -66,7 +65,7 @@ internal constructor(
      */
     public fun tripSpecificSignificance(trip: String): AlertSignificance? {
 
-        if (this.anyInformedEntitySatisfies { checkTripStrict(trip) }) {
+        if (this.anyInformedEntity { it.trip == trip }) {
             if (effect == Effect.Cancellation) {
                 return AlertSignificance.Major
             } else {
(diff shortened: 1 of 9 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
```diff
@@ -300,136 +299,58 @@ internal constructor(
...
 
         /** A more expressive way to represent a set of constraints than [appliesTo]. */
         internal inner class PredicateBuilder {
             var isSatisfied = true
 
-            fun checkActivity(activity: Activity) {
+            fun checkActivity(matcher: Matcher<Activity>) {
                 if (!isSatisfied) return
-                if (activity !in this@InformedEntity.activities) {
+                if (!activities.any { matcher.matches(it) }) {
```
KaylaBrady: question(non-blocking): In the past when looking at these predicates, I've been confused by the !isSatisfied return and isSatisfied setting at the end. Now that these have gotten simpler / each function is basically the same, can that be abstracted away?

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Matcher.kt
```diff
@@ -0,0 +1,19 @@
+package com.mbta.tid.mbta_app.model
+
+public sealed class Matcher<T : Any> {
+    public data class Data<T : Any>(val value: T) : Matcher<T>() {
+        override fun matches(actual: T?) = actual == null || actual == this.value
```
KaylaBrady: suggestion: Maybe worth some documentation here. I know we want actual == null to return true because in the context of informed entities null is a wildcard, but seeing this separate from informed entities makes that seem less intuitive

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Matcher.kt
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public data class AnyOf<T : Any>(val values: Collection<T>) : Matcher<T>() {
+        public constructor(vararg values: T) : this(values.toList())
```
KaylaBrady, line 9: I like this!

KaylaBrady (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
boringcactus: The entire `PredicateBuilder` can be abstracted away, actually, because it offers the same flexibility as `Matcher.AnyOf`.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Matcher.kt
boringcactus: Disallowing nulls here and moving the knowledge that null is a wildcard into the `.matches` caller seems like it might clarify this more explicitly.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Alert.kt
KaylaBrady: Very clean!

PR #1859, opened 2026-07-15
fix: Nearby Transit loads while on a moving train

KaylaBrady:
### Summary

_Ticket:_ [Stuck shimmer loading when app opened in motion?](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216216419090857?focus=true)

What is this PR for?

Instead of debouncing a user's changing location, we now throttle it so that we don't have to wait for them to stop moving in order to load nearby transit. 

iOS

android

### Testing

What testing have you done?
* Ran in iOS and android simulators while simulating opening the app in motion, without location shared, and with maps turned off.
    * iOS map panning felt really laggy in simulator, but that may have just been an artifact of the simulator itself. It does feel better on real device.
* TODO: finish repairing tests on both platforms, but want to get out for review of the general approach


iosApp/iosApp/Fetchers/ViewportProvider.swift
```diff
@@ -25,37 +25,39 @@ class ViewportProvider: ObservableObject, Shared.ViewportManager {
     @Published var isTargeting: Bool = false
     @Published var lastLoadedLocation: CLLocationCoordinate2D? = nil
 
-    @Published var viewport: Viewport
+    @Published var viewport: Viewport?
```
KaylaBrady, line 28: Making viewport nullable and the related changes are all so that nearby transit doesn't blip in for the default map location on initial load. This didn't appear to be an issue on android.

thedecaftony (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NearbyViewModel.kt
```diff
@@ -119,7 +119,7 @@ public class NearbyViewModel(
             val resolvedLocation = location
             if (globalData == null || resolvedLocation == null) return@LaunchedEffect
             nearbyResponse = nearbyRepository.getStopIdsNearby(globalData, resolvedLocation)
-            stopIds = nearbyResponse?.filter(globalData, alerts, now)
+            stopIds = nearbyResponse?.filter(globalData, alerts, now)?.sorted()
```
thedecaftony: Why do stopIds need to be sorted?

KaylaBrady: This way we don't re-fetch data if the set of stops I'm near stays the same but the order changes. 

I think we thought we were doing this already (maybe @boringcactus mentioned it?), maybe because of the LaunchedEffect on 125 doing set equality for the list of stopIds. But I think we'd want to do it instead at this higher-level so that we can ensure that we aren't leaving / rejoinging predictions channels via `subscribeToPredictions` every time the order changes. 

I can add a test for this to confirm if this change is necessary.

KaylaBrady: Testing this was deceptively sneaky due to `MockNearbyRepository` always returning the same list of stops. I might do a follow-up to change that behavior so that our tests are more realistic going forward.

PR #1873, opened 2026-07-17
feat: Hide error banners after backgrounding

EmmaSimon:
### Summary

_Ticket:_ [Error banner flicker when returned from background](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216216419090859?focus=true)

Hide the error banner for 5 seconds after the app returns from the background. This also clears any existing errors when the app is sent to the background.

I was initially setting it to 3 seconds, but our shortest timeout for static schedule requests is 4 seconds, so I set it to be a little longer than that. Other static requests and the socket connection have a 10 second timeout, which would be too long.

On Android, I had to move the backgrounding logic into `ContentView` because putting it directly in the `ErrorBanner` would cause the lifecycle effects to run any time the page changed and it switched to a different banner. Along those lines, I also added the background VM calls to the iOS content view, so that the banner repo state can be cleared when backgrounding on a page that has no displayed banner. In theory it should already be cleared in this situation, but it seemed worth it just in case.

iOS

android

### Testing

Verified the behavior manually, added and updated unit tests


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/ErrorBannerViewModel.kt
```diff
@@ -43,30 +50,57 @@ public class ErrorBannerViewModel(
...
 
         EventSink(eventHandlingTimeout = 1.seconds, sentryRepository = sentryRepository) { event ->
             when (event) {
                 is Event.ClearState -> errorRepository.clearState()
+                is Event.ReturnFromBackground ->
+                    if (hideBanner) backgroundReturnInstant = EasternTimeInstant.now()
+                is Event.SendToBackground -> {
+                    hideBanner = true
+                    backgroundReturnInstant = null
+                    errorRepository.clearState()
```
KaylaBrady: question: do we want to clearState when sent to the background? I think you previously raised this being risky for static requests where we don't re-fetch when coming back from the backround

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/ErrorBannerViewModel.kt
```diff
@@ -43,30 +50,57 @@ public class ErrorBannerViewModel(
 
     public data class State(
         val loadingWhenPredictionsStale: Boolean = false,
+        val hideBanner: Boolean = false,
```
KaylaBrady: suggestion(non-blocking): I wonder if baking some of the reasoning for hiding the banner into the name could be helpful for readability, ex: `isBannerDirty`.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/ManageErrorBannerBackgrounding.kt
```diff
@@ -0,0 +1,39 @@
...
+        vm.returnFromBackground()
+        if (state.hideBanner) startTimer = true
+    }
+    LifecycleEventEffect(Lifecycle.Event.ON_PAUSE) {
+        startTimer = false
+        vm.sendToBackground()
+    }
+
+    val timer: State<EasternTimeInstant>? =
+        if (startTimer) timer(updateInterval = 1.seconds) else null
```
KaylaBrady: question(non-blocking): could the VM manage the timer too in the `sendToBackground` / `returnFromBackground` events? I think I'm getting why we need to sendToBackground / returnFromBackground outside the view + viewModel, but wondering if the just the timer piece can be in shared logic since it is a bit different across platforms now.

KaylaBrady (commented):
Great, I really like this approach! A few minor comments, and just one blocking around clearing.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/ErrorBannerViewModel.kt
EmmaSimon: I think the thing that I raised was clearing on return from background, and potential race conditions between requests that could have also been triggered on the return, but now that you mention it, I think you're right that this could be an issue, I'll remove it.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/util/ManageErrorBannerBackgrounding.kt
EmmaSimon: I had the impression that timers didn't work in VMs because we use `setNow` in all the other ones, but maybe that's just for syncing the time to be consistent across multiple VMs. I'll give it a shot.

EmmaSimon: It works! It now seems a bit silly to keep `ManageErrorBannerBackgrounding`, but I figure that's the direction we should be moving for a lot of the behavior in `ContentView`, so I kept it.

KaylaBrady (approved):

PR #1876, opened 2026-07-20
fix: audit mutable collection usage

boringcactus:
### Summary

_Ticket:_ [Sentry | ErrorBanner ConcurrentModificationException](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216302146592362)

I checked every `mutableMapOf`/`mutableListOf`/`mutableSetOf` in our code. Most of them are just for local variables that only last within a single function call, so there can’t be concurrent modification; a handful are just unnecessarily mutable for no reason; and the `DebugRepository` and `ErrorBannerStateRepository` are actually concurrent in a way that could be a problem. The `DebugRepository` only ever reads or writes to its mutable state inside its mutex, but the `ErrorBannerStateRepository` reads from its mutable state outside its mutex, which is where the issues must be coming from. Propagating `suspend`s wound up being less work than I was expecting.

iOS

android

### Testing

Checked that tests still pass and the app isn’t obviously broken.


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/repositories/ErrorBannerStateRepository.kt
```diff
@@ -78,8 +79,12 @@ internal constructor(initialState: ErrorBannerState? = null) : KoinComponent {
      */
     public open fun subscribeToNetworkStatusChanges() {
         this.networkConnectivityMonitor.registerListener(
-            onNetworkAvailable = { setNetworkStatus(NetworkStatus.Connected) },
-            onNetworkLost = { setNetworkStatus(NetworkStatus.Disconnected) },
+            onNetworkAvailable = {
+                runBlocking(Dispatchers.Default) { setNetworkStatus(NetworkStatus.Connected) }
```
KaylaBrady: question: Could onNetworkAvailable be made to take a suspend fn as well so that we don't have to introduce use of `runBlocking`?

boringcactus: Not easily — Android’s ConnectivityManager doesn’t take suspend functions in its callbacks, so there’s going to be a non-suspend-to-suspend transition somewhere anyway, and it may as well live in the ErrorBannerStateRepository to make testing a little simpler — but `CoroutineScope().launch` might be more useful than `runBlocking` here.

KaylaBrady (approved):

PR #1877, opened 2026-07-21
Add smoke test plan to be used for validating previous ios versions

thedecaftony:
### Summary

_Ticket:_ [smoke test different versions of iOS](https://app.asana.com/1/15492006741476/project/1215456342910576/task/1216219303948090?focus=true)

#### What is this PR for?
Add new test plan with only testAccessibilityNearbyToTripDetails

iOS

android

#### What testing have you done?
Ran the test plan and it only executed that test

### DIFF
--- a/iosApp/iosAppSmokeTest.xctestplan
+++ b/iosApp/iosAppSmokeTest.xctestplan
@@ -1,0 +1,29 @@
+{
+  "configurations" : [
+    {
+      "id" : "13C85839-B2C7-430C-A2BF-EA3FBE5FFA0D",
+      "name" : "Configuration 1",
+      "options" : {
+
+      }
+    }
+  ],
+  "defaultOptions" : {
+    "testTimeoutsEnabled" : true
+  },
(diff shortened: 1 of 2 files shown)


iosApp/iosAppSmokeTest.xctestplan
```diff
@@ -0,0 +1,29 @@
...
+      "name" : "Configuration 1",
+      "options" : {
+
+      }
+    }
+  ],
+  "defaultOptions" : {
+    "testTimeoutsEnabled" : true
+  },
+  "testTargets" : [
```
EmmaSimon, line 14: Just to confirm, this isn't actually meant to test all the versions, just to set up the test plan to make it possible to test the other versions in a follow up?

EmmaSimon (approved):

iosApp/iosAppSmokeTest.xctestplan
thedecaftony: That's correct

PR #1879, opened 2026-07-21
Add smoke test to test-ios workflow

thedecaftony:
This change also checks for the latest iOS/iPhone versions and picks the latest two

### Summary

_Ticket:_ [smoke test different versions of iOS](https://app.asana.com/1/15492006741476/project/1215456342910576/task/1216219303948090?focus=true)

What is this PR for?
- Add smoke test to test-ios workflow
iOS

android

### Testing

What testing have you done?

Will rely on the workflows for the PR

### DIFF
--- a/.github/workflows/test-ios.yaml
+++ b/.github/workflows/test-ios.yaml
@@ -63,13 +63,39 @@ jobs:
           SENTRY_ENVIRONMENT: staging
           FIREBASE_KEY: ${{ secrets.FIREBASE_KEY }}
           GOOGLE_APP_ID_IOS: ${{ secrets.GOOGLE_APP_ID_IOS }}
+      - name: Get latest two iOS versions
+        if: ${{ !inputs.skip-everything }}
+        id: ios_versions
+        run: |
+          DEVICES=$(xcrun simctl list devices | awk '/^-- iOS/ {v=$3; next} v && /iPhone/ {print $1, $2, $3, "("v")"; v=""}' | sort -r | awk -F'[(.]' '!seen[$1]++')
+          LATEST_DEVICE=$(echo "$DEVICES" | head -n 1)
+          SECOND_LATEST_DEVICE=$(echo "$DEVICES" | head -n 2 | tail -n 1)
+          echo "LATEST_DEVICE=$LATEST_DEVICE" >> $GITHUB_OUTPUT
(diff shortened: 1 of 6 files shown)


.github/workflows/test-ios.yaml
```diff
@@ -63,13 +63,39 @@ jobs:
...
+            device:"${IOS_DEVICE}"
+        env:
+          IOS_DEVICE: ${{ steps.ios_versions.outputs.LATEST_DEVICE }}
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: failure() && !inputs.skip-everything
         with:
           name: ios-ios-results
           path: fastlane/test_output/*
+
+      - name: Run emulator smoke tests
```
KaylaBrady: @thedecaftony was there any discussion yesterday while I was out about how frequently to run this test? I see it took ~3.5 minutes to run. Since the iOS tests already take ~15-19 minutes to run looking at some other recent PRs, I'd like to avoid making them too much longer if we can. 

If we still want to run on every PR, can we do it in [parallel](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsparallel) with the other tests?

thedecaftony: That's a good callout, let me bring up the discussion on slack

.github/workflows/test-ios.yaml
```diff
@@ -63,11 +63,21 @@ jobs:
...
+        run: |
+          DEVICES=$(xcrun simctl list devices | awk '/^-- iOS/ {v=$3; next} v && /iPhone/ {print $1, $2, $3, "("v")"; v=""}' | sort -r | awk -F'[(.]' '!seen[$1]++')
+          LATEST_DEVICE=$(echo "$DEVICES" | head -n 1)
+          echo "LATEST_DEVICE=$LATEST_DEVICE" >> $GITHUB_OUTPUT
       - name: Run emulator tests
         if: ${{ !inputs.skip-everything }}
         run: |
           bundle exec fastlane ios test \
-            xcodebuild_formatter:"xcbeautify --renderer github-actions"
+            xcodebuild_formatter:"xcbeautify --quieter --is-ci --renderer github-actions" \
```
thedecaftony, line 77: Reduced the logging but kept the test results. This makes the warning messages while compiling dependencies to go away since they create a lot of noise

KaylaBrady (approved):

PR #1892, opened 2026-07-27
feat: Remove loading flickers on nearby/favorites

EmmaSimon:
### Summary

_Ticket:_ [Error banner flicker when returned from background](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216216419090859?focus=true)

Based onto #1859 

Remove the loading state between different positions being loaded on nearby transit and favorites. This removes all the logic around deleting loaded data any time the location changes, and adds additional checks to make sure that the latest data loaded matches the current map position.

While testing, I realized that putting the route card data updates in a `LaunchedEffect` keyed by all the params was resulting in the slow `RouteCardData.routeCardsForStopList` call to be interrupted any time any of the params changed, which could cause long gaps where updates were coming in too frequently to ever update the RCD in state. I changed the nearby and favorites VMs to instead use a `LaunchedEffect` to update a parameter object containing all the required data, which is then piped through a `snapshotFlow` which is sampled and conflated before being collected. Sample ensures that if the params are updated multiple times in quick succession (like a few different prediction updates coming in at the same time), the flow will only emit the latest value at most once every 100 milliseconds. Then the conflate will only emit values as fast as the collector can keep up, so once the value changes and `RouteCardData.routeCardsForStopList` is run, the flow will wait for that to complete and `routeCardData` to be set before it emits the most recent value to the collector again. This ensures that once a set of params begins being processed, it will be completed, and also that only the latest params are processed whenever the collector is ready.

Another unrelated and maybe excessive change I made in here was for Android to collapse the `NearbyTransitView` into `NearbyTransitPage`, and same for the favorites view and page, since the pages had become just direct calls to the views with no other logic.

iOS

android

### Testing

Update existing tests to pass, this involved fixing a longstanding issue in the Molecule testing helpers where `modelsForUnitTests` was not properly setting the test scope, which could result in cases where a hot flow (like the new sampled snapshot flow) would prevent the test from ever completing because it was running within a separate scope that didn't get closed when the test completed. This had the side effect of making `advanceUntilIdle()` not work in some cases, so we may need to use `advanceTimeBy()` instead in the future.

---
  - https://app.asana.com/0/0/1216721045855861


shared/src/commonTest/kotlin/com/mbta/tid/mbta_app/viewModel/MapViewModelTests.kt
```diff
@@ -534,7 +534,7 @@ internal class MapViewModelTests : KoinTest {
         testViewModelFlow(viewModel).test {
             viewModel.setViewportManager(viewportManager)
             viewModel.locationPermissionsChanged(true)
-            advanceUntilIdle()
+            advanceTimeBy(11.seconds)
```
thedecaftony, line 537: is this because it's waiting for the timeout to happen?

EmmaSimon: Yeah exactly. It also doesn't actually take 10 seconds in the test, the test uses a special clock that fast forwards when you call this function.

thedecaftony (approved):

PR #1911, opened 2026-07-31
feat: switch from FCM token to Firebase installation ID

boringcactus:
### Summary

_Ticket:_ [Move from Firebase tokens to installation IDs](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216241757035733)

Swaps out the old FCM tokens for the new installation IDs everywhere. Pairs with https://github.com/mbta/mobile_app_backend/pull/655.

iOS

android

### Testing

Updated unit tests where needed. End-to-end delivery test still needs to be done.


iosApp/iosApp/IOSApp.swift
```diff
@@ -36,13 +36,7 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
 
     func application(_: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
         Messaging.messaging().apnsToken = deviceToken
-        Messaging.messaging().token { token, error in
```
KaylaBrady, line 39: question(non-blocking): What does the new code do here since it doesn't use the token? Is `didRegisterForRemoteNotificationsWithDeviceToken` hit if the device token isn't used anymore? 

I am least familiar with how piece is set up, but it seems like the most obviously not 1-for-1 change, so figured I'd highlight it to learn more.

KaylaBrady (approved):

iosApp/iosApp/IOSApp.swift
boringcactus: The `deviceToken` here is the APNs token we get from iOS, and we still need to pass that to FCM so it can deliver messages correctly.

boringcactus:
Things are [busted](https://mbta.slack.com/archives/C05QMG9GS9M/p1786046272829149); an eventual iOS bug fix should take care of things on iOS, but it’s not obvious what’s wrong on Android.

PR #1918, opened 2026-08-07
fix: Prevent onboarding advancement on double tap

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Accessibility onboarding step is frequently skipped after allowing location permission](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217000391129153?focus=true)

Prevent the onboarding buttons from being double tapped and advancing past the next page unintentionally. The initial double advance past the permission page was slightly different, double tapping it would call `shareLocation` multiple times, and the second `launchMultiplePermissionRequest` would trigger a call to `onPermissionsResult` with an empty result map. This can still happen, but I added a check for the empty result. If the user actually approves or denies the permission, the map will include them.

I also added the screens being advanced from to the `advance` call, and wrapped the completion call in a mutex, since it was being run in a coroutine, it was possible run it multiple times using stale values, which would mark the same page complete and advance past the next page.

### Testing

Existing unit tests pass

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/OnboardingPage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/OnboardingPage.kt
@@ -2,7 +2,7 @@ package com.mbta.tid.mbta_app.android.pages
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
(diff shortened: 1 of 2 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/OnboardingPage.kt
```diff
@@ -22,20 +24,29 @@ fun OnboardingPage(
     onboardingRepository: IOnboardingRepository = koinInject(),
     skipLocationDialogue: Boolean = false,
 ) {
-    var selectedIndex by remember { mutableIntStateOf(0) }
+    var selectedIndex by remember { mutableStateOf<Int?>(0) }
```
thedecaftony, line 27: If we want the index to be able to be null to know it has been selected, should the initial value be null?

EmmaSimon: No, we always want to start with the first value in the list of uncompleted screens. When this page is opened, it should always be with a populated list of screens, so there should always be at least 1 element. The index is now set to null when the final screen is advanced past and onboarding is finished, so that we can know in the coroutine to stop processing any future advance calls.

thedecaftony (approved):

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/OnboardingPage.kt
thedecaftony: Oh perfect, thanks for the explanation

PR #1921, opened 2026-08-11
feat: display the summaries from the backend

boringcactus:
### Summary

_Ticket:_ [Alert Summary Consolidation | Make the frontend display backend generated summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215637601111273?focus=true)

Wires everything up without deleting the old path.

iOS

android

### Testing

Updated unit tests. In testing, asserted that the old value and the new value match, but took that out so it doesn’t break everything at runtime.

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCardTests.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCardTests.kt
@@ -11,6 +11,7 @@ import com.mbta.tid.mbta_app.android.util.fromHex
 import com.mbta.tid.mbta_app.model.Alert
 import com.mbta.tid.mbta_app.model.AlertCardSpec
 import com.mbta.tid.mbta_app.model.AlertSummary
+import com.mbta.tid.mbta_app.model.AlertSummaryEntity
 import com.mbta.tid.mbta_app.model.Facility
 import com.mbta.tid.mbta_app.model.ObjectCollectionBuilder
 import com.mbta.tid.mbta_app.model.RouteType
@@ -42,6 +43,7 @@ class AlertCardTests {
             AlertCard(
                 alert,
                 null,
+                null,
                 AlertCardSpec.Downstream,
                 routeAccents,
                 { onViewDetailsClicked = true },
@@ -64,6 +66,7 @@ class AlertCardTests {
         composeTestRule.setContent {
             AlertCard(
                 alert,
+                null,
                 null,
                 AlertCardSpec.Takeover,
                 routeAccents,
@@ -89,6 +92,7 @@ class AlertCardTests {
             AlertCard(
                 alert,
                 null,
+                null,
                 AlertCardSpec.Basic,
                 routeAccents,
                 { onViewDetailsClicked = true },
@@ -111,6 +115,7 @@ class AlertCardTests {
         composeTestRule.setContent {
             AlertCard(
                 alert,
+                null,
                 null,
                 AlertCardSpec.Elevator,
                 routeAccents,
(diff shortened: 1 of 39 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCardTests.kt
```diff
@@ -189,6 +197,13 @@ class AlertCardTests {
                     AlertSummary.Location.WholeRoute("Red Line", RouteType.HEAVY_RAIL),
                     timeframe = AlertSummary.Timeframe.StartingLaterToday(time),
                 ),
+                AlertSummaryEntity(
+                    null,
+                    null,
+                    null,
+                    null,
+                    "**Delay** on **Red Line** starting **9:00\u202FAM** today",
```
EmmaSimon: It would be nice to have a helper constructor for situations like this where you just pass in a string and all the params are set to null.

iosApp/iosApp/ComponentViews/StopListRow.swift
```diff
@@ -20,8 +20,29 @@ struct StopPlacement {
 }
 
 enum StopListContext {
```
EmmaSimon, line 22: Definitely not worth including another refactor on top of this PR, but do you know if there's a particular reason this isn't in shared code?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
```diff
@@ -57,9 +58,24 @@ class StopPlacement(val isFirst: Boolean = false, val isLast: Boolean = false)
 
 @Serializable
 sealed class StopListContext {
-    data object Trip : StopListContext()
+    abstract val routeIdMatcher: Matcher<Route.Id>
+    abstract val directionId: Int
+    abstract val tripIdMatcher: Matcher<String>
```
EmmaSimon, line 63: non-blocking: It feels a bit weird to me to be mixing matchers and basic values like this. I don't have a super well formed idea of an ideal data structure, but could it make sense to have something which contains all the relevant IDs (or nulls for wildcards), and has helpers which return another new kind of data class with all the relevant matchers for a summary? It could possibly also be extended to entity matching. Something like an `AlertSummaryMatchers` (and maybe an `AlertEntityMatchers`), which can be passed into `.summary` (and `.matches`). Any change is probably too much for this PR, but having all the params encapsulated with default values seems like it might make things a bit cleaner when you're passing stuff around.

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCardTests.kt
boringcactus: Good idea.

iosApp/iosApp/ComponentViews/StopListRow.swift
boringcactus: I suspect it’s just that it was trivial when it was first introduced.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/component/StopListRow.kt
boringcactus: I guess just because both the `StopListContext.Trip` and the `StopListContext.RouteDetails` have exactly one direction ID doesn’t mean the `StopListContext` shouldn’t expose its direction ID as a `Matcher`.

EmmaSimon (approved):

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/stopDetails/AlertCardTests.kt
EmmaSimon: rip 400 lines of nulls 😭

PR #1938, opened 2026-08-20
Feat: Group Android notifications by alert id + Overwrite update noti…

thedecaftony:
…fications that have the same title and body

### Summary

_Ticket:_ [All Clear | Investigate tagging notifications for grouping by OS](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217410813479642?focus=true)

Backend PR: https://github.com/mbta/mobile_app_backend/pull/698

What is this PR for?
Group Notifications for Android devices by alert id

iOS

android

### Testing

What testing have you done?
I ran a few tests
- Create alert and received notification on the background => Update it while app on the foreground => Update it again while phone app on the background => Clear notification while app on the foreground
- Create alert and received notification on the foreground => Update it while app on the background => Update it again while phone app on the foreground => Clear notification while app on the background

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
@@ -1,12 +1,14 @@
 package com.mbta.tid.mbta_app.android.notification
 
 import android.Manifest
+import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.PendingIntent
 import android.content.Intent
 import android.media.AudioAttributes
 import android.media.RingtoneManager
+import android.service.notification.StatusBarNotification
 import androidx.annotation.RequiresPermission
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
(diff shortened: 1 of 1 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/notification/MBTAGoMessagingService.kt
```diff
@@ -76,6 +68,97 @@ class MBTAGoMessagingService : FirebaseMessagingService() {
...
+    private fun updateNotification(
+        notificationManager: NotificationManagerCompat,
+        notificationBuilder: NotificationCompat.Builder,
+        prevNotification: StatusBarNotification,
+        alertId: String?,
+        channelId: String,
+        message: RemoteMessage,
+    ) {
+        // Try using the title and body from the previous notification, if they exist.
+        // Otherwise, use the title and body from the new message.
```
KaylaBrady, line 117: question: I think I'm missing why we'd want to send a notification with the old title and body. Wouldn't we always want to use the new?

thedecaftony: I see the confusion, in this case previous notification means a notification regular notification that was received that wasn't grouped yet, which will be group by the data-only notification (which is basically the same notification just data-only).

I can't really thing of a better name but I see why it's confusing, maybe I can rename it to ungrouped notification?

KaylaBrady: I think something like ungrouped would help! And/or just some extra explanation comments around this behavior,

KaylaBrady (approved):

PR #1956, opened 2026-08-28
feat(android): Notification window preset UX

KaylaBrady:
### Summary

What is this PR for?
Adds the preset notification window UX on android behind a feature flag. 
This does not address creating windows that cross the midnight boundary. 

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally in android, with and without TalkBack

https://github.com/user-attachments/assets/ca7b2612-c5c4-45c3-9b84-32abef1cccdf

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
@@ -1,0 +1,3 @@
+package com.mbta.tid.mbta_app.android.favorites
+
+class PresetWindowSelector {}
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
@@ -3,7 +3,6 @@ package com.mbta.tid.mbta_app.android.favorites
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.selection.selectable
-import androidx.compose.foundation.selection.selectableGroup
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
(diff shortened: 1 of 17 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NotificationWindowPresets.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    public object Custom : PresetSelection()
+
+    public companion object {
+        public fun selectedPresetFromSettings(
+            settings: FavoriteSettings.Notifications,
+            presetOptions: List<List<PresetWindow>>,
+        ): PresetSelection =
+            when {
+                settings.windows.isEmpty() -> Preset(1, 0)
```
thedecaftony: Does this mean when there are no windows it will default to All day?

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
```diff
@@ -0,0 +1,127 @@
...
+) {
+    val maxColumnCount = presetRows.firstOrNull()?.size ?: 0
+
+    // TODO: Accessibility of this segmented control
+    Column(
+        modifier =
+            Modifier.semantics {
+                collectionInfo =
+                    CollectionInfo(
+                        rowCount = presetRows.size + 1, // +1 for Custom
```
thedecaftony, line 42: Is custom omitted to make it simpler to handle with feature flag?

thedecaftony (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NotificationWindowPresets.kt
KaylaBrady: It does! We don't really hit that state when first adding notifications because windows isn't empty, but I think this could mess things up when deleting windows down to 0. I may make this return null instead for this case

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
KaylaBrady: I kept custom out of presetRows to simplify the typing. The entire PresetWindowSelector is shown/hidden based on the feature flag, but custom is still kind of a special case since it doesn't have a fixed window like the other options. 

I'll remove the stray TODO: This configuration of `collectionInfo` fixes how the tabs are read by indicating that it is a 3x3 grid.

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/NotificationWindowPresets.kt
KaylaBrady: Looks like we actually never have an empty list of windows - I'll default to custom anyway since that feels like a safer fallback

PR #1964, opened 2026-08-31
feat(ios): Notification window preset UX

KaylaBrady:
### Summary

What is this PR for?
iOS version of https://github.com/mbta/mobile_app/pull/1956
* Adds the preset notification window UX behind a feature flag
* Doesn't address creating windows across the midnight boundary

Note: This PR also involves a refactor to remove the iOS-specific `MutableNotificationSettings` in favor of the `NotificationSettings` in shared. This made it easier to take advantage of the new shared functions, and I suspect there will be even more shared logic once we introduce the 3AM service boundary enforcement for time pickers.

iOS

android

### Testing

What testing have you done?
* Added unit tests
* Ran locally

https://github.com/user-attachments/assets/765e781e-32f0-4404-9313-2ff4c399aaa9


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Favorite.kt
```diff
@@ -163,6 +166,72 @@ constructor(val notifications: Notifications = Notifications.disabled) {
...
+                    }
+                }
+
+                /**
+                 * The earliest possible end time for a given start time - one minute after start.
+                 */
+                public fun minimumEndTime(startTime: LocalTime): LocalTime {
+                    val startHour = startTime.hour
+                    val startMinute = startTime.minute
+                    if (startHour == 23 && startMinute == 59) {
```
thedecaftony, line 207: Would this only cover this minute or would it cover nothing?

iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
```diff
@@ -27,86 +27,112 @@ private extension DateComponents {
...
+                            setSettings(settings.doCopy(enabled: settings.enabled, windows: newWindows))
+                        },
+                        deleteWindow: { if settings.windows.count > 1 {
+                            setSettings(
+                                settings.doCopy(
+                                    enabled: settings.enabled,
+                                    windows: settings.windows.filter { $0.id != window.id }
+                                )
+                            )
+                        }
```
thedecaftony: Shouldn't the delete button be hidden until user has more than one window?

thedecaftony (approved):

iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
KaylaBrady: Yes, thanks for catching!

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/Favorite.kt
KaylaBrady: It would cover nothing, which isn't ideal but is the current behavior. With the new boundary enforcement rules I _think_ this won't be an issue once we allow crossing midnight

PR #1968, opened 2026-09-01
fix: Use departure time for CR and ferry trips

EmmaSimon:
### Summary

_Ticket:_ [Use departure_time for Commuter Rail and Ferry](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217793966475173?focus=true)

On nearby transit and stop details, show the departure times for schedule based modes, instead of arrivals. Times in trip details are still the arrival times.

Screenshot_20260901_143640

### Testing

Updated tests to check for new behavior

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
@@ -13,7 +13,7 @@ public interface TripStopTime : Comparable<TripStopTime> {
      * For Ferry and Commuter Rail, the departure time is the primary time that should be displayed,
      * but if arrival is the only time that exists, we should still fall back to that.
      */
-    public val scheduleBasedStopTime: EasternTimeInstant?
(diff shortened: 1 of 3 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/TripStopTime.kt
```diff
@@ -9,6 +9,13 @@ public interface TripStopTime : Comparable<TripStopTime> {
     public val stopTime: EasternTimeInstant?
         get() = arrivalTime ?: departureTime
 
+    /**
+     * For Ferry and Commuter Rail, the departure time is the primary time that should be displayed,
+     * but if arrival is the only time that exists, we should still fall back to that.
+     */
+    public val scheduleBasedStopTime: EasternTimeInstant?
```
KaylaBrady: suggestion(non-blocking): "schedule based" to me already has a specific meaning in the context of predictions (predictions at the start of the block before a vehicle is assigned yet). 

Maybe `departureBased` would be clearer?

KaylaBrady (approved):

PR #1971, opened 2026-09-02
feat(android.NotificationSettingsWidget): Preserve custom when switching between presets

KaylaBrady:
### Summary

What is this PR for?

If a user edits a custom window, then selects a preset, going back to custom should preserve their earlier work. 

iOS

android

### Testing

What testing have you done?
* Ran locally on iOS & android
* Wrote unit tests - this was straightforward on android, but I really struggled to get something working on iOS. Will add a separate comment to solicit ideas

### DIFF
--- a/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
+++ b/androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
@@ -385,4 +385,38 @@ class NotificationSettingsWidgetTest : KoinTest {
         composeTestRule.waitUntilExactlyOneExistsDefaultTimeout(hasText("Morning"))
         composeTestRule.onNodeWithText("Midday").assertIsSelected()
     }
+
+    @Test
+    fun testCustomPresetRestoredAfterSelectingAnotherPreset() {
+        loadKoinMocks {
+            settings =
+                MockSettingsRepository(settings = mapOf(Settings.NotificationPresetWindows to true))
+        }
+        composeTestRule.setContent {
+            var settings by remember { mutableStateOf(FavoriteSettings.Notifications.disabled) }
+            NotificationSettingsWidget(
+                settings,
+                setSettings = { settings = it },
+                notificationPermissionState = permissionGranted,
+                hasRequestedPermission = true,
+                now = EasternTimeInstant(LocalDateTime(2026, 8, 27, 4, 30, 0)),
+            )
+        }
+
+        composeTestRule.onNodeWithText("Get disruption notifications").performClick()
+        composeTestRule.waitUntilExactlyOneExistsDefaultTimeout(hasText("Morning"))
+
+        composeTestRule.onNodeWithText("Custom").performClick()
+        composeTestRule.onNodeWithText("Sunday").performClick()
+        composeTestRule.onNodeWithText("Custom").assertIsSelected()
+        composeTestRule.onNodeWithText("Sunday").assertIsOn()
+
+        composeTestRule.onNodeWithText("Morning").performClick()
+        composeTestRule.onNodeWithText("Morning").assertIsSelected()
+        composeTestRule.onNodeWithText("Sunday").assertIsOff()
+
+        composeTestRule.onNodeWithText("Custom").performClick()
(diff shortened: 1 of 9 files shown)


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
```diff
@@ -385,4 +385,38 @@ class NotificationSettingsWidgetTest : KoinTest {
         composeTestRule.waitUntilExactlyOneExistsDefaultTimeout(hasText("Morning"))
         composeTestRule.onNodeWithText("Midday").assertIsSelected()
     }
+
+    @Test
+    fun testCustomPresetRestoredAfterSelectingAnotherPreset() {
```
KaylaBrady, line 390: this test was really straightforward on android, but I haven't been able to replicate it on iOS with ViewInspector. 

I think it is a combination of ViewInspector limitations
1. Not able to assert on accessibility traits (though this is coming soon: https://github.com/nalexn/ViewInspector/issues/413)
2. `NotificationSettingsWidget` updates state `customPreset` based on the value of prop `settings`, and `settings` prop isn't changing when using ViewInspector
3. If we try to hoist NotificationSettingsWidget into a wrapper so that it is passed `settings` from a view that is managing `settings` as state, then we hit https://github.com/nalexn/ViewInspector/issues/231


I am a bit rusty on my ViewInspector testing though, and I would like to find a way to write this analogous test even if it is ugly. Please let me know if you have any suggestions to try!

iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
```diff
@@ -66,14 +68,17 @@ struct NotificationSettingsWidget: View {
...
+            windows: settings.windows,
             presetOptions: presetOptions
         )
     }
 
+    let inspection = Inspection<Self>()
+
     var body: some View {
         let permissionDenied = authorizationStatus == .denied
+        let _ = print("HELLO CUSTOM PRESET \(customPreset)")
```
EmmaSimon: Remove the debug print 😛

iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
```diff
@@ -88,16 +93,39 @@ struct NotificationSettingsWidget: View {
...
+                    .onChange(of: settings.windows) { newWindows in
+                        if PresetSelection.companion.selectedPresetFromWindows(
+                            windows: newWindows,
+                            presetOptions: presetOptions
+                        ) == PresetSelection.Custom() {
+                            print("HELLO C \(newWindows)")
+
+                            customPreset = newWindows
+                        } else {
+                            print("HELLO D \(newWindows)")
```
EmmaSimon: All of these too

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
EmmaSimon: Ughh, yeah that's frustrating, I wish iOS testing was better. I don't have any good ideas, the only thing I can think of would be moving the small amount of logic and state vars out into a view model, but that seems like overkill and would maybe just overcomplicate things...

EmmaSimon (approved):

iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
KaylaBrady: 🫣

androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
KaylaBrady: Oh hmmm yeah I could actually see that working well. I think it would conflict with the changes you're working on now, but may also be useful for encapsulating the time boundary enforcement rules... let's sync on it in the office. 

I'll merge this as-is for now at least so that if we do move to a VM, that can be just a pure refactor

PR #1985, opened 2026-09-03
refactor(NotificationSettingsWidget): Adopt new VM

KaylaBrady:
### Summary

_Ticket:_ No ticket, follow up to https://github.com/mbta/mobile_app/pull/1971

What is this PR for?

Adds a new NotificationSettingsViewModel and adopts it on android and iOS. I intended to just get this PR out adopting it for android... but I wanted to make sure the pieces would all fit on iOS too, so this PR is bigger than I would have liked. Sorry for the increased size!!

Overall I'm happier with this modelling of `Preset` and think that pushing more logic into shared is an improvement.

iOS

android

### Testing

What testing have you done?
* Ran locally
* Added & updated unit tests
    * Biggest change is to the iOS tests - now that we are using a VM, these have been re-written to be true unit tests. I ended up making the tests for `NotificationSettingsWidget` use a pure helper view instead of dealing with the mock VM setup for simplicity.


androidApp/src/androidTest/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidgetTest.kt
```diff
@@ -117,13 +111,11 @@ class NotificationSettingsWidgetTest : KoinTest {
                 )
         }
 
-        lateinit var settings: MutableState<FavoriteSettings.Notifications>
```
KaylaBrady, line 120: Some of these tests were testing a mix of UI changes + data changes - I've opted to update them to just UI changes rather than testing the innerworkings of calls to the VM

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
```diff
@@ -49,6 +49,7 @@ public fun viewModelModule(): Module = module {
         .bind(IMapViewModel::class)
     single { NearbyViewModel(get(), get(), get(named("coroutineDispatcherDefault"))) }
         .bind(INearbyViewModel::class)
+    factory { NotificationSettingsViewModel(get()) }.bind(INotificationSettingsViewModel::class)
```
KaylaBrady: If we make this a single, it would effectively save your last selected window for the app session. Will check with design if that is desirable

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/NotificationSettingsViewModel.kt
```diff
@@ -0,0 +1,184 @@
...
+import com.mbta.tid.mbta_app.model.FavoriteSettings.Notifications
+import com.mbta.tid.mbta_app.model.FavoriteSettings.Notifications.*
+import com.mbta.tid.mbta_app.model.Preset
+import com.mbta.tid.mbta_app.repositories.ISentryRepository
+import com.mbta.tid.mbta_app.utils.EasternTimeInstant
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+public interface INotificationSettingsViewModel {
```
KaylaBrady: TODO: actually load settings user has already saved

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
KaylaBrady: Update: We could make this a single, but I'm going to consider it as a follow up in case there are any other implications of that change to consider.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidget.kt
```diff
@@ -109,128 +110,116 @@ fun NotificationSettingsWidget(
...
     }
 
-    val presetWindowsEnabled = SettingsCache.get(UserSettings.NotificationPresetWindows)
+    val settings = notificationSettingsState.settings
 
-    Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
-        Column(
-            Modifier.haloContainer(borderWidth = 1.dp).clickable(
-                enabled = showPermissionSettingsLink
+    if (settings != null) {
```
thedecaftony: nit would doing something like if `(settings == null) return` reduce the nesting

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidget.kt
```diff
@@ -109,128 +110,116 @@ fun NotificationSettingsWidget(
...
+                            window,
+                            setWindow = { newWindow ->
+                                val windows = settings.windows.toMutableList()
+                                val index = windows.indexOf(window)
+                                if (index != -1) windows[index] = newWindow
+                                viewModel.setCustomWindows(windows)
+                            },
+                            deleteWindow =
+                                {
+                                        viewModel.setCustomWindows(settings.windows - window)
```
thedecaftony: nit: I'm not sure if this would be a problem, but it seems that if we do this, then if there are multiple windows with the same data, it would remove the first one it finds and it will keep the ones that appear later in the list, independently on which one was selected for deletion

iosApp/iosApp/ContentView.swift
```diff
@@ -653,6 +653,7 @@ struct ContentView: View {
                 }
 
             case let .saveFavorite(routeId, stopId, selectedDirection, context):
+                let _ = Self._printChanges() // Prints what triggered the refresh
```
thedecaftony: Should this be left here?

thedecaftony (commented):
Looks good

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/NotificationSettingsWidget.kt
KaylaBrady: I'm inclined to leave as is since I think this matches the previous behavior. It seems like an edgy edge case for a user to make duplicates of a window, and the functionality would be correct even if the ordering might shift around.

thedecaftony (approved):

PR #1990, opened 2026-09-08
feat: Update Favorites empty state

EmmaSimon:
### Summary

_Ticket:_ [Favorites empty state](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217698141303766?focus=true)

Also snuck in a small fix to not show hint when loading, and added gradle stop to fix-cocoapods script.

Simulator Screenshot - Clone 1 of iPhone 17 Pro - 2026-09-08 at 13 50 35 Screenshot_20260908_135317

iOS

android

### Testing

Updated tests relying on the empty state text

### DIFF
--- a/androidApp/src/main/res/values-es/strings.xml
+++ b/androidApp/src/main/res/values-es/strings.xml
@@ -7,9 +7,9 @@
     <string name="add_another_time_period">Agregar otro período de tiempo</string>
     <string name="add_confirmation_button">Agregar</string>
     <string name="add_favorite_label">agregar favorito</string>
-    # Used as the header for the route picker when the user is adding a favorite stop, the button
-    # to add favorite stops on the empty favorites page, and the voiceover text for the button to
-    # add more favorites when other favorites have already been added.
+    <!-- Used as the header for the route picker when the user is adding a favorite stop, the button
+    to add favorite stops on the empty favorites page, and the voiceover text for the button to
+    add more favorites when other favorites have already been added. -->
     <string name="add_favorite_stops">Agregar paradas favoritas</string>
     <string name="add_favorite_title">Agregar a favorito</string>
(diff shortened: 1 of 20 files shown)


androidApp/src/main/res/values-es/strings.xml
```diff
@@ -255,7 +258,7 @@
     <string name="no_service">Sin servicio</string>
     <string name="no_service_sentence_case">Sin servicio</string>
     <string name="no_service_today">No hay servicio hoy</string>
-    <string name="no_stops_added">No se agregaron paradas</string>
+    <string name="no_stops_added">Agrega atajada favoritas para facilitar el acceso y notificaciones de interrupciones</string>
```
thedecaftony, line 261: this should be `paradas` instead of `atajada`, but the translation seems off even with that change. I think it's because "for easy access" and "disruption notifications" are full sentence/actions in english but the translation doesn't separate them so it's translating it to something like "To facilitate the access and notification of disruptions"

thedecaftony: It makes me wonder if it's still a bit ambiguous in english as well

thedecaftony (approved):
The Spanish string translation isn't clear, can't really evaluate other languages, but I'm happy to merge this and update it smartling and put it separate

androidApp/src/main/res/values-es/strings.xml
EmmaSimon: 😞 Yeah that sounds like what I'd expect. Smartling was emailed yesterday about switching over to a different MT model, so hopefully that will help at all. But in the meantime, sounds good to update in Smartling and wait for the daily sync. Thanks for spot checking the Spanish, it's way better than the rest of us fumbling around trying to guess about languages we don't speak.

PR #1993, opened 2026-09-09
feat(NotificationSettingsViewModel): New VM to hold notification settings state

KaylaBrady:
### Summary

Ticket: No ticket, follow up to https://github.com/mbta/mobile_app/pull/1971

What is this PR for?
Adds a new NotificationSettingsViewModel for representing the in progress state of notification settings. Overall I'm happier with this simpler model for Preset and think the testability + that pushing more logic into shared is an improvement.

iOS
~  - [ ] Add temporary machine translations, marked "Needs Review"~

android

### Testing

What testing have you done?
* Added unit tests
* Adopted on iOS & android in https://github.com/mbta/mobile_app/pull/1985

### DIFF
--- a/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
+++ b/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
@@ -49,6 +49,7 @@ public fun viewModelModule(): Module = module {
         .bind(IMapViewModel::class)
     single { NearbyViewModel(get(), get(), get(named("coroutineDispatcherDefault"))) }
         .bind(INearbyViewModel::class)
+    factory { NotificationSettingsViewModel(get()) }.bind(INotificationSettingsViewModel::class)
     singleOf(::RouteCardDataViewModel).bind(IRouteCardDataViewModel::class)
     singleOf(::SearchRoutesViewModel).bind(ISearchRoutesViewModel::class)
     singleOf(::SearchViewModel).bind(ISearchViewModel::class)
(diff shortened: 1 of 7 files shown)


shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
```diff
@@ -49,6 +49,7 @@ public fun viewModelModule(): Module = module {
         .bind(IMapViewModel::class)
     single { NearbyViewModel(get(), get(), get(named("coroutineDispatcherDefault"))) }
         .bind(INearbyViewModel::class)
+    factory { NotificationSettingsViewModel(get()) }.bind(INotificationSettingsViewModel::class)
```
KaylaBrady, line 52: If we make this a single, it would effectively save your last selected window for the app session. Will check with design if that is desirable

Update: We could make this a single, but I'm going to consider it as a follow up in case there are any other implications of that change to consider.

EmmaSimon: That seems fair, but why not start with a single then see if we run into any issues with it?

EmmaSimon (approved):

shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/viewModel/viewModelModule.kt
KaylaBrady: Mostly because we already had[ a separate ticket ](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217828663233953?focus=true)it fits into better

PR #1995, opened 2026-09-10
feat(SaveFavoritesPage): Default to last saved notification settings within session

KaylaBrady:
### Summary

_Ticket:_ [Disruption Windows | Recently Used Presets](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217828663233953?focus=true)

As suspected in https://github.com/mbta/mobile_app/pull/1993, making the NotificationSettingsViewModel a singleton did most of the work of defaulting to the recently used notification window. 

There was just some additional minor work to ensure that we only used the last *saved* window rather than just the last window from an opened editor.

> [!NOTE]
> **AI Usage:** VM test boilerplate
>

#### Localization
- [ ] If you added any user-facing strings, have they been included in Localizable.xcstrings and/or the android resource files?

#### Performance

### Testing
* Added unit tests
* Ran locally and confirmed that the last saved settings were used when adding favorites and not when editing favorites.

### DIFF
--- a/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/SaveFavoritePage.kt
+++ b/androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/SaveFavoritePage.kt
@@ -127,10 +127,10 @@ fun SaveFavoritePage(
 
     LaunchedEffect(isFavorite, favorites) { if (favorites != null && !isFavorite) wasAdding = true }
 
-    val existingSettings = favorites?.get(selectedRouteStopDirection) ?: FavoriteSettings()
+    val existingSettings = favorites?.get(selectedRouteStopDirection)
     var updatedSettings by
         rememberSaveable(saver = stateJsonSaver()) { mutableStateOf<FavoriteSettings?>(null) }
-    val settings = updatedSettings ?: existingSettings
+    val settings = updatedSettings ?: existingSettings ?: FavoriteSettings()
 
     var hasRequestedPermission by rememberSaveable { mutableStateOf(false) }
     val notificationPermissionState = notificationPermissionState { hasRequestedPermission = true }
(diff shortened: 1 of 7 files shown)


androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/SaveFavoritePage.kt
```diff
@@ -127,10 +127,10 @@ fun SaveFavoritePage(
 
     LaunchedEffect(isFavorite, favorites) { if (favorites != null && !isFavorite) wasAdding = true }
 
-    val existingSettings = favorites?.get(selectedRouteStopDirection) ?: FavoriteSettings()
```
KaylaBrady, line 130: Note: [Open question to design on this](https://www.figma.com/design/NgYCfGlnM14lq8s02JPZ38/Notifications-updates?node-id=33853-756103#1921587120). Removing the default here makes it so that the last saved default is automatically used and "Get disruption notifications" toggled on if it was previously toggled on, rather than only applying the last saved default *after* the user manually toggles this on. I realize now that may not have been the intention though; will edit if needed.

KaylaBrady:
Moving back to draft for a bit - the last saved settings _should_ only populate after the toggle has been turned on

androidApp/src/main/java/com/mbta/tid/mbta_app/android/pages/SaveFavoritePage.kt
KaylaBrady: Updated to reflect design discussion: the previously saved window is only shown if the user manually toggles "Get disruption notifications" on.

thedecaftony (approved):
Looks good, is the part of the code to limit it to the last hour in a different PR?

KaylaBrady:
@thedecaftony ty for asking - I should have updated the context! Got agreement from design[ in slack](https://mbta.slack.com/archives/C05QMG9GS9M/p1788466701239189) that it is okay if we limit to using the last saved value to the app session for simplicity.

PR #2002, opened 2026-09-15
feat(NotificationSettingsWidget): UI Polish

KaylaBrady:
### Summary

_Ticket:_ [Disruption Windows | Preset options UI](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217946121625390?focus=true)

Updates the Preset UI to match design. Doesn't change the time input UI at all. 

I noticed that the last used custom windows were being used by default even if they weren't saved, so this also includes a minor VM change to clear the customWindows value when the saved settings are cleared.

> [!NOTE]
> **AI Usage:** Updating the button styling on android & copying the style changes to iOS
>

#### Localization
- [X] If you added any user-facing strings, have they been included in Localizable.xcstrings and/or the android resource files?

#### Performance

### Testing

* Added unit test for VM change
* Ran locally in iOS & android
imageimage


iosApp/iosApp/Pages/SaveFavorite/NotificationSettingsWidget.swift
```diff
@@ -9,7 +9,7 @@
 import Shared
 import SwiftUI
 
-private extension DateComponents {
+public extension DateComponents {
```
boringcactus: If these aren’t going to be file-private, they should probably be moved to their own file in `Utils/Extensions`.

androidApp/src/main/java/com/mbta/tid/mbta_app/android/favorites/PresetWindowSelector.kt
```diff
@@ -40,9 +54,29 @@ fun PresetWindowSelector(
...
+                modifier = Modifier.padding(bottom = 4.dp).fillMaxWidth().height(IntrinsicSize.Min),
+            ) {
                 windows.forEachIndexed { presetIndex, preset ->
                     val isSelected = selectedPreset == preset
+                    val amMarker = stringResource(R.string.am_marker)
+                    val pmMarker = stringResource(R.string.pm_marker)
+                    val timeFormat = LocalTime.Format {
+                        amPmHour(Padding.NONE)
+                        amPmMarker(amMarker, pmMarker)
+                    }
```
boringcactus: If someone’s phone is set to 24-hour time, and especially if it’s the default for their language, it seems like we’d benefit from respecting that. I think a
```kotlin
private val formatHour by lazy {
    DateFormat.getInstanceForSkeleton(DateFormat.HOUR).inEasternTime()
}
```
in `EasternTimeInstantExtension.kt` would do the trick.

iosApp/iosApp/Pages/SaveFavorite/PresetWindowSelector.swift
```diff
@@ -13,23 +13,44 @@ struct PresetWindowSelector: View {
...
                         let label = switch preset {
                         case .morning: NSLocalizedString("Morning", comment: "Notification window preset label")
                         case .midday: NSLocalizedString("Midday", comment: "Notification window preset label")
                         case .evening: NSLocalizedString("Evening", comment: "Notification window preset label")
                         case .allDay: NSLocalizedString("All day", comment: "Notification window preset label")
                         }
+                        let description = switch preset {
+                        case .allDay:
+                            NSLocalizedString(
+                                "start of service",
```
EmmaSimon: This should be "start to end of service"

KaylaBrady: 🤦‍♀️ I noticed the translations were wrong but not the source string!

boringcactus (approved):

PR #2006, opened 2026-09-16
fix(Alert): Return the current or next active period instead of an ended one

gregmarra:
### Summary

_Ticket:_ Community contribution, no ticket

`Alert.currentOrNextPeriod` matched periods whose end was *before* now, so it returned an already-ended period, and `null` for a bounded period that is current or upcoming. Its only caller is the "earliest later first" tiebreaker in `DisplayAlerts.forAlertsAtStop`, where `null` sorts first, so alerts with bounded upcoming periods were not ordered by start in the Later list. Now `end == null || end >= now`, the same inclusive end `activeAt` uses.

> [!NOTE]
> **AI Usage:** Claude Code found the bug and wrote the test and the one-line fix; I reviewed the change and ran the tests.
>

#### Localization
~- [ ] If you added any user-facing strings, have they been included in Localizable.xcstrings and/or the android resource files?~

#### Performance

### Testing

`AlertTest` covers the function directly; `DisplayAlertsTest` covers the ordering with two bounded upcoming alerts. Both fail on the first commit and pass on the second. `./gradlew :shared:jvmTest` is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


KaylaBrady (approved):
Thanks for the fix!

KaylaBrady:
@gregmarra sorry for the delay in merging. We have some issues to work through with our iOS test setup, but this is still on our radar!

